PR #12, opened 2018-07-31
HDB installation: Replacing bash script (called from TF) with Ansible role

tniek:
In an effort to run configuration tasks in Ansible (instead as bash scripts called from TF), this PR replaces install_HANA.sh with the saphana-install Ansible role.
Instead of setting the variables in TF, the parameters need to be specified in default config for the new Ansible role (i.e., roles/saphana-install/defaults/main.yml).


experiment/ansible/roles/saphana-install/templates/hdbserver_passwords.j2
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Replace the 3 asterisks with the password -->
+<Passwords>
+    <root_password><![CDATA[***]]></root_password>
```
persiaAziz: is this an example config file? I am asking since this is being used in main.yml

tniek: No, it's actually a template file for the hdblcm install.

PR #13, opened 2018-08-01
Separate disk creation and attachment from VM creation

pabowers:
The disk creation and attachment has been moved out of the vm creation in Terraform to allow the user to specify setting the sizes of several disks at the same time for one VM. This allows the user to use one disk for data, log, and shared, or as many as they want for each.

### DIFF
--- a/experiment/modules/single_node_hana/single-node-hana.tf
+++ b/experiment/modules/single_node_hana/single-node-hana.tf
@@ -58,21 +58,6 @@ resource "azurerm_storage_account" "mystorageaccount" {
   }
 }
 
-/*
-module "attached_disk" "disk" {
-  source = "../disk_create"
-  #count                = "${length(var.storage_disk_sizes_gb)}"
-  disk_name                 = "db${var.db_num}-disk${0}"
-  az_region            = "${var.az_region}"
-  storage_account_type = "Premium_LRS"
-  resource_group_name  = "${var.az_resource_group}"
-  create_option        = "Empty"
-  disk_size_gb         = "${var.storage_disk_sizes_gb[0]}"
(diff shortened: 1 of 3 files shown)


experiment/modules/single_node_hana/single-node-hana.tf
```diff
@@ -58,13 +58,47 @@ resource "azurerm_storage_account" "mystorageaccount" {
   }
 }
 
+/*
```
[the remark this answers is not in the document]
pabowers: @lagalbra it should now be fixed with 67ab909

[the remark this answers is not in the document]
lagalbra: Looks good.

lagalbra:
At least one person from the Landscape area should review and approve. I can approve, but I am only able to review on non-design aspects.

experiment/modules/single_node_hana/single-node-hana.tf
```diff
@@ -58,13 +58,32 @@ resource "azurerm_storage_account" "mystorageaccount" {
   }
 }
 
+resource "azurerm_managed_disk" "disk" {
+  count                = "${length(var.storage_disk_sizes_gb)}"
+  name                 = "db${var.db_num}-disk${count.index}"
```
rkrishna12: Why not keep the names consistent with what you had before? I think just using indices would make it more difficult to determine the purpose for each disk when debugging.

rkrishna12: Yes, it does makes sense to have the user-specific names for the disks in Ansible.

pabowers: sorry for the monolithic response

PR #14, opened 2018-08-02
Add Travis CI to build

pabowers:
@Azure/azure-sap-hana For now we have to be cleanly formatted to merge to master. Will look to extend later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+env:
+    - tf_version=0.11.7
+
+sudo: required
+
+language: bash
+
+before_install:
+    - wget https://releases.hashicorp.com/terraform/${tf_version}/terraform_${tf_version}_linux_amd64.zip -O /tmp/terraform.zip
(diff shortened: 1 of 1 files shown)


dwj300:
build is failing :(

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+language: bash
+
+before_install:
+    - wget https://releases.hashicorp.com/terraform/${tf_version}/terraform_${tf_version}_linux_amd64.zip -O /tmp/terraform.zip
+    - sudo unzip -d /usr/local/bin/ /tmp/terraform.zip
+
+script:
+    - terraform fmt > /tmp/terraform.out
+    - cat /tmp/terraform.out
+    - CODE=`cat /tmp/terraform.out | wc -l`
```
jovillas: It is best practice to use $() instead of ``. It will work on this case, but it's safer to do $() in general.

PR #15, opened 2018-08-03
Fixing master's build

pabowers:
@Azure/azure-sap-hana Files were removed in a previous commit, but were still referenced in the single-node-hana.tf, which caused errors in the build.  Terraform format had not been run on master, so I'm also cleaning up the formatting there.  I was having lots of errors waiting for ssh to connect with the cleanup tool and I would rather the ssh connection fails in 5 minutes than waiting 20 minutes to tell me about the nsg error.  Another realization was that null type resources cannot have tags, so that was also removed.

### DIFF
--- a/experiment/modules/single_node_hana/single-node-hana.tf
+++ b/experiment/modules/single_node_hana/single-node-hana.tf
@@ -125,28 +125,18 @@ resource "azurerm_virtual_machine" "db" {
 }
 
 resource null_resource "mount-disks-and-configure-hana" {
-  depends_on = ["azurerm_virtual_machine.db"]
+  depends_on = ["azurerm_virtual_machine.db", "azurerm_virtual_machine_data_disk_attachment.disk"]
 
   connection {
     user        = "${var.vm_user}"
     private_key = "${file("${var.sshkey_path_private}")}"
-    timeout     = "20m"
+    timeout     = "5m"
     host        = "${local.vm_fqdn}"
   }
 
   provisioner "file" {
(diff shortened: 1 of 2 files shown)


experiment/modules/single_node_hana/single-node-hana.tf
```diff
@@ -125,12 +125,12 @@ resource "azurerm_virtual_machine" "db" {
...
 
 resource null_resource "mount-disks-and-configure-hana" {
-  depends_on = ["azurerm_virtual_machine.db"]
+  depends_on = ["azurerm_virtual_machine.db", "azurerm_virtual_machine_data_disk_attachment.disk"]
 
   connection {
     user        = "${var.vm_user}"
     private_key = "${file("${var.sshkey_path_private}")}"
-    timeout     = "20m"
+    timeout     = "5m"
```
persiaAziz: Hmm , thinking if we can reduce to 1-2 minute. Even 5 minutes looks a lot for ssh timeout.

pabowers: I like the idea for the future, but I'm going to simply fix master for now.

dwj300: Agreed with Persia. Also, maybe I missed this, but make sure you are giving reviewers enough time to look at PRs. We have the 1/2 day rule for a reason.

jovillas: Agree with Douglas here.

PR #24, opened 2018-08-20
Use flag to choose between Hana 1.0 and 2.0

pabowers:
This allows the user to pick between installing HANA 1.0 or 2.0 with a simple flag.  The user will also need to supply the correct bits for HANA 2 if they use this option.   HANA 2.0 is needed for SHINE, which is a helpful tool for demos.

### DIFF
--- a/experiment/ansible/roles/saphana-install/templates/hdbserver_hana2.j2
+++ b/experiment/ansible/roles/saphana-install/templates/hdbserver_hana2.j2
@@ -1,0 +1,108 @@
+[Server]
+
+# Enable usage of persistent memory ( Default: n )
+use_pmem=n
+
+# Enable the installation or upgrade of the SAP Host Agent ( Default: y )
+install_hostagent=y
+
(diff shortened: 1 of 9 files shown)


experiment/ansible/roles/saphana-install/templates/hdbserver_hana2.j2
```diff
@@ -0,0 +1,108 @@
...
+# Enable HANA repository ( Default: y )
+repository=y
+
+# Inter Service Communication Mode ( Valid values: standard | ssl )
+isc_mode=
+
+[Action]
+
+# Action ( Default: exit; Valid values: install | update | extract_components )
+action=install
```
persiaAziz: nit: Please add a blank line at the end of file

pabowers: @persiaAziz For consistency do you want me to also change `hdbserver_install.j2`?

PR #25, opened 2018-08-23
HA pair hardware deployment works with single Ansible playbook

pabowers:
This PR will make some changes to the way that the terraform code is called.  The different pieces of infrastructure have been split into modules.  Additionally, if you would like to use a single node deployment, terraform will need to be run from the ` experiment/modules/single_node_hana` directory.  To create the infrastructure for the HA pair, terraform will be run from the `experiment/modules/ha_pair` directory.  This will allow us to have new modules for each configuration of the HANA database. Currently, both of the databases, `db0` and `db1` have HANA installed, but are not configured in an HA pair.  This configuration will happen in an upcoming PR with the Ansible plays.


persiaAziz:
Can you please put the description above in a readme?

experiment/modules/single_node_hana/variables.tf
```diff
@@ -41,11 +41,6 @@ variable "url_sap_sapcar" {
   description = "The url that points to the SAPCAR bits"
 }
 
-variable "url_sap_hostagent" {
-  type        = "string"
```
persiaAziz: does this affect the single hana installation or you removed the url_sap_hostagent entirely?

persiaAziz: is the ha_pair module sharaing any code with single_node_hana. According to the description these two should be separate. I was thinking this PR should only contain PRs related to HA Pair

persiaAziz:
Also is there  a way that user can pass ha_pair or single_node as input and the code will make terraform run from the appropriate directory?

persiaAziz:
Can we put the common files in a separate directory called "common something"? I believe it will be clearer

experiment/modules/single_node_hana/variables.tf
pabowers: @persiaAziz  re question 1: So in Tobias' Ansible code, the hostagent isn't actually being installed, and it isn't needed.  Since the variable isn't being used, I removed it.

re question 2: The modules have similar things, but since I changed the way that the database deployment happens for the HA pair, it also made sense to update the single node case so that it would continue to work.  Unfortunately, I am not sure that there is an easy way to specify the type to deploy in Terraform as an input, although I do agree that this would be a better option.

pabowers:
@persiaAziz I'm a little unsure what you are asking for with this comment https://github.com/Azure/sap-hana/pull/25#issuecomment-415607557.

experiment/modules/generic_nic_and_pip/main.tf
```diff
@@ -0,0 +1,36 @@
+# Create public IPs
+resource "azurerm_public_ip" "pip" {
+  name                         = "${var.name}-pip"
+  location                     = "${var.az_region}"
+  resource_group_name          = "${var.az_resource_group}"
+  public_ip_address_allocation = "dynamic"
```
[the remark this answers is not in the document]
pabowers: yes, will fix now.

experiment/modules/ha_pair/main.tf
```diff
@@ -0,0 +1,181 @@
...
+    private_ip_address_allocation = "Static"
+    private_ip_address            = "10.0.0.13"
+  }
+}
+
+resource "azurerm_lb_probe" "health-probe" {
+  resource_group_name = "${azurerm_resource_group.hana-resource-group.name}"
+  loadbalancer_id     = "${azurerm_lb.ha-pair-lb.id}"
+  name                = "health-probe"
+  port                = "625${var.sap_instancenum}"
```
rkrishna12: Add comment about why this port number was chosen.

experiment/modules/ha_pair/main.tf
```diff
@@ -0,0 +1,181 @@
...
+  enable_floating_ip             = true
+  frontend_ip_configuration_name = "hsr-front"
+}
+
+resource "azurerm_lb_rule" "lb-rule1" {
+  resource_group_name            = "${azurerm_resource_group.hana-resource-group.name}"
+  loadbalancer_id                = "${azurerm_lb.ha-pair-lb.id}"
+  name                           = "lb-rule-1"
+  protocol                       = "Tcp"
+  frontend_port                  = "3${var.sap_instancenum}17"
```
rkrishna12: Ditto for these ports, add comment.

experiment/modules/ha_pair/main.tf
```diff
@@ -0,0 +1,181 @@
...
+  source = "../create_db_node"
+
+  availability_set_id   = "${azurerm_availability_set.ha-pair-availset.id}"
+  az_resource_group     = "${azurerm_resource_group.hana-resource-group.name}"
+  az_region             = "${var.az_region}"
+  backend_ip_pool_ids   = ["${azurerm_lb_backend_address_pool.availability-back-pool.id}"]
+  db_num                = "1"
+  hana_subnet_id        = "${module.vnet.vnet_subnets[0]}"
+  nsg_id                = "${module.nsg.nsg-id}"
+  private_ip_address    = "10.0.0.7"
```
rkrishna12: Use variable

experiment/modules/playbook-execution/main.tf
```diff
@@ -0,0 +1,9 @@
+resource null_resource "mount-disks-and-configure-hana" {
+  provisioner "local-exec" {
+    command = "ANSIBLE_HOST_KEY_CHECKING=\"False\" ansible-playbook -u ${var.vm_user} --private-key '${var.sshkey_path_private}' --extra-vars='{\"url_sapcar\": \"${var.url_sap_sapcar}\", \"url_hdbserver\": \"${var.url_sap_hdbserver}\", \"sap_sid\": \"${var.sap_sid}\", \"sap_instancenum\": \"${var.sap_instancenum}\", \"pwd_os_sapadm\": \"${var.pw_os_sapadm}\", \"pwd_os_sidadm\": \"${var.pw_os_sidadm}\", \"pwd_db_system\": \"${var.pw_db_system}\", \"use_hana2\": \"${var.useHana2}\", \"resource_group\": \"${var.az_resource_group}\" }' -i '../../ansible/azure_rm.py' ../../ansible/playbook.yml"
```
rkrishna12: The line is quite long. Add new lines between the params.

experiment/modules/ha_pair/main.tf
```diff
@@ -0,0 +1,181 @@
...
+  loadbalancer_id     = "${azurerm_lb.ha-pair-lb.id}"
+  name                = "BackEndAddressPool-HA"
+}
+
+resource "azurerm_lb_rule" "lb-rule0" {
+  resource_group_name            = "${azurerm_resource_group.hana-resource-group.name}"
+  loadbalancer_id                = "${azurerm_lb.ha-pair-lb.id}"
+  name                           = "lb-rule-0"
+  protocol                       = "Tcp"
+  frontend_port                  = "3${var.sap_instancenum}15"
```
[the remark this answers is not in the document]
pabowers: These are standard for an SAP HA pair.  I will add some clarifying comments to the code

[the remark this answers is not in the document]
rkrishna12: In your comment, also mention if these should never be changed.

PR #27, opened 2018-08-27
Adding the ability to choose playbook for single node or HA pair

pabowers:
This probably doesn't make much sense yet, but the HANA install happens differently for the single node case and HA pair.  To make the code clean, these playbooks will be separate.  Right now, the HANA install on the HA pair infrastructure doesn't work for an HA pair configuration, so I might as well not install yet.  The PR that will do that is upcoming.


experiment/ansible/playbook.yml
```diff
@@ -23,4 +23,3 @@
         mountpoint: /hana/log/{{ sap_sid }}
   roles:
```
persiaAziz: which module does this playbook belong to?

experiment/modules/ha_pair/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "ansible_playbook_path" {
+  description = "Path from this module to the playbook"
+  default     = "../../ansible/playbook.yml"
```
persiaAziz: Can you please rename this file to something like HAPair_playbook.yml if this yaml file is only for HA pair?

experiment/modules/playbook-execution/main.tf
```diff
@@ -13,7 +13,7 @@ resource null_resource "mount-disks-and-configure-hana" {
      \"pwd_db_system\": \"${var.pw_db_system}\", \
      \"use_hana2\": \"${var.useHana2}\", \
      \"resource_group\": \"${var.az_resource_group}\" }" \
-     -i '../../ansible/azure_rm.py' ../../ansible/playbook.yml
+     -i '../../ansible/azure_rm.py' ${var.ansible_playbook_path}
      EOT
```
persiaAziz: is main.tf like an entry path for both single and ha pair installation

experiment/ansible/playbook.yml
pabowers: ha_pair for now. Would it make more sense for me to rename it?

experiment/modules/playbook-execution/main.tf
pabowers: yes

PR #29, opened 2018-08-30
set dependecy between nic and azurem_public_ip

persiaAziz:


experiment/modules/generic_nic_and_pip/main.tf
```diff
@@ -33,4 +33,5 @@ resource "azurerm_network_interface" "nic" {
   tags {
     environment = "Terraform SAP HANA single node deployment"
   }
+  depends_on = ["azurerm_public_ip.pip"]
```
pabowers: Nit: move the depends_on tag to beginning of resource instead of bottom.

persiaAziz:
done!.. please review again @pabowers

PR #31, opened 2018-09-11
Setup for HA-pair.  Exchanges Keys and sets up hostnames

pabowers:
Ensures that all necessary steps in the cookbook take place before the ha-cluster-init


experiment/ansible/roles/set-up-sbd-device/tasks/main.yml
```diff
@@ -72,32 +87,37 @@
...
+- debug:
+    msg: "{{ softdog_config }}"
+  when: softdog_config
 
 - name: Load the softdog module
-  command: modprobe -v softdog
+  shell: modprobe -v softdog
+  register: load_softdog
 
-- name: Update SLES
```
tniek: I guess we'll revisit this topic later :)

PR #37, opened 2018-09-15
Flag to enable installation of XSA and SHINE

persiaAziz:
User can choose to install XSA and SHINE on HANA VM using the flag "install_xsa_shine". The variable can be defined in a tfvars file. It is a Boolean flag meaning it accepts 'true' or 'false'. The default value is 'false'.

### DIFF
--- a/experiment/ansible/roles/shine-install/default/main.yml
+++ b/experiment/ansible/roles/shine-install/default/main.yml
@@ -7,3 +7,6 @@ pwd_db_shine:
 
 # e-mail address for SHINE user (not needed)
 email_shine: shinedemo@microsoft.com
+
+# Flag to determine whether SHINE needs to be installed or not
+install_xsa_shine:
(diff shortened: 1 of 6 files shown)


experiment/ansible/roles/shine-install/default/main.yml
```diff
@@ -7,3 +7,6 @@ pwd_db_shine:
 
 # e-mail address for SHINE user (not needed)
 email_shine: shinedemo@microsoft.com
+
+# Flag to determine whether SHINE needs to be installed or not
+install_shine:
```
tniek: It should be install_xsa_shine, since not only SHINE is installed, but XSA and then SHINE on top.

experiment/modules/playbook-execution/variables.tf
```diff
@@ -118,3 +118,8 @@ variable "email_shine" {
   description = "e-mail address for SHINE user"
   default     = ""
 }
+
+variable "install_shine" {
+  description = "Flag to determine whether to install SHINE on the host VM"
+  default     = "no"
+}
```
tniek: This should be default = false (just to be consistent with the other boolean variables)

persiaAziz:
done, please review again @tniek

PR #40, opened 2018-09-17
Decouple flags for XSA, SHINE and Cockpit

persiaAziz:


persiaAziz:
done @pabowers

experiment/modules/single_node_hana/variables.tf
```diff
@@ -132,6 +132,17 @@ variable "url_cockpit" {
   default     = ""
 }
 
-variable "install_xsa_shine" {
+variable "install_xsa" {
+  description = "Flag that determines whether to install XSA on the host"
```
tniek: nit: Why is description here different from playbook-execution/variables.tf?
I would use something for both like "Flag to determine whether to install XSA on top of HDB instance"

PR #51, opened 2018-09-21
Update README.md

persiaAziz:
forced newlines


pabowers:
why are these spaces being added?

persiaAziz:
The spaces are forcing newlines at the end of each item on the tfvars. Otherwise READMe is looking weird. Look at the current master

PR #52, opened 2018-09-24
set FIO output format

persiaAziz:


deploy/vm/Test/FIO/fio_playbook.yml
```diff
@@ -21,7 +21,7 @@
                 src: config.fio
                 dest: fio_output/config.fio
           - name: Run FIO to test
-            shell: fio config.fio > fio_out.log
+            shell: fio config.fio --output-format=json> fio_out.log
```
pabowers: nit: space between "--output-format=json" and ">"

persiaAziz: I didn't put space there so FIO does not mistake > for a parameter. Putting no space is going to clarify that it is supposed to redirect the output to the file

pabowers: ok, just wanted to clarify that there was reasoning behind it

PR #56, opened 2018-09-28
Ansible script to test using hwcct

persiaAziz:
No need to merge now


deploy/vm/test/hwcct/hwcct_playbook.yml
```diff
@@ -0,0 +1,31 @@
...
+            get_url: {{ sapcar_url }}
+            dest: /hana/shared/SAPCAR.EXE
+          - name: change permission of SAPCAR.EXE
+            file: /hana/shared/SAPCAR.EXE mode=a+x
+          - name: download HWCCT
+            get_url: {{ hwcct_url }}
+            dest: /hana/shared/hwcct.SAR
+          - name: extract hwcct.SAR using SAPCAR
+            shell: ./SAPCAR.EXE -xf HWCCT_122_17-20011536.SAR hwcct
+            args:
```
pabowers: add a creates argument so that the call will be idempotent and know if it had been previously run

persiaAziz: good catch!

deploy/vm/test/hwcct/hwcct_config.json
```diff
@@ -0,0 +1,12 @@
+{
+"use_hdb":false,
+"blades":["localhost"],
```
tniek: nit: spacing between key and value (i.e. write "use_hdb":SPACEfalse instead of "user_hdb":false)

tniek: nit: also, need to review naming convention (camel case?)

PR #59, opened 2018-10-10
Added documentation for required SAP downloads

tniek:


downloads.md
```diff
@@ -0,0 +1,26 @@
+# Required SAP packages
+This document contains a "laundry list" of SAP packages ("bits") used by the automated deployment templates. Before you can run the Terraform modules and Ansible playbooks/roles, you'll need to:
```
abhinababasu: I don't think Laundry list sounds right.

abhinababasu: This document contains a list of SAP packages used by the automated deployment templates. Before you can run the Terraform modules and Ansible playbooks/roles, you will need to do the following steps.

downloads.md
```diff
@@ -0,0 +1,26 @@
...
+* use the table below to identify which bits are required for your scenario;
```
abhinababasu: I thought you wanted to provide some sort of script to manage the downloaded  bits? E.g. put them into different blobs? Also having the az cli commands to upload to a blob is good.

downloads.md
tniek: @abhinababasu I'm fine with your version.

downloads.md
tniek: @abhinababasu That will be the long-term goal; however, there are a few legal and technical challenges (i.e. download of bits for non-SAP employees is only possible through a proprietary download manager). Just wanted to add these instructions in the meantime, so our users know where to get the bits from.

PR #60, opened 2018-10-11
Make service principal a user input rather than creating on each depl…

persiaAziz:
…oyment

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,4 +50,6 @@ Example terraform.tfvars:
  pw_db_system =  
  useHana2 =  
  install_xsa =  
+ azure_service_principal_id = 
+ azure_service_principal_pw = 
 
--- a/deploy/vm/ansible/roles/stonith-device-creation/defaults/main.yml
+++ b/deploy/vm/ansible/roles/stonith-device-creation/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+azure_service_principal_id:
+azure_service_principal_pw:
--- a/deploy/vm/modules/ha_pair/main.tf
+++ b/deploy/vm/modules/ha_pair/main.tf
@@ -182,4 +182,6 @@ module "configure_vm" {
   install_xsa                    = "${var.install_xsa}"
   install_shine                  = "${var.install_shine}"
   install_cockpit                = "${var.install_cockpit}"
+  azure_service_principal_id     = "${var.azure_service_principal_id}"
+  azure_service_principal_pw     = "${var.azure_service_principal_pw}"
 }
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -49,7 +49,10 @@ Example terraform.tfvars:
...
  install_xsa =  
+ service_principal_id = 
+ service_principal_password = 
+
 
 Additional steps
 -------------------------------------
-Please download the relevant SAP installers and store them in accessible blobs, and use SAS urls to those blobs for installation.
\ No newline at end of file
+Please download the relevant SAP installers and store them in accessible blobs, and use SAS urls to those blobs for installation.
```
pabowers: Tobias updated the documentation in the most recent version of master. Please be consistent with that.

README.md
```diff
@@ -49,7 +49,10 @@ Example terraform.tfvars:
  pw_db_system =  
  useHana2 =  
  install_xsa =  
+ service_principal_id = 
+ service_principal_password = 
+
```
tniek: We should keep the nomenclature more consistent; e.g. all passwords start with "pw_" (see line 45 -> "pw_db_system").

README.md
tniek: Maybe link to downloads.md for the time being (at least until we have a script that downloads the bits)?

deploy/vm/ansible/roles/stonith-device-creation/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+spn_id:
+spn_pw:
```
tniek: spn = service principal? Not consistent with the above naming

deploy/vm/modules/ha_pair/main.tf
```diff
@@ -182,4 +182,6 @@ module "configure_vm" {
   install_xsa                    = "${var.install_xsa}"
   install_shine                  = "${var.install_shine}"
   install_cockpit                = "${var.install_cockpit}"
+  service_principal_id           = "${var.service_principal_id}"
+  service_principal_password     = "${var.service_principal_password}"
```
tniek: Naming (see above); just from the variable name, it's hard to tell which category it belongs to (OS/HANA/Azure/...)

deploy/vm/ansible/roles/stonith-device-creation/defaults/main.yml
persiaAziz: yes.. spn is service principal. This is not a customer facing variable, so I went cheap on the naming. I can change it now.

README.md
persiaAziz: done

deploy/vm/ansible/roles/stonith-device-creation/defaults/main.yml
pabowers: I agree with Tobias with the naming

PR #67, opened 2018-10-15
Completely re-wrote README

tniek:
- Added different sections (including Getting Started and Required SAP Downloads)
- ToDo: Add separate boilerplate templates for different scenarios


persiaAziz:
Also none of the commits really reflects the changes you made.

pabowers:
I agree with Persia. Will approve once these things have been cleaned up

PR #80, opened 2018-10-23
If the setup is failing, fail instead of ending up in a limbo state

pabowers:
At present, if the HA pair setup fails on one of the two databases, the playbook continues to run, but the installation may not end up being successful.  Failing when the error actually happens makes it much easier to track down the root cause.

### DIFF
--- a/deploy/vm/ansible/ha_pair_playbook.yml
+++ b/deploy/vm/ansible/ha_pair_playbook.yml
@@ -25,6 +25,7 @@
 
 - hosts: db0,db1
   become: true
+  any_errors_fatal: true
   vars:
     db_num: "{{ ansible_hostname[-1] }}"
   roles:


deploy/vm/ansible/ha_pair_playbook.yml
```diff
@@ -25,6 +25,7 @@
 
 - hosts: db0,db1
   become: true
+  any_errors_fatal: true
```
persiaAziz: Was there any incident where it fired off?

pabowers: I ran it twice.  Once it caught an error and the other time it ran through successfully.  I think that if it has errors when I run locally, then I think that there are several that the runner may not be catching

PR #81, opened 2018-10-23
Network Security Group creation only occurs if one is not passed in

pabowers:
Terraform doesn't short-circuit logic, so the code looks a little hacky, but it does work and has been tested.

### DIFF
--- a/deploy/vm/modules/nsg_for_hana/nsg.tf
+++ b/deploy/vm/modules/nsg_for_hana/nsg.tf
@@ -5,6 +5,7 @@ data "http" "local_ip" {
 
 # Create Network Security Group and rule
 resource "azurerm_network_security_group" "sap-nsg" {
+  count               = "${var.enable ? 1 : 0}"
   name                = "${var.sap_sid}-nsg"
   location            = "${var.az_region}"
   resource_group_name = "${var.resource_group_name}"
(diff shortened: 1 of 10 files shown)


deploy/vm/modules/nsg_for_hana/variables.tf
```diff
@@ -1,5 +1,10 @@
 variable "az_region" {}
 
+variable "enable" {
```
persiaAziz: Can you rename the variable name to be more descriptive? like use_existing_nsg or something?

deploy/vm/modules/nsg_for_hana/nsg.tf
```diff
@@ -54,7 +55,7 @@ resource "azurerm_network_security_group" "sap-nsg" {
 
 # The Ports that HANA 1 uses are different from the ones HANA 2 uses
 resource "azurerm_network_security_rule" "hana1-http" {
-  count                       = "${!var.useHana2 ? 1 : 0}"                       # The rule is only created if we use HANA 1
+  count                       = "${var.enable ? (!var.useHana2 ? 1 : 0) : 0}"    # The rule is only created if we use HANA 1
```
persiaAziz: how is using existing nsg related to useHana2?

pabowers: done

pabowers: sorry, I replied to the wrong comment.  It isn't related, but the rules already had a conditional based on which version of HANA to use, so it needs to also check if we would like to create a new NSG.  If we would like the NSG to be created, and we actually want the rule because of the HANA version that we picked, then the resource gets created. @persiaAziz

PR #85, opened 2018-10-29
Can now distribute ssh keys to the hosts without storing locally at all

pabowers:
Previously, the public keys were distributed to the two databases, but this is more secure.

### DIFF
--- a/deploy/vm/ansible/roles/ssh-key-distribute/tasks/main.yml
+++ b/deploy/vm/ansible/roles/ssh-key-distribute/tasks/main.yml
@@ -1,24 +1,48 @@
-- name: Generate SSH keys
-  shell: ssh-keygen -b 2048 -t rsa -f /root/.ssh/id_rsa -q -N ""
-  args:
-    creates: /root/.ssh/id_rsa
+- name: Generate SSH key for current user
+  user:
+    name: root
+    generate_ssh_key: yes
+    ssh_key_bits: 2048
+  register: user_info
 
-- fetch:
-    src: /root/.ssh/id_rsa.pub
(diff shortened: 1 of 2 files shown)


deploy/vm/ansible/roles/ssh-key-distribute/tasks/main.yml
```diff
@@ -1,24 +1,48 @@
...
-- name: tell the host about our servers it might want to ssh to with ip addresses
-  shell: ssh-keyscan {{ hostvars[item]['ansible_default_ipv4']['address'] }} >> ~/.ssh/known_hosts
-  with_items: "{{ play_hosts }}"
+- debug:
+    var: host_key_rsa
+
+
+- name: tell the host about our servers it might want to ssh to with hostnames
+  known_hosts:
+    path: /root/.ssh/known_hosts
```
persiaAziz: root does not looks like a good idea

deploy/vm/ansible/roles/ssh-key-distribute/tasks/main.yml
```diff
@@ -1,24 +1,48 @@
-- name: Generate SSH keys
-  shell: ssh-keygen -b 2048 -t rsa -f /root/.ssh/id_rsa -q -N ""
-  args:
-    creates: /root/.ssh/id_rsa
+- name: Generate SSH key for current user
+  user:
+    name: root
```
persiaAziz: DO we want root?

deploy/vm/ansible/roles/ssh-key-distribute/tasks/main.yml
pabowers: Yes, if it isn't root, then the subsequent tasks will fail

PR #89, opened 2018-11-02
Allow user to choose between single/multi tenant DB install for HANA 1

persiaAziz:


deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
```diff
@@ -8,7 +9,16 @@
   args: 
     executable: '/bin/bash'
   register: check_hdbsql
-  when: use_hana2 == true or hana1_db_mode == "multiple_containers"
+  when: use_hana2 == true
```
tniek: I'm not sure I understand this change:
The SQL above backups up SYSTEMDB and DB tenant in an MDC system, i.e. HANA 1.0 (with MDC) or HANA 2.0 (explicitly MDC).
The condition you added, however, is only for HANA 2.0 - so what about HANA 1.0 installations with MDC?

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
```diff
@@ -8,7 +9,16 @@
...
+
+- name: Backup the database
+  shell: |
+    . ~/.bashrc
+    hdbsql -d SYSTEMDB -u SYSTEM -p "{{ pwd_db_system }}" -i {{ sap_instancenum }} "BACKUP DATA USING FILE ('initialbackupSYS')"
+  args:
+    executable: '/bin/bash'
+  register: check_hdbsql
+  when: use_hana2 == false and hana1_db_mode == "multiple_containers"
```
tniek: OTOH, this SQL gets executed for HANA 1.0 with MDC; why is the DB tenant not backed up as well?

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
```diff
@@ -1,5 +1,6 @@
 - debug:
         msg: "using hana2: {{use_hana2}} and hana1 db mode:{{hana1_db_mode}}"
+
 - name: Backup the database
```
tniek: The descriptions should specify which case you're covering (e.g. "Backup MDC DB (SYSTEMDB + DB tenant)")

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
persiaAziz: Because there is no DB tenant at this point. HANA 1.0 is only creating the SYSTEM DB.

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
persiaAziz: The later tasks cover the HANA1.0 scenarios

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
persiaAziz: Yes, this case handles a multi tenant DB mode with HANA 1.0. When I manually checked for the list of DBs after installing HANA 1.0 with MDC,, I just found SYSTEM DB and no other DB .

deploy/vm/ansible/roles/configure-hsr-hana/tasks/main.yml
persiaAziz: done

PR #90, opened 2018-11-09
Made the usage of pre-existing nsg more user friendly

pabowers:
The nsg-id is now really difficult to find in the portal.  This should make using a custom nsg much easier for the user

### DIFF
--- a/deploy/vm/modules/nsg_for_hana/variables.tf
+++ b/deploy/vm/modules/nsg_for_hana/variables.tf
@@ -1,7 +1,7 @@
 variable "az_region" {}
 
 variable "nsg_name" {
-  description = "The name of the nsg group to be created."
+  description = "The name of the NSG to be created."
 }
 
 variable "use_existing_nsg" {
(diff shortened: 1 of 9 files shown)


deploy/vm/modules/nsg_for_hana/variables.tf
```diff
@@ -1,8 +1,12 @@
 variable "az_region" {}
 
+variable "nsg_name" {
+  description = "The name of the nsg group to be created."
```
tniek: Nit: "NSG", "nsg group"

deploy/vm/modules/nsg_for_hana/variables.tf
```diff
@@ -13,10 +17,6 @@ variable "sap_instancenum" {
   description = "The sap instance number which is in range 00-99"
 }
 
-variable "sap_sid" {
-  default = "PV1"
-}
-
```
tniek: I believe this is an unrelated change?

pabowers: As I changed the nsg for hana variables to take a name of the nsg and the `sap_sid` was only needed for the name of the nsg, I thought that since it was no longer needed, it made sense to remove it in this PR

PR #91, opened 2018-11-12
fixing NSG not found bug

pabowers:
Terraform was trying to read the data before the Network security group was being created.  This weird workaround fixes this issue

### DIFF
--- a/deploy/vm/modules/common_setup/main.tf
+++ b/deploy/vm/modules/common_setup/main.tf
@@ -36,6 +36,6 @@ module "nsg" {
 }
 
 data "azurerm_network_security_group" "nsg_info" {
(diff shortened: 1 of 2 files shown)


deploy/vm/modules/common_setup/main.tf
```diff
@@ -36,6 +36,6 @@ module "nsg" {
 }
 
 data "azurerm_network_security_group" "nsg_info" {
-  name                = "${var.use_existing_nsg ? var.existing_nsg_name : local.new_nsg_name}"
```
persiaAziz: if we don't need local.new_nsg_name here, where are we using this?

pabowers: we use it to pass in the nsg name into the nsg module

PR #98, opened 2018-11-20
DO NOT MERGE: for review only

pabowers:
trying to figure out variables in Ansible


deploy/vm/ansible/roles/update-security-group/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    extend: "{{ allow_ips }}"
+  with_items: "{{ nsg_rules }}"
+  when: item['properties']['access'] != 'Deny' and not 'sourceAddressPrefix' in item.properties
+  register: new_rules_result
+
+- debug:
+    var: item.result
+  with_items: "{{ new_rules_result.results }}"
+- name: make a list of rules
+  set_fact: nsg_rules_updated="{{ new_rules_result.results | map(attribute='result') | to_json }}"
```
persiaAziz: Have you tried using azure_rm_security_group on the existing nsg to update using new rule? I see that you are using to update using the new list. Is it not possible to pass just this new ip to the rule for updating the nsg? One way to do this will be to put the IPs in a host group using add_hosts. Then iterate through the IPs and add them to the existing nsg.

deploy/vm/ansible/roles/delete-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
+---
+# defaults file for create-nsg
+allow_ips: "*"
```
persiaAziz: why does the delete nsg file contain all these rules?

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    action: extend
+    extend: "{{ allow_ips }}"
+  with_items: "{{ nsg_rules }}"
+  when: item['properties']['access'] != 'Deny' and not 'sourceAddressPrefix' in item.properties
+  register: new_rules_result
+
+- debug:
+    var: item.result
+  with_items: "{{ new_rules_result.results }}"
+- name: make a list of rules
```
persiaAziz: I am not really sure if ansible understands this. Have you tried working with "new_rules_result.results | map(attribute='result')" first and put it in a variable and then run to_json on that variable?

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
pabowers: I don't think it's possible to just add the ip to the rule. I need to know which rule to add the ip to, which is why I was going about it this way.

PR #102, opened 2018-11-29
 create n+m HANA cluster for scale out

persiaAziz:
For now, what this feature does is the following:

1. Creates resource group, vnet, security group
2. Create Network interface control for each VM
3. Creates n number of worker VMs
4. Attaches Managed disks to each of the worker VMs
5. Sets up n load balancers. Each worker VM is behind one internal load balancer. The load balancer is configured with high availability ports
6. Creates m number of stand by VMs

### DIFF
--- a/deploy/vm/ansible/create_n_m_hana_cluster.yml
+++ b/deploy/vm/ansible/create_n_m_hana_cluster.yml
@@ -1,8 +1,19 @@
 ---
-- name: "Create HANA cluster of {{ n }} VMs"
+- name: prepare configs
+  hosts: localhost
+  tasks:
+     - set_fact:
+          m_count: "1"
+       when: m < 0
+     - set_fact:
+          m_count: "{{ m }}"
+       when: m > 0
+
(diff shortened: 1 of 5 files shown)


deploy/vm/ansible/create_n_m_hana_cluster.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: prepare configs
+  hosts: localhost
+  tasks:
+     - name: This is a hack to get the iteration functionality of Ansible working
+       set_fact:
+          m_count: "1"
+       when: m < 1
```
pabowers: I'm a little confused by what you are doing with this logic.  Why would `m` ever be less than 1?

persiaAziz: in case there is no standby then m will be 0, which is a problem for the Ansible loop (with_sequence). Ansible loop is not smart enough to understand that the end condition is met when start =1 and end =0 (m), hence it throws an exception during evaluation. Hence this hack

PR #103, opened 2018-11-29
consolidate url_sap_sapcar and url_sapcar_linux

persiaAziz:


persiaAziz:
In that case, for the sake of consistency, do we want to change url_sapcar_windows to url_sap_sapcar_windows?

tniek:
For consistency, probably better to name it "url_sap_sapcar_linux"

persiaAziz:
Done @tniek

tniek:
Looks good, thanks

PR #107, opened 2018-12-07
Allow custom ip range

pabowers:
This lets you specify the ip addresses that will be allowed to connect to the VMs


deploy/vm/ansible/roles/create-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    protocol: Tcp
+    priority: 105
+    destination_port_range: "4000-4999"
+    source_address_prefix: "{{ allow_ips }}"
+  - name: XSA
+    protocol: Tcp
+    priority: 106
+    destination_port_range: "50000-59999"
+    source_address_prefix: "{{ allow_ips }}"
+
```
tniek: This is not accurate; hana2_rules should really be "hana_xsa_rules" (since it's independent of HANA1.0/2.0), since it opens ports for XSA/web dispatcher. hana1_rules are really the rules for installations with XS Classic (predecessor to XSA).

deploy/vm/ansible/roles/delete-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    destination_port_range: "80{{ sap_instancenum }}"
+    source_address_prefix: "{{ allow_ips }}"
+  - name: HTTPS
+    protocol: Tcp
+    priority: 104
+    destination_port_range: "43{{ sap_instancenum }}"
+    source_address_prefix: "{{ allow_ips }}"
+
+hana2_rules:
+  - name: XSA-HTTP
```
tniek: See comment from above for hana1_rules/hana2_rules

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
+- name: Get info about the security group that you had created
+  azure_rm_securitygroup_facts:
+    name: "{{ hana_nsg_name }}"
+    resource_group: "{{ resource_group }}"
+  register: nsg_info
+
+- name: Get the current NSG rules
+  set_fact:
+    nsg_rules: "{{ nsg_info['ansible_facts']['azure_securitygroups'][0]['properties']['securityRules'] }}"
```
tniek: What about the case that there are no azure_securitygroups (for whatever reason) -> index out of range?

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+- name: Create a list of NSG rules when the rule is an Allow rule
+  set_fact:
+    nsg_allow_rule: "{{ item.properties.destinationPortRange }}"
+  register: nsg_allow_rule_obj
+  when: item.properties.access == "Allow" and item.properties.destinationPortRange is defined
+  with_items: "{{ nsg_rules }}"
+
+- name: Create list of Ports to allow in new rules
+  set_fact:
+    dest_port_range="{{ nsg_allow_rule_obj.results | selectattr('ansible_facts', 'defined') | selectattr('ansible_facts.nsg_allow_rule', 'defined') | map(attribute='ansible_facts.nsg_allow_rule') | list }}"
```
tniek: I trust this will work, but can we make it more readable?
What if one of those functions returns None?

deploy/vm/modules/single_node_hana/single-node-hana.tf
```diff
@@ -2,7 +2,9 @@
 provider "azurerm" {}
 
 module "common_setup" {
-  source            = "../common_setup"
+  source = "../common_setup"
```
tniek: nit: indent w/ tab

deploy/vm/modules/single_node_hana/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "allow_ips" {
+  description = "The ip addresses that will be allowed by the nsg"
```
tniek: nit: write "The IP addresses"

deploy/vm/modules/windows_bastion_host/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "allow_ips" {
+  description = "The ip addresses that will be allowed by the nsg"
```
tniek: nit: "IP addresses"

deploy/vm/modules/windows_bastion_host/variables.tf
```diff
@@ -36,6 +41,7 @@ variable "windows_bastion" {
 }
 
 locals {
+  all_ips      = "*"
```
tniek: Why is this one not escaped as in variables.tf? I.e.,
all_ips      = "\\\"*\\\""

deploy/vm/ansible/roles/delete-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    priority: 105
+    destination_port_range: "4000-4999"
+    source_address_prefix: "{{ allow_ips }}"
+  - name: XSA
+    protocol: Tcp
+    priority: 106
+    destination_port_range: "50000-59999"
+    source_address_prefix: "{{ allow_ips }}"
+
+hana_nsg_rules: "{{ hana_nsg_rules_common }} + {{ hana2_rules if use_hana2 else hana1_rules }}"
```
tniek: This should be...
... + {{ hana_xsa_rules if install_xsa else hana_xsc_rules }}

deploy/vm/ansible/roles/create-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    protocol: Tcp
+    priority: 102
+    destination_port_range: "3{{ sap_instancenum }}00-3{{ sap_instancenum }}99"
+    source_address_prefix: "{{ allow_ips }}"
+  - name: deny-all-inbound
+    priority: 600
+    access: Deny
+
+hana1_rules:
+  - name: HTTP
```
tniek: hana1_rules should really be hana_xsc_rules (see my next comment on hana2_rules for details)

deploy/vm/ansible/roles/delete-nsg/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    protocol: Tcp
+    priority: 102
+    destination_port_range: "3{{ sap_instancenum }}00-3{{ sap_instancenum }}99"
+    source_address_prefix: "{{ allow_ips }}"
+  - name: deny-all-inbound
+    priority: 600
+    access: Deny
+
+hana1_rules:
+  - name: HTTP
```
tniek: hana1_rules should be hana_xsc_rules, see above

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
pabowers: you have to pass one in, or one is created for you

deploy/vm/modules/single_node_hana/single-node-hana.tf
pabowers: that was autoformatted by terraform fmt, so I removed the empty line after it to fix the tab

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
pabowers: I had an issue with that earlier, that is why I am using the function `selectattr()` with the requirement that the key is `defined`.  This will avoid issues if one of the functions returns None

pabowers:
@tniek I believe that I addressed your comments. Can you please let me know if you would like more changes before I can merge in this PR?

deploy/vm/modules/ha_pair/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "allow_ips" {
```
persiaAziz: Can you please add this to the list of user input variables in the README?

persiaAziz:
overall looks goo d to me. Please ad the nre config to the README. Have you tested this with a fresh Ubuntu VM? I think we should test big features like this in a newly created VM to determine if the feature adds a new requirement.

deploy/vm/modules/single_node_hana/variables.tf
```diff
@@ -1,3 +1,8 @@
+variable "allow_ips" {
+  description = "The IP addresses that will be allowed by the nsg"
+  default     = []
```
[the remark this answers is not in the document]
pabowers: If you don't pass one, it opens to all.  I know that it is hacky, but you can't simply put a * in the list.  It won't work.  For type consistency it needed to be a list.

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
tniek: My point was that the five concatenated functions in a single line are hard to read.
Can we make it more readable?

tniek:
My only comment would be regarding line 20 of update-security-group/tasks/main.yml (see above). Other than that, looks okay to me.

deploy/vm/modules/ha_pair/variables.tf
pabowers: done

deploy/vm/ansible/roles/update-security-group/tasks/main.yml
pabowers: unfortunately it isn't possible.  I wish there were a way to make it look cleaner

tniek: Ok.

PR #110, opened 2018-12-13
Pabowers hapair add timesync

pabowers:
Although the errors are happening infrequently, this PR should fix the occasional issue that we get with the HA pair deployment where NTP is not installed.  According to SUSE, this is a requirement for the HA cluster even though the failure is not consistent.

### DIFF
--- a/deploy/vm/ansible/requirements.yml
+++ b/deploy/vm/ansible/requirements.yml
@@ -1,0 +1,4 @@
+
+- name: Install NTP ansible role
+  src: geerlingguy.ntp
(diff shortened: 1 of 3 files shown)


deploy/vm/ansible/requirements.yml
```diff
@@ -0,0 +1,3 @@
+
+- src: geerlingguy.ntp
```
tniek: This appears to be a good choice for the NTP support and the repo is regularly maintained.

PR #111, opened 2018-12-14
Retrieve azure service principal from the exported environment variable

persiaAziz:
This PR removes the azure_service_principal_id and azure_service_principal_pw. The code retrieves these information from the environment variable. This PR also introduces a configuration check script. For future validation checking we can extend this script


deploy/vm/ConfigurationCheck/configCheck.py
```diff
@@ -0,0 +1,17 @@
+import os
+import sys,errno
+
```
pabowers: Did we get this to reliably work?

deploy/vm/modules/common_setup/main.tf
```diff
@@ -1,7 +1,15 @@
 # Create a resource group.
+
+resource null_resource "configuration-check" {
```
pabowers: This looks like it should work, but I just wanted to make sure that it reliably fails

persiaAziz: It does. It fails early in the deployment stage when the some module calls the common-setup module which is dependent on configuration-check.

PR #113, opened 2018-12-14
correct the task name of removing a VM

persiaAziz:


deploy/vm/ansible/roles/delete-azure-vm/tasks/main.yml
```diff
@@ -1,4 +1,4 @@
-- name: remove vm and all resources except public ips
+- name: remove vm and all resources
```
pabowers: can you change it to `remove vm and all associated resources`?
I think this might be a little clearer since the resource group, subnet, and vnet should still exist after executing this command

PR #118, opened 2018-12-28
Failover test for HA pair

persiaAziz:


deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,35 @@
...
+         name: "{{ fail_hana_vm }}"
+         group: fail_hana_instance
+- hosts: "fail_hana_instance"
+  tasks:
+    - iptables:
+         chain: INPUT
+         source: "{{ db_host_name }}"
+         jump: DROP
+      become: yes
+
```
tniek: As discussed, not sure if this resembles a valid test case; the following page lists a comprehensive set of scenarios we could use here:

https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#test-the-cluster-setup

deploy/vm/test/ha_pair_failover_test/test_db_connection.py
```diff
@@ -0,0 +1,8 @@
+import pyhdb
+import sys
+
+connection = pyhdb.connect(host=sys.argv[1],port=sys.argv[2],user=sys.argv[3],password=sys.argv[4])
+cursor = connection.cursor()
+cursor.execute("SELECT HOST AS CURRENT_MASTER, CURRENT_TIME FROM M_CONNECTIONS WHERE CONNECTION_ID = CURRENT_CONNECTION")
+print(cursor.fetchone())
```
tniek: You're not doing anything with the result?
What about error handling (e.g. no connection to HDB can be established)?

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: sounds good. Creating work items :)

deploy/vm/test/ha_pair_failover_test/test_db_connection.py
persiaAziz: hmm yes some error handling can be done to exit with certain error code

tniek:
I would suggest to add rudimentary error handling (check connection didn't fail/time out; result from SQL matches hostname) to this PR; the different test scenarios can be added eventually

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+  tasks:
+    - copy:
+          src: parse_landscapeconfig.py
+          dest: /tmp/
+    - shell: |
+          . ~/.bashrc
+          HDB stop
+    - include_role: 
+         name: check-hdb-status
+      loop: [1, 2, 3]
```
pabowers: Why do you need to specifically pass in this list? [This](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-sequence) is the recommended way to achieve this behavior

deploy/vm/test/ha_pair_failover_test/parse_landscapeconfig.py
```diff
@@ -0,0 +1,35 @@
+'''
+Format to parse:
```
pabowers: Please provide unit tests with expected inputs and outputs

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+      copy:
+         src: test_db_connection.py
+         dest: /tmp
+    - name: Test the db connection
+      command: "python /tmp/test_db_connection.py {{ db_host_name }} {{ db_host_port }} {{ db_user }} {{ db_pwd }}"
+      register: testout
+    - debug:
+         msg: "{{ testout.stdout }}"
+    - fail:
+         msg: "{{ master_hdb }} is not the primary"
```
pabowers: Please include a description in the failure message of why it should fail if `master_hdb` is not the primary.  What is the expected state here?

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+      become: yes
+      become_method: sudo
+    - pip:
+         name: pip
+         state: latest
+      become: yes
+      become_method: sudo
+    - pip:
+         name: pyhdb
+         state: latest
```
pabowers: We should tag a specific version number so our script will not fail if pyhdb gets updated

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+  hosts: jumpbox_vm
+  tasks: 
+    - package:
+         name: python-pip
+         state: present
+      become: yes
+      become_method: sudo
+    - pip:
+         name: pip
+         state: latest
```
pabowers: Tag a specific version.  We will probably want latest, but don't want the user to run into issues if an update breaks this without us testing first

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+      become: yes
+      become_method: sudo
+    - pip:
+         name: pyhdb
+         state: latest
+      become: yes
+      become_method: sudo
```
pabowers: Is sudo needed to execute the python script?

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
```diff
@@ -0,0 +1,82 @@
...
+
+- hosts: "jumpbox_vm"
+  tasks: 
+    - name: Test the db connection again
+      command: "python /tmp/test_db_connection.py {{ db_host_name }} {{ db_host_port }} {{ db_user }} {{ db_pwd }}"
+      register: testout
+    - debug:
+         msg: "{{ testout.stdout }}"
+    - fail:
+         msg: "{{ slave_hdb }} should be the primary after the failover but it is not"
```
pabowers: I like the specificity of the error message.  Use something like this in the comment I had above.

deploy/vm/test/ha_pair_failover_test/roles/check-hdb-status/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+- pause:
```
pabowers: nit: name the task and explain why the pause is needed

deploy/vm/test/ha_pair_failover_test/roles/check-hdb-status/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+     minutes: 2
+- name: Run the landscapeconfig script to get the configuration table 
+  shell: . ~/.bashrc && python /usr/sap/"{{ sap_sid }}"/HDB"{{ instance_num }}"/exe/python_support/landscapeHostConfiguration.py > /tmp/configOut.txt
+  ignore_errors: True
```
pabowers: Why are we ignoring errors? It seems like a test should not ignore errors

deploy/vm/test/ha_pair_failover_test/test_db_connection.py
```diff
@@ -0,0 +1,12 @@
+import pyhdb
```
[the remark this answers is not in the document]
pabowers: ok, that will work.  Perhaps have two excepts.  One that excepts incorrect input error and the other that will display that the connection to the database cannot be established.  If it is failing due to an incorrect input, the error message "Connection to the database could not be established" is not particularly helpful to the user.

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: The HDB takes time to come up after a failover.  So the first run might not find the HDB to be up. If the status does is not active after all the status check runs (~6 moinutes) then that is not expected and the test fails.

deploy/vm/test/ha_pair_failover_test/parse_landscapeconfig.py
persiaAziz: the format is in the script already

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: If the user inputs a master_hdb which actuallyis not the primary, then this is an error.

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: latest is what we want here

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: this is installing pyhdb and not running any python script. Hence Yes

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
persiaAziz: see my response above

deploy/vm/test/ha_pair_failover_test/roles/check-hdb-status/tasks/main.yml
persiaAziz: see my response above

deploy/vm/test/ha_pair_failover_test/roles/check-hdb-status/tasks/main.yml
persiaAziz: already explained above

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
pabowers: I misread.  Sorry, I thought it was on the whole set of tasks

deploy/vm/test/ha_pair_failover_test/playbook_failover_test.yml
pabowers: I understand wanting to check several times, it was more an ansible specific comment.  Please explain in person.

persiaAziz:
I have renamed all the master/slave to primary/secondary @pabowers 

The PR is blocked on you @tniek . Please review again

tniek:
> I have renamed all the master/slave to primary/secondary @pabowers
> 
> The PR is blocked on you @tniek . Please review again

Looks good, approved

PR #119, opened 2019-01-03
Install required packages for azure python sdk

persiaAziz:


deploy/vm/ansible/roles/install-python-sdk/tasks/requirements.txt
```diff
@@ -0,0 +1,7 @@
+cffi>=1.4.1
```
pabowers: I wonder if this creates the file with the name "=1.4.1" on the machine.  Can you confirm that this doesn't have this behavior?

deploy/vm/ansible/roles/install-python-sdk/tasks/requirements.txt
```diff
@@ -0,0 +1,7 @@
...
+adal==1.2.0
+azure>=4.0
```
pabowers: same question as above

deploy/vm/ansible/roles/install-python-sdk/tasks/requirements.txt
persiaAziz: this is being used by the pip module, not shell. So this will not happen. I have also confirmed that this does not happen

PR #126, opened 2019-01-08
Moved Windows Bastion host creation to Ansible

pabowers:
Would like feedback before moving other pieces to Ansible


deploy/vm/modules/playbook-execution/main.tf
```diff
@@ -35,6 +35,13 @@ resource null_resource "mount-disks-and-configure-hana" {
      \"pwd_db_tenant\": \"${var.pwd_db_tenant}\", \
      \"pwd_db_shine\": \"${var.pwd_db_shine}\", \
      \"email_shine\": \"${var.email_shine}\", \
+     \"vnet_name\": \"${var.vnet_name}\", \
```
persiaAziz: I still feel like this passing should be consistent with what we are doing for Linux bastion host, i.e., read from the json file

deploy/vm/modules/playbook-execution/variables.tf
```diff
@@ -169,3 +169,40 @@ variable "linux_bastion" {
   description = "flag to determine if linux bastion host is needed or not"
   default     = false
 }
+
+variable "vnet_name" {
```
persiaAziz: yes this is the thing I suggest to avoid. Since we are moving to Ansible, adding these variables to terraform defeats the purpose. You will have to do it eventually as a part of the whole moving to ansible project.

deploy/vm/modules/playbook-execution/main.tf
pabowers: Ok, will look into doing it that way.  Should it all be part of the same JSON file?

deploy/vm/modules/playbook-execution/variables.tf
persiaAziz: As discussed, I think having a completely separate playbook for bastion host is useful in case someone wants to have a bastion host later. But having separate playbook requires the azure specific information to be accessible somehow. Hence I prefer storing the information in a file.

PR #134, opened 2019-01-25
Fixing the NSG dependency bug

pabowers:
In making this change, I moved the NSG creation back into Terraform so that the state is not split across Ansible and Terraform.


deploy/vm/modules/common_setup/variables.tf
```diff
@@ -38,6 +38,18 @@ variable "sap_sid" {
 }
 
 locals {
+  all_ips      = "*"
```
persiaAziz: Do we still need this now? Can't we just keep it a list with 0.0.0.0/0? That will be consistent everywhere

pabowers: Fixed

PR #144, opened 2019-03-05
pass config as a parameter to hwcct v1 test

persiaAziz:


deploy/vm/test/hwcct/hwcct_v1_playbook.yml
```diff
@@ -27,14 +27,21 @@
                 creates: /hana/shared/hwcct/
           - name: copy config file
             copy:
-                src: hwcct_config.json
+                src: "{{ test_config }}"
```
pabowers: Since you removed the json file, can you please include an example of what the `test_config` file might look like?

persiaAziz: sure I will do that, once I right a documentation for this hwcct.

PR #145, opened 2019-03-05
pass output file as argument to the parser

persiaAziz:
(cherry picked from commit 5d3392a68ae699f75fbddf553f1199c828683972)


deploy/vm/test/hwcct/hwcct_parser.py
```diff
@@ -99,7 +99,7 @@ def skip_until(string, untilString, f_iter):
         string = next(f_iter)
     return string
 
-def parseDISKTest(logfile):
+def parseDISKTest(logfile, outputfile):
```
pabowers: Why is disk capitalized? It seems inconsistent

PR #151, opened 2019-05-31
Add contributing guidelines

tniek:
Initial version

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+Contributing Guidelines
+=======================
+
+Thanks for taking the time to contribute!
+
+This document summarizes the deployment principles which govern our project.
+
+Building "in the Open"
+----------------------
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+----------------------
+* Our users appreciate the transparency of our project; in this context, building "in the open" means that anyone can:
+  * see the individual pull requests this solution is comprised of;
+  * understand particular changes by going back and forth between pull requests;
+  * submit pull requests with their own changes. 
+* In addition, we use open-source software (OSS) tools rather than proprietrary technology to build the solution.
+
+Fully Maintained
+----------------
+* Our project is not an as-is tool that may or may not work; instead we strive to:
```
abhinababasu: The license is as-is. However, we want to call out it is not some sample that just "works on the dev's machine". Rather a tested solution is run every day

PR #152, opened 2019-06-05
Initial version of SapMonitor payload

tniek:
In its initial version, sapmon.py fulfills two main functions:

(1) init: first-time setup; store credentials in customer KeyVault
    (to be called only once upon VM deployment via script extension)
(2) monitor: actual payload; retrieve metrics from registeered HDB
    and emit to LogAnalytics (to be called periodically as cronjob)

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -10,10 +10,7 @@ import requests, json
 import sys, argparse
 import datetime, decimal
 import hashlib, hmac, base64
-try:
-   import http.client as http_client  # Python 3
-except ImportError:
-   import httplib as http_client      # Python 2
+import http.client as http_client
 import logging
 
 ###############################################################################
@@ -108,7 +105,7 @@ class REST:
       if response.status_code == requests.codes.ok:
          contentType = response.headers.get("content-type")
          if contentType and contentType.find("json") >= 0:
-            return json.loads(response.content)
+            return json.loads(response.content.decode("utf-8"))
          else:
             return response.content
       else:
@@ -232,14 +229,15 @@ class AzureLogAnalytics:
 application/json
 x-ms-date:%s
 /api/logs""" % (len(content), timestamp)
-         bytesHash   = bytes(stringHash).encode("utf-8")
+         bytesHash   = bytes(stringHash, encoding="utf-8")
          decodedKey  = base64.b64decode(self.sharedKey)
          encodedHash = base64.b64encode(hmac.new(
             decodedKey,
             bytesHash,
             digestmod=hashlib.sha256).digest()
          )
-         return "SharedKey %s:%s" % (self.workspaceId, encodedHash)
+         stringHash  = encodedHash.decode("utf-8")
+         return "SharedKey %s:%s" % (self.workspaceId, stringHash)
 
       timestamp   = datetime.datetime.utcnow().strftime("%a, %d %b %Y %H:%M:%S GMT")
       headers = {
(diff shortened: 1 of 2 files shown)


monitor/sapmon.py
```diff
@@ -0,0 +1,366 @@
+import pyhdb
```
pabowers: you should probably have a copyright header

monitor/sapmon.py
```diff
@@ -0,0 +1,366 @@
...
+         data    = data,
+         )
+      if response.status_code == requests.codes.ok:
+         contentType = response.headers.get("content-type")
+         if contentType and contentType.find("json") >= 0:
+            return json.loads(response.content)
+         else:
+            return response.content
+      else:
+         print response.content
```
pabowers: please put in parenthesis to support python3

monitor/sapmon.py
```diff
@@ -0,0 +1,366 @@
...
+###############################################################################
+
+def init(args):
+   """
+   Store credentials in the customer KeyVault.
+   (To be executed as custom script upon initial deployment of collector VM.)
+   """
+   # Credentials (provided by user) to the existing HANA Dinstance
+   hanaSecretName  = "SapHana-%s" % args.HanaDbName
+   hanaSecretValue = json.dumps({
```
pabowers: ensure that this doesn't log the secret values to the console

monitor/sapmon.py
```diff
@@ -0,0 +1,366 @@
...
+      "HanaDbUsername": args.HanaDbUsername,
+      "HanaDbPassword": args.HanaDbPassword,
+      "HanaDbSqlPort":  args.HanaDbSqlPort,
+      })
+   ctx.azKv.setSecret(hanaSecretName, hanaSecretValue)
+
+   # Credentials (created by HanaRP) to the newly created Log Analytics Workspace
+   laSecretName  = "AzureLogAnalytics"
+   laSecretValue = json.dumps({
+      "LogAnalyticsWorkspaceId": args.LogAnalyticsWorkspaceId,
```
pabowers: nit: ID

monitor/sapmon.py
tniek: This was developed using the standard SLES4SAP 12.3 image (which ships with Python 2.7) but I'll need to make it Python 3-compatible if we decide to go with a different image.

monitor/sapmon.py
tniek: There's no proper logging yet; the only way traffic between collector VM <-> HANA instance gets written out is if you manually force it by setting debug=True for REST.sendRequest().

monitor/sapmon.py
tniek: Camel casing; if applied consequently, "Sap" would have to be "SAP" and "Hana" would have to be "HANA" (actually both the correct spelling).

monitor/sapmon.py
tniek: This wasn't the highest priority for this prototype and should be governed by our repo's policy (https://github.com/Azure/sap-hana#license--copyright).

monitor/sapmon.py
pabowers: not necessarily, as a team, there was a movement a while ago to use ID uppercase regardless of other camelcasing. It might just be for go, but I wanted to be consistent with those guidelines. That said, I can't seem to find it in the onenote

monitor/sapmon.py
pabowers: not the biggest deal, but thought it would be a good idea since the files are open source

monitor/sapmon.py
PakDLiu: My understanding was that is only for Go

monitor/sapmon.py
tniek: Added copyright header

monitor/sapmon.py
tniek: Fixed compatibility issues with Python 3

monitor/sapmon.py
tniek: Kept it as-is for now

monitor/sapmon.py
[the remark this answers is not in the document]
tniek: Done

PR #153, opened 2019-06-05
Configuration file to configure Monitor VM

persiaAziz:


tniek:
Thanks, looks good to me

PR #156, opened 2019-06-11
Logic to pull HANA load history since last run

tniek:
* Read/write last successful pull to state file
* Normalize server timestamps to UTC

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -10,8 +10,7 @@ import requests, json
 import sys, argparse
 import datetime, decimal
 import hashlib, hmac, base64
-import http.client as http_client
-import logging
+import logging, http.client as http_client
 
 ###############################################################################
 
@@ -264,7 +263,7 @@ class _Context:
    def __init__(self):
       self.vmInstance = AzureInstanceMetadataService.getComputeInstance()
       vmTags          = dict(map(lambda s : s.split(':'), self.vmInstance["tags"].split(";")))
-      self.sapmonId   = vmTags["sapmonId"]
+      self.sapmonId   = vmTags["SapMonId"]
       self.azKv       = AzureKeyVault("sapmon-%s" % self.sapmonId)
 
    def parseSecrets(self):
@@ -354,7 +353,8 @@ def monitor(args):
       
 def main():
    parser     = argparse.ArgumentParser(description="SAP on Azure Monitor Payload")
-   subparsers = parser.add_subparsers(help="main functions")
+   subparsers = parser.add_subparsers(dest="command", help="main functions")
+   subparsers.required = True
    initparser = subparsers.add_parser("init", help="Initialize client KeyVault with SAP credentials")
    initparser.set_defaults(func=init)
    initparser.add_argument("HanaHostname", type=str, help="Hostname of the HDB to be monitored")
@@ -367,12 +367,7 @@ def main():
    monparser  = subparsers.add_parser("monitor", help="Execute the monitoring payload")
    monparser.set_defaults(func=monitor)
    args       = parser.parse_args()
-   try: # see https://bugs.python.org/issue16308
-      f = getattr(args, "func")
-   except AttributeError:
-      parser.print_help()
-      sys.exit(0)
-   f(args)
+   args.func(args)
 
 ctx = _Context()
 if __name__ == "__main__":
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -8,9 +8,16 @@
 import pyhdb
 import requests, json
 import sys, argparse
-import datetime, decimal
+import decimal
+from datetime import datetime, timedelta, date
 import hashlib, hmac, base64
 import logging, http.client as http_client
+
+###############################################################################
+
+STATE_FILE                   = "sapmon.state"
+INITIAL_LOADHISTORY_TIMESPAN = -(60 * 1)
+LOG_TYPE                     = "SapHana_Infra"
 
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -340,16 +384,29 @@ def monitor(args):
...
+      lastPull = resultRows[-1][colIndex["UTC_TIMESTAMP"]]
       logData = []
       for r in resultRows:
          logItem = {}
          for c in colIndex.keys():
+            if c.startswith("_"): # remove internal fields
+               continue
             logItem[c] = r[colIndex[c]]
+         jsonData = json.dumps(logItem, sort_keys=True, indent=4, cls=_JsonEncoder)
+         # isoTime  = (r[colIndex["UTC_TIMESTAMP"]]).isoformat()
```
persiaAziz: are these commented code needed?

monitor/sapmon.py
```diff
@@ -265,6 +281,34 @@ def __init__(self):
...
+
+   def readLastPullTimestamp(self):
+      """
+      Read the timestamp (UTC) of the last successful pull from state file
+      """
+      lastPull = None
+      try:
+         with open(STATE_FILE, "r") as f:
+            data = json.load(f)
+         lastPull = datetime.strptime(data["lastPullUTC"], "%Y-%m-%d %H:%M:%S.%f")
```
persiaAziz: Please put the key lastPullUTC in a global constant/static variable and use the variable wherever needed

monitor/sapmon.py
```diff
@@ -218,9 +232,9 @@ def __init__(self, workspaceId, sharedKey):
       self.sharedKey   = sharedKey
       self.uri         = "https://%s.ods.opinsights.azure.com/api/logs?api-version=2016-04-01" % workspaceId
 
-   def ingest(self, logType, jsonData):
+   def ingest(self, logType, jsonData, timeGenerated = None):
```
pabowers: I know that Python doesn't require types, but what type are we expecting here? Can we have a comment so it is clear to future people working on this?

monitor/sapmon.py
```diff
@@ -265,6 +281,34 @@ def __init__(self):
...
+      finally:
+         return lastPull
+
+   def setLastPullTimestamp(self, timestamp):
+      """
+      Write the timestamp (UTC) of the last successful pull to state file
+      """
+      try:
+         data = {
+            "lastPullUTC": timestamp.strftime("%Y-%m-%d %H:%M:%S.%f")
```
pabowers: This format string is used several places in this file. Please make it a constant somewhere so that in case we decide to change it, it only needs to be updated in one place.

monitor/sapmon.py
```diff
@@ -340,16 +384,29 @@ def monitor(args):
    ctx.parseSecrets()
    for h in ctx.hanaInstances:
       h.connect()
-      fromTimestamp  = datetime.datetime.now() - datetime.timedelta(minutes=1)
+      if not ctx.lastPull:
+         fromTimestamp = None
+      else:
+         fromTimestamp  = ctx.lastPull + timedelta(seconds=1) 
       colIndex, resultRows = h.getLoadHistory(fromTimestamp)
+      lastPull = resultRows[-1][colIndex["UTC_TIMESTAMP"]]
```
pabowers: Although it is unlikely, we should handle index out of range error and log in that case

monitor/sapmon.py
pabowers: Please remove commented out code

tniek: As discussed, I removed the option to overwrite the TimeGenerated field in Log Analytics as it would require one-by-one ingestion (rather than batch ingestion) of metrics.

monitor/sapmon.py
tniek: Following this suggestion, other variables pertinent to a specific execution (KeyVault, VM tags and others) would have to be global as well; I consider the _Context class (which currently holds these variables) a cleaner approach and it also provides convenient access to all variables.

monitor/sapmon.py
tniek: We could introduce (more detailed) docstrings, but I don't see that in the Go code either. I'll note this down as a to-do, although this particular instance is resolved since I removed the ability to overwrite the TimeGenerated field in Log Analytics (see above).

monitor/sapmon.py
tniek: Totally agree. Proper error handling is still missing and will be added (together with proper logging) once we complete the PoC.

monitor/sapmon.py
tniek: strftime() is used several times in the payload; in different contexts, the format is actually different (e.g. HANA expecting a different format in a SQL statement than the Log Analytics REST API). That said, I created different templates for the specific purpose.

monitor/sapmon.py
pabowers: Isn't this just `TIME_FORMAT_HANA`, why don't you simply use that here?

monitor/sapmon.py
pabowers: since you aren't adding it here, can you please add a TODO comment with a workitem number indicating when it is expected to be done?

monitor/sapmon.py
tniek: I believe we're talking about different things here:
Presumably, Persia was referring to the actual timestamp of the last successful pull; you were referring to the time format template passed into strftime/strptime. For the latter, I'm already using templates (see commit 1aa514f).

tniek:
@pabowers In commit 1aa514f, I added time templates for HANA/Log Analytics.

PR #162, opened 2019-06-24
Adding our team to be automatically added for reviews

pabowers:
Trying to make the review process easier by automatically requesting reviewers. Not 100% sure how to test this. I followed the instructions provided by GitHub https://help.github.com/en/articles/about-code-owners#codeowners-file-location. If you have written something like this in the past, let me know.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# This is to automatically notify people on our team when a new PR is opened
+*	@Azure/azure-dedicated-team


CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# This is to automatically notify people on our team when a new PR is opened
+*	@Azure/azure-dedicated-team
```
jeffaco: I'm not sure this is going to work. According to the documentation that I read, CODEOWNERS need to have write access to the repository. However, that GitHub team won't be logging in, as such.

Do groups work in this context? Do you have examples where this works?

jeffaco: I take this back. I can take back, right? The link you gave (which I didn't find when doing Google searches) specifically mentioned that teams worked.

However, it also mentioned this can be done on a per-branch basis. Thus, you could fully test it in a branch, if you wanted. But based on what I read on second pass, this is good. Signing off.

PR #164, opened 2019-06-27
Set up CI with Azure Pipelines

persiaAziz:
This is a simple version of the pipeline code. This will grow more in the later iterations. Right now enabling full deploy is not secure


azure-pipelines.yml
```diff
@@ -0,0 +1,43 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
```
tniek: Maybe need to re-write this paragraph, since it's a public repo.

azure-pipelines.yml
```diff
@@ -0,0 +1,43 @@
...
+    vmImage: "ubuntu-16.04"
+  displayName: "Test deployment of single node hana"
+  steps:
+  - checkout: self  # self represents the repo where the initial Pipelines YAML file was found
+    clean: "all"
+    path: "sap-hana"  # path to check out source code, relative to the agent's build directory
+    persistCredentials: boolean  # set to 'true' to leave the OAuth token in the Git config after the initial fetch
+  - script: |
+      ssh-keygen -b 4096 -t rsa -f /tmp/sshkey -q -N ""
+      cp deploy/vm/test/terraform.tfvars deploy/vm/modules/single_node_hana
```
tniek: This is ok for the current version of our templates, but this structure will change (i.e. one single TF module as entry point, and one JSON config for both TF and Ansible). Would it make sense to abstract this and make it more general?

azure-pipelines.yml
```diff
@@ -0,0 +1,43 @@
...
+      TF_VAR_pw_os_sapadm: $(hana-int-pw-sap-adm)
+      TF_VAR_pw_hacluster: $(hana-int-pw-sap-adm)
+      AZURE_CLIENT_ID: $(hana-pipeline-spn-id)
+      AZURE_SECRET: $(hana-pipeline-spn-pw)
+      AZURE_TENANT: $(landscape-tenant)
+      AZURE_SUBSCRIPTION_ID: $(landscape-subscription)
+      ARM_CLIENT_ID: $(hana-pipeline-spn-id)
+      ARM_CLIENT_SECRET: $(hana-pipeline-spn-pw)
+      ARM_TENANT_ID: $(landscape-tenant)
+      ARM_SUBSCRIPTION_ID: $(landscape-subscription)
```
tniek: nit: newline at EOF

deploy/vm/test/terraform.tfvars
```diff
@@ -0,0 +1,11 @@
+az_region = "westus"
+vm_user = "azureuser"
+az_domain_name = "singlehana"
```
tniek: Probably need to go for something more "unique" to avoid deployment failures

azure-pipelines.yml
```diff
@@ -0,0 +1,43 @@
...
+# https://aka.ms/yaml
+
+jobs:
+- job: "single_node_hana"
+  variables:
+    - group: azure-config-variables
+    - group: azure-sap-hana-pipeline-secrets
+  pool:
+    vmImage: "ubuntu-16.04"
+  displayName: "Test deployment of single node hana"
```
tniek: nit: "single-node HANA"

azure-pipelines.yml
persiaAziz: Yah this is the default headlines that azure pipelines put. I forgot to change that.

azure-pipelines.yml
persiaAziz: Our current version is dependent on tfvars. I will need the exact format and schema for the new structure. I think we should keep it like this for our version. We can change accordingly when the newer version arrives.

deploy/vm/test/terraform.tfvars
persiaAziz: good catch

PR #169, opened 2019-06-28
[Hotfix] pip upgrade failing the build

persiaAziz:


lagalbra:
The PR description should include the impact of the break necessitating the fix, and why this fixes it.

PR #179, opened 2019-07-08
Add HDB connectivity check to onboarding payload

tniek:
- Validate provided HDB credentials by connecting and running a dummy query

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -363,7 +363,7 @@ def onboard(args):
    Store credentials in the customer KeyVault.
    (To be executed as custom script upon initial deployment of collector VM.)
    """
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -382,6 +382,13 @@ def onboard(args):
       })
    ctx.azKv.setSecret(laSecretName, laSecretValue)
 
+   # Check connectivity to HANA instance
```
persiaAziz: Eventually, having an exception handler that returns a well composed error message will be nice

tniek: Yep, will need to add that to the logging/error handling efforts.

PR #180, opened 2019-07-08
HOTFIX: Explicitly made context global (python3)

tniek:
- As CSE uses python3 and since context was not globally declared, deployment of SapMonitor failed during onboarding

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -420,6 +420,7 @@ def monitor(args):
       h.disconnect()
       
 def main():
+   global ctx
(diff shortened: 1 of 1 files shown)


tniek:
(Stopped Azure Pipelines build as it is relevant for Automated Deployment only)

tniek:
> Hotfix looks good, but why wasn't this caught in development? Is the logic only tested after merging?

That's right. Need to build an Azure Pipeline for SAP Monitor as well to catch these scenarios.

PR #181, opened 2019-07-09
Add HANA Landscape Host Configuration monitoring

tniek:
- Add M_LANDSCAPE_HOST_CONFIGURATION query
- Add logic to store last result hash (for point-in-time queries)
- Rename custom logs to match underlying HANA queries

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -12,13 +12,13 @@ import decimal
 from datetime import datetime, timedelta, date
 import hashlib, hmac, base64
 import logging, http.client as http_client
+import hashlib
 
 ###############################################################################
 
-PAYLOAD_VERSION              = "0.2"
+PAYLOAD_VERSION              = "0.3"
 STATE_FILE                   = "sapmon.state"
 INITIAL_LOADHISTORY_TIMESPAN = -(60 * 1)
-LOG_TYPE                     = "SapHana_Infra"
 TIME_FORMAT_HANA             = "%Y-%m-%d %H:%M:%S.%f"
 TIME_FORMAT_LOG_ANALYTICS    = "%a, %d %b %Y %H:%M:%S GMT"
 TIMEOUT_HANA                 = 5
@@ -72,6 +72,7 @@ class SapHana:
       colIndex = {col[0] : idx for idx, col in enumerate(self.cursor.description)}
       return colIndex, self.cursor.fetchall()
 
+   # TODO(tniek): Refactor monitoring into Query class
    def getLoadHistory(self, fromTimestamp):
       """
       Get infrastructure utilization via HANA Load History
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -288,38 +323,60 @@ class _Context:
...
-   def setLastPullTimestamp(self, timestamp):
+   def writeStateFile(self):
       """
-      Write the timestamp (UTC) of the last successful pull to state file
+      Persist current state (with hashes from point-in-time queries and last pull timestamp) into a local file
       """
       try:
          data = {
-            "lastPullUTC": timestamp.strftime(TIME_FORMAT_HANA)
+            "resultHashes": self.resultHashes,
```
persiaAziz: Can this key be renamed to something like previousResultHashes just for clarity?

monitor/sapmon.py
```diff
@@ -288,38 +323,60 @@ class _Context:
...
-      self.vmInstance = AzureInstanceMetadataService.getComputeInstance(operation)
-      vmTags          = dict(map(lambda s : s.split(':'), self.vmInstance["tags"].split(";")))
-      self.sapmonId   = vmTags["SapMonId"]
-      self.azKv       = AzureKeyVault("sapmon%s" % self.sapmonId)
-      self.lastPull   = self.readLastPullTimestamp()
+      self.vmInstance   = AzureInstanceMetadataService.getComputeInstance(operation)
+      vmTags            = dict(map(lambda s : s.split(':'), self.vmInstance["tags"].split(";")))
+      self.sapmonId     = vmTags["SapMonId"]
+      self.azKv         = AzureKeyVault("sapmon%s" % self.sapmonId)
+      self.readStateFile()
```
persiaAziz: It will be more readable if self.resultHashes and lastPull is initialized in the __init__. This will give a clarity on the member variables of the context class

monitor/sapmon.py
```diff
@@ -288,38 +323,60 @@ class _Context:
...
 
-   def readLastPullTimestamp(self):
+   def readStateFile(self):
       """
-      Read the timestamp (UTC) of the last successful pull from state file
+      Get most recent state (with hashes from point-in-time queries and last pull timestamp) from a local file
       """
-      lastPull = None
+      lastPull     = None
```
persiaAziz: Please move this to __init__

monitor/sapmon.py
tniek: +1
How about "lastResultHashes" to be analogous to "lastPullUTC"?

monitor/sapmon.py
tniek: -1
Note that lastPull and resultHashes here in lines 336/337 are local variables which get populated as the state file is read successfully. I assign their values to the class variables (self.lastPull, self.resultHashes) at the end of this function.

monitor/sapmon.py
tniek: I agree although that would introduce some redundancy, since readStateFile() would set lastPullUTC and resultHashes to empty (None and {}, respectively) if the state file doesn't exist yet or contains no values yet.

monitor/sapmon.py
persiaAziz: sorry I meant to point to the member variables of self . (i.e., self.lastPull, self.resultHashes)

monitor/sapmon.py
```diff
@@ -72,6 +72,7 @@ def runQuery(self, sql):
       colIndex = {col[0] : idx for idx, col in enumerate(self.cursor.description)}
       return colIndex, self.cursor.fetchall()
 
+   # TODO(tniek): Refactor monitoring into Query class
```
jovillas: It would be nicer to have each class on a separate file.

tniek: +1

persiaAziz: -1 for now. This code is going to be run by custom script. So having multiple files will break CS for now. Let's discuss before splitting these

PR #182, opened 2019-07-09
Exclude monitor and README from the pipeline trigger path

persiaAziz:
The monitor folder contains scripts for collecting load history from DB. Hence changes to scripts in monitor is not relevant to automated deployment. Same applies to README.md

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,7 +5,9 @@
 trigger: none
 # Only trigger build when a PR is opened
 pr:
-- master
+  branches:
+    include:
+    - master
   paths:
     exclude:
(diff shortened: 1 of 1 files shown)


persiaAziz:
> Change PR title to reflect what type of things are being excluded (because right now you only call out monitor/*)
> Add a PR description describing why changes in the excluded paths will never be relevant to test with the pipeline

Done

persiaAziz:
> I'm very fond of this change.
> But rather than excluding monitor/*, would it make more sense to explicitly include deploy/* (since that particular pipeline is for deployments only)?

Umm.. I don't see much difference in doing so.

PR #183, opened 2019-07-10
sapmon will now also accept keyvault url and msi client id instead of password

PakDLiu:
The sapmon.py script will now also accept KeyVault URL and MSI Client ID to get the password.

Testing manually on my own collector VM, no errors. Saw password get fetched correctly

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -33,37 +33,27 @@ class SapHana:
    connection = None
    cursor     = None
 
-   def __init__(self, host = None, port = None, user = None, password = None, passwordKeyVault = None, msiClientId = None, hanaDetails = None):
+   def __init__(self, host = None, port = None, user = None, password = None, hanaDetails = None):
       if hanaDetails:
-         self.host             = hanaDetails["HanaHostname"]
-         self.port             = hanaDetails["HanaDbSqlPort"]
-         self.user             = hanaDetails["HanaDbUsername"]
-         self.password         = hanaDetails["HanaDbPassword"]
-         self.passwordKeyVault = hanaDetails["HanaDbPasswordKeyVault"]
-         self.msiClientId      = hanaDetails["MsiClientId"]
+         self.host     = hanaDetails["HanaHostname"]
+         self.port     = hanaDetails["HanaDbSqlPort"]
+         self.user     = hanaDetails["HanaDbUsername"]
+         self.password = hanaDetails["HanaDbPassword"]
       else:
-         self.host             = host
-         self.port             = port
-         self.user             = user
-         self.password         = password
-         self.passwordKeyVault = passwordKeyVault
-         self.msiClientId      = msiClientId
+         self.host     = host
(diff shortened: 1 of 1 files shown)


pabowers:
Why is this PR affecting the Pipeline build check? It seems a little unrelated

tniek:
> Why is this PR affecting the Pipeline build check? It seems a little unrelated

Because PR #182 has not been merged yet.

tniek:
> Looks good to me, just look into why this is causing the pipeline to fail

Bad branch name (@PakDLiu - I made the same mistake, check the last sentence in our [Contributing Guidelines](https://github.com/Azure/sap-hana/blob/master/CONTRIBUTING.md) :)) causes the VNET creation of the (unrelated) auto-deployment to fail.

monitor/sapmon.py
```diff
@@ -32,27 +33,37 @@ class SapHana:
    connection = None
    cursor     = None
 
-   def __init__(self, host = None, port = None, user = None, password = None, hanaDetails = None):
+   def __init__(self, host = None, port = None, user = None, password = None, passwordKeyVault = None, msiClientId = None, hanaDetails = None):
```
tniek: I would actually prefer to rename `passwordKeyVault` into `keyVaultUrl`, as this describes more adequately what it actually is.

monitor/sapmon.py
```diff
@@ -203,13 +214,13 @@ def getComputeInstance(operation):
          )["compute"]
 
    @staticmethod
-   def getAuthToken(resource):
+   def getAuthToken(resource, clientId = None):
```
tniek: For consistency (line 38), I would recommend renaming `clientId` into `msiClientId`.

monitor/sapmon.py
```diff
@@ -220,9 +231,9 @@ class AzureKeyVault:
    """
    params  = {"api-version": "7.0"}
 
-   def __init__(self, keyvaultName):
+   def __init__(self, keyvaultName, clientId = None):
```
tniek: See above: For consistency (line 38), I would recommend renaming `clientId` into `msiClientId`.

persiaAziz:
Thanks for this PR Donald, this PR tests PR 182 ✔️

monitor/sapmon.py
tniek: See my review comment: It actually makes more sense to remove the entire KeyVault business from the `Hana` class and move it into `_Context`, presumably there into `parseSecrets()`

PakDLiu:
Note:
If the VM has more than one user assigned identities assigned to it, the client ID is required.

Update:
If we are fetching the password from the customer's keyvault (IE: multiple identities assigned to the VM), this script will check the tags for the sapmon MSI client id.

There is not changes in the current flow, since it will try to find the tag, won't find it, and fetch the MSI token with no client id (will work since there is only 1 MSI assigned to it), and everything will work as normal.

PR #186, opened 2019-07-15
Fix HA pair by disabling Cloud-Netconfig

rukawata:
- SLES12 includes [Cloud-Netconfig](https://github.com/SUSE-Enceladus/cloud-netconfig), which is a tool that automatically control network interface cards on Azure VMs. This apparently has impact on the stability of Pacemaker clusters as it interferes with the configuration of the floating IP address.
- This change will manually disable Cloud-Netconfig during automated HANA HA pair deployment by adding an Ansible role that sets `CLOUD_NETCONFIG_MANAGE=no` in `/etc/sysconfig/network/ifcfg-eth0`.

### DIFF
--- a/deploy/vm/ansible/roles/disable-cloud-netconfig/tasks/main.yml
+++ b/deploy/vm/ansible/roles/disable-cloud-netconfig/tasks/main.yml
@@ -1,6 +1,7 @@
-- name: set CLOUD_NETCONFIG to no
+- name: disable CLOUD_NETCONFIG_MANAGE
   shell: |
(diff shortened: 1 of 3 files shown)


deploy/vm/ansible/roles/disable-cloud-netconfig/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+- name: set CLOUD_NETCONFIG to no
```
tniek: nit: "disable CLOUD_NETCONFIG_MANAGE"

PR #175, opened 2019-07-02
New module structure based on Reference Architectures

suhanselvan-wft:
This branch has new Terraform modules created to implement the reference architectures for different scenarios. 

The proposed module workflow can be found here,
[SAP on Azure Terraform Workflow.docx](https://github.com/Azure/sap-hana/files/3348262/SAP.on.Azure.Terraform.Workflow.docx)

Currently, the modules to create the common infrastructure resources have been added.

PR #187, opened 2019-07-16
v2 : Deploy azure resources common for all the architectures for different scenarios

suhanselvan-wft:
This pull request is a part of the v2 implementation of SAP automation on Azure.
Currently, the Terraform root module (single_node_hana) for single HANA instance architecture and common infrastructure module (common_infrastructure) have been created.

->The common infrastructure module deploys the resource group, vnets, subnets, nsgs and performs vnet peering between the management and sap vnets.
-> vnet peering is done when either the management or sap vnet is not existing or when neither of them are already existing in the user's subscription.
->The json user input file (terraform.tfvars.json) has been updated according to the specifications discussed.


deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
```diff
@@ -0,0 +1,60 @@
...
+                    "is_existing" : false,
+                    "name" : "test-rg"
+            },
+
+            "vnets" : {
+                    "management" : {
+                            "is_existing" : false,
+                            "name" : "mgmt-vnet",
+                            "address_space" : "10.0.0.0/16",
+                            "default_subnet" : {
```
tniek: I see that you changed the structure here (and in `infrastructure.vnets.sap`).
Given the amount of subnets (one in `infrastructure.vnets.management` and four in `infrastructure.vnets.sap`), I think it would be better to have a `subnets` key inside the `vnets.*` dictionary.
`subnets` would be a dictionary of dictionaries, something along the lines of:
`"subnets" : {
   "default": {
      "is_existing": false,
      "name": "subnet-mgmt-default",
      "prefix": "10.0.1.0/24",
      "nsg": {
         "is_existing": false,
         "name": "nsg-mgmt-default"
      }
   }
}`

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
```diff
@@ -0,0 +1,60 @@
...
+                                    "is_existing" : false,
+                                    "name" : "mgmt-default-subnet",
+                                    "prefix" : "10.0.1.0/24",
```
tniek: I see that the azurerm TF provider distinguishes between `address_space` and `prefix`; would it make sense to keep the same in the JSON?

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
```diff
@@ -0,0 +1,60 @@
...
+
+            "resource_group" : {
+                    "is_existing" : false,
+                    "name" : "test-rg"
+            },
+
+            "vnets" : {
+                    "management" : {
+                            "is_existing" : false,
+                            "name" : "mgmt-vnet",
```
tniek: Would the VNET name even have to be declared? (Rather than us providing a default value in `terraform.tfvars`?)

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
```diff
@@ -0,0 +1,60 @@
...
+                            "admin_subnet" : {
+                                    "is_existing" : false,
+                                    "name" : "db-admin-subnet",
+                                    "prefix" : "10.1.1.0/24",
+                                    "nsg" : {
+                                            "is_existing" : false,
+                                            "name" : "db-admin-nsg"
+                                    }
+                            },
+                            "client_subnet" : {
```
tniek: Why the renaming from `db` to `client`?

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
```diff
@@ -0,0 +1,60 @@
...
+                                    "is_existing" : false,
+                                    "name" : "db-client-subnet",
```
tniek: Which one is it, client or DB?

deploy/v2/terraform/modules/single_node_hana/variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "infrastructure" {
+  description = "Details of the azure infrastucture required for the deployment"
```
tniek: nit: "Azure"

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
+##################################################################################################################
+# RESOURCES
+##################################################################################################################
+
+
```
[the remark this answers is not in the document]
suhanselvan-wft: Reduced to single blank lines.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  name     = var.resource_group_name
+  location = var.region
+}
+
+
+# VNETs ==========================================================================================================
+
+
+# Creates the management vnet
+resource "azurerm_virtual_network" "mgmt-vnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count               = var.mgmt_vnet_existing ? 0 : 1
+  name                = var.mgmt_vnet_name
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+  address_space       = [var.mgmt_vnet_address_space]
+}
+
+# Creates the sap vnet
```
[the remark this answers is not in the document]
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+
+
+# RESOURCE GROUP =================================================================================================
+
+
+# Creates the resource group
+resource "azurerm_resource_group" "rg" {
+  count    = var.resource_group_existing ? 0 : 1
+  name     = var.resource_group_name
+  location = var.region
```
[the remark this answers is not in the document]
suhanselvan-wft: > As discussed, I would propose to reference the input variables without assigning an "intermediate" variable first.
> I.e. `location = infrastructure.region` (you could pass in the infrastructure section from the JSON, since this module is limited to infra only).

The latest commit has this change implemented. Since we are now passing the whole infrastructure section as a single variable, the default values for resources in `common_infrastructure` module have been defined separately in its `variables.tf` file.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+resource "azurerm_virtual_network" "mgmt-vnet" {
+  count               = var.mgmt_vnet_existing ? 0 : 1
+  name                = var.mgmt_vnet_name
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+  address_space       = [var.mgmt_vnet_address_space]
+}
+
+# Creates the sap vnet
+resource "azurerm_virtual_network" "sap-vnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count               = var.sap_vnet_existing ? 0 : 1
+  name                = var.sap_vnet_name
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+  address_space       = [var.sap_vnet_address_space]
+}
+
+# Imports existing management vnet data
```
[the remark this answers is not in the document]
suhanselvan-wft: Fixed

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+data "azurerm_virtual_network" "mgmt-vnet" {
+  count               = var.mgmt_vnet_existing ? 1 : 0
+  name                = var.mgmt_vnet_name
+  resource_group_name = var.resource_group_name
+}
+
+# Imports existing sap vnet data
```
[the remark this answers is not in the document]
suhanselvan-wft: Fixed

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  name                = var.sap_vnet_name
+  resource_group_name = var.resource_group_name
+}
+
+
+# SUBNETs ========================================================================================================
+
+
+# Creates the management default subnet
+resource "azurerm_subnet" "mgmt-default-subnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count                = var.mgmt_default_subnet_existing ? 0 : 1
+  name                 = var.mgmt_default_subnet_name
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.mgmt_vnet_existing ? var.mgmt_vnet_name : azurerm_virtual_network.mgmt-vnet[0].name
+  address_prefix       = var.mgmt_default_subnet_prefix
+}
+
+# Creates the sap admin subnet
+resource "azurerm_subnet" "sap-admin-subnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: > To be consistent, rename into `subnet-sap-admin`

Have renamed them accordingly.

> Also, I see that this subnet was not part of our original design.
> What exactly is it used for? How is `subnet-sap-admin` different from `subnet-mgmt-default`?

This was discussed during the meeting (also referenced in the sync notes).The database nodes will be having two vNICs. One for admin traffic and one for client traffic which need to be isolated. The application servers will be in a separate subnet. So there will be three subnets in the SAP VNET, `subnet-sap-admin`, `subnet-sap-client` and `subnet-sap-app`. The `subnet-mgmt-default` is a part of the management VNET where the jumpbox will reside.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count                = var.sap_admin_subnet_existing ? 0 : 1
+  name                 = var.sap_admin_subnet_name
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.sap_vnet_existing ? var.sap_vnet_name : azurerm_virtual_network.sap-vnet[0].name
+  address_prefix       = var.sap_admin_subnet_prefix
+}
+
+# Creates the sap client subnet
+resource "azurerm_subnet" "sap-client-subnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: > Should be `subnet-sap-client`

Have renamed them accordingly.

> I didn't see this one in the original design either; instead, `subnet-sap-db` is missing. Or is this what `subnet-sap-client` is?

Please refer to this reply,
https://github.com/Azure/sap-hana/pull/187#discussion_r304821393

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count                = var.sap_client_subnet_existing ? 0 : 1
+  name                 = var.sap_client_subnet_name
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.sap_vnet_existing ? var.sap_vnet_name : azurerm_virtual_network.sap-vnet[0].name
+  address_prefix       = var.sap_client_subnet_prefix
+}
+
+# Creates the sap app subnet
+resource "azurerm_subnet" "sap-app-subnet" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count                = var.sap_app_subnet_existing ? 0 : 1
+  name                 = var.sap_app_subnet_name
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.sap_vnet_existing ? var.sap_vnet_name : azurerm_virtual_network.sap-vnet[0].name
+  address_prefix       = var.sap_app_subnet_prefix
+}
+
+
+# NSGs ===========================================================================================================
+
```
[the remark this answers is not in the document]
suhanselvan-wft: > Does it really make sense to create a separate NSG for each subnet? Wouldn't they share some common rules?

They may share some common rules but, there will also be rules specific to each of the subnet. This was also agreed upon during the meeting. (listed in meeting notes).


> Also, I don't see the logic to specify (and import) existing NSGs?

Since the user input file terraform.tfvars.json has a `is_existing` flag for each resource, only those resources with the flag set to `false` are created. If they are set to `true`, the arguments of the resource, say `"name"` of  `"resource_group"` , will be used just to create other resources such as VNET which require that property.

Importing existing resources would import them into Terraform's state. So if the user runs `terraform destroy` Terraform would try to destroy the existing resources along with the resources created by it. This will be a problem if the user has other resources(e.g. VNET) in their subscription that depend on this existing resource(e.g. resource group).

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+
+# Creates management default subnet nsg
+resource "azurerm_network_security_group" "mgmt-default-nsg" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count               = var.mgmt_default_nsg_existing ? 0 : 1
+  name                = var.mgmt_default_nsg_name
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+}
+
+# Creates sap admin subnet nsg
+resource "azurerm_network_security_group" "sap-admin-nsg" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count               = var.sap_admin_nsg_existing ? 0 : 1
+  name                = "${var.sap_admin_nsg_name}"
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+}
+
+# Creates sap client subnet nsg
+resource "azurerm_network_security_group" "sap-client-nsg" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count               = var.sap_client_nsg_existing ? 0 : 1
+  name                = var.sap_client_nsg_name
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+}
+
+# Creates sap app subnet nsg
+resource "azurerm_network_security_group" "sap-app-nsg" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  name                = "${var.sap_app_nsg_name}"
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+}
+
+
+# VNET PEERINGs ==================================================================================================
+
+
+# Peers management vnet to sap vnet
```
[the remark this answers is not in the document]
suhanselvan-wft: Fixed

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+resource "azurerm_virtual_network_peering" "mgmt-sap-peering" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  name                         = "${var.mgmt_vnet_name}-${var.sap_vnet_name}"
+  resource_group_name          = var.resource_group_name
+  virtual_network_name         = var.mgmt_vnet_existing ? var.mgmt_vnet_name : azurerm_virtual_network.mgmt-vnet[0].name
+  remote_virtual_network_id    = var.sap_vnet_existing ? data.azurerm_virtual_network.sap-vnet[0].id : azurerm_virtual_network.sap-vnet[0].id
+  allow_virtual_network_access = true
+}
+
+
+# Peers sap vnet to management vnet
+resource "azurerm_virtual_network_peering" "sap-mgmt-peering" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have renamed them accordingly.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+  count                        = signum((var.mgmt_vnet_existing ? 0 : 1) + (var.sap_vnet_existing ? 0 : 1))
+  name                         = "${var.mgmt_vnet_name}-${var.sap_vnet_name}"
+  resource_group_name          = var.resource_group_name
+  virtual_network_name         = var.mgmt_vnet_existing ? var.mgmt_vnet_name : azurerm_virtual_network.mgmt-vnet[0].name
+  remote_virtual_network_id    = var.sap_vnet_existing ? data.azurerm_virtual_network.sap-vnet[0].id : azurerm_virtual_network.sap-vnet[0].id
+  allow_virtual_network_access = true
+}
+
+
+# Peers sap vnet to management vnet
```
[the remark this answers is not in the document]
suhanselvan-wft: Fixed

deploy/v2/terraform/modules/single_node_hana/main.tf
```diff
@@ -0,0 +1,38 @@
+# Initalizes azure modules and authenticates to subscription
+provider "azurerm" {
+  version = "~> 1.30.1"
+}
+
+# Setup common infrastructure
+module "common_infrastructure" {
```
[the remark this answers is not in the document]
suhanselvan-wft: Please refer the following response,
https://github.com/Azure/sap-hana/pull/187#discussion_r304821034

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: The initial input JSON file had VNETs as a list of dictionaries, which would make referring to each VNET using an index number. So the only change I did was to make it into a dictionary of dictionaries. As for the number of subnets, there will be a total of four(one in management VNET and three in SAP VNET). This is one more than what was initially specified in the input JSON file. 

Since, subnets are not independent resources, rather a sub resource of a VNET, having a `subnets` key in the `vnets.* ` dictionary will break the logical hierarchy while referencing them. Also, creating a subnet requires the VNET name as an argument. So creating a separate `subnets` dictionary would require the VNET name argument to be specified redundantly.

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: Please refer to the following response,
https://github.com/Azure/sap-hana/pull/187#discussion_r304821393

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: Please refer to the following response,
https://github.com/Azure/sap-hana/pull/187#discussion_r304821393

Have removed the ambiguous `db` text.

deploy/v2/terraform/modules/single_node_hana/variables.tf
suhanselvan-wft: Fixed

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: If the users want to use the default values they can do so by leaving the value for the `"name"` key blank. e.g.` "name" : ""`. I think with this kind of implementation, the user will not have to modify the structure of the JSON file. Thereby reducing any chances breaking the JSON file due to human error.

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: I'm sorry, I don't get it. Can you please elaborate?

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
pabowers: While I understand the complexities of `terraform destroy` removing the imported resources, this is particularly confusing to a customer. When we give them the option to use an existing resource, updates to that resource should also be reflected in the terraform state.  It is also possible to prevent the destruction of these imported resources in Terraform. See this setting that you can add to resources that will prevent them from being destroyed from a call to `terraform destroy` https://www.terraform.io/docs/configuration/resources.html#prevent_destroy

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,141 @@
+##################################################################################################################
+# RESOURCES
+##################################################################################################################
```
persiaAziz: Please include a README.md here with a diagram of the resources deployed and how those are related. THe diagram does not have to be comprehensive at this point, but it will be easy to visualize the relationship between the subnets and resources etc.  An example README could be https://github.com/Azure/sap-hana/tree/master/deploy/vm/modules/single_node_hana

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
suhanselvan-wft: @tniek If this is the requirement, it would negate the requirement of using the `is_existing` argument in the input JSON file. If there are existing resources with the arguments specified by the user, a Terraform resource block automatically imports it into its state . Please let me know if this is okay.

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
suhanselvan-wft: All the resources created are based on the architecture provided. Maybe @tniek or @persiaAziz could provide you that information.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+# Creates sap app subnet nsg
+resource "azurerm_network_security_group" "sap-app-nsg" {
+  count               = var.sap_app_nsg_existing ? 0 : 1
+  name                = "${var.sap_app_nsg_name}"
+  location            = var.region
+  resource_group_name = var.resource_group_existing ? var.resource_group_name : azurerm_resource_group.rg[0].name
+}
+
+
+# VNET PEERINGs ==================================================================================================
```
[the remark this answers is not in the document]
suhanselvan-wft: Thsi particular formatting is used to group resource blocks as requested by @tniek here,
https://github.com/Azure/sap-hana/pull/175#discussion_r299650268

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
tniek: Ok, thanks for clarifying. So if only a HANA instance is deployed (no application server), which subnet would it reside in? In that case, would `subnet-sap-app` really be needed?

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
tniek: I think it bloats the input JSON unnecessarily to have all variables for which the user wants to use the default value set to empty.
In this specific example, `name` should only have to be declared if the user wants to override its default value.

deploy/v2/terraform/modules/common_infrastructure/main.tf
tniek: I think that's a good idea.
@suhanselvan-wft Please use the diagrams of the reference architecture as a basis for the README.md

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,152 @@
...
+# Imports existing sap vnet data
+data "azurerm_virtual_network" "sap-vnet" {
+  count               = var.sap_vnet_existing ? 1 : 0
+  name                = var.sap_vnet_name
+  resource_group_name = var.resource_group_name
+}
+
+
+# SUBNETs ========================================================================================================
+
```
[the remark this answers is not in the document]
tniek: I like @pabowers 's idea of still importing existing resources, but flagging them so they wouldn't get deleted upon running `terraform destroy`

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
tniek: See above - supporting @pabowers 's proposal to set `prevent_destroy` on imported existing resources

deploy/v2/terraform/modules/single_node_hana/main.tf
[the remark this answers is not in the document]
tniek: I see how you did that, and it's not particularly nice to have the default JSON inside the variables.tf.
Would it make sense to instead have a `default.json` with all variables (that could potentially be overwritten by the user-supplied `input.json`)? In this case, you would have to refer to infrastructure-specific variables as `var.infrastructure.*`

deploy/v2/terraform/modules/common_infrastructure/variables.tf
```diff
@@ -0,0 +1,53 @@
+variable "infrastructure" {
```
rukawata: What is the difference between variable infrastructure and infra_default? 
Currently there is nothing in variable infrastructure except description.

deploy/v2/terraform/modules/common_infrastructure/variables.tf
```diff
@@ -0,0 +1,53 @@
...
+        "subnet_client" = {
+          "name"   = "subnet-sap-client"
+          "prefix" = "10.1.2.0/24"
+          "nsg" = {
+            "name" = "nsg-sap-client"
+          }
+
+        }
+        "subnet_app" = {
+          "name"   = "subnet-app"
```
rukawata: Please keep the name conversion consistent.
```suggestion
          "name"   = "subnet-sap-app"
```

deploy/v2/terraform/modules/common_infrastructure/variables.tf
```diff
@@ -0,0 +1,53 @@
...
+          "prefix" = "10.1.3.0/24"
+          "nsg" = {
+            "name" = "nsg-app"
```
rukawata: Same as above:
```suggestion
            "name" = "nsg-sap-app"
```

deploy/v2/terraform/modules/single_node_hana/README.md
```diff
@@ -0,0 +1,5 @@
+# HANA Single-Node Instance
+
+This scenario deploys a single-node HANA instance and a Windows and/or Linux Jumpbox on Azure Cloud.
+
+<img src="https://user-images.githubusercontent.com/5745963/61615719-c0892100-ac34-11e9-9fe0-2ea4e17d8598.png" alt="Landscape Diagram" width="450"/>
```
rukawata: The Subnet and NSG for HANA Database Node are missing in the graph.

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
suhanselvan-wft: As I said, currently a resource block automatically imports resources if they do not exist already. But this is actually an issue with the current Terraform azurerm provider and will be removed in an upcoming release.

https://www.terraform.io/docs/providers/azurerm/guides/2.0-upgrade-guide.html#changes-when-importing-existing-resources

Then users will be required to manually run `Terraform import` command to import existing resources into state before running our scripts. Also, we do not update existing resources rather only add resources/sub-resources which is why I'm just importing the details of the existing resources as `data sources`.

deploy/v2/terraform/modules/single_node_hana/main.tf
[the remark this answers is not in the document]
suhanselvan-wft: I tried implementing this. But the `input.json`  overrides the keys along with their values defined in `default.json`.  So if the user wants to use the default value for `name` key from `default.json` and skips the `name` key declaration in `input.json`. The key itself is removed as the JSON structure defined in `input.json` takes precedence over `default.json`.

deploy/v2/terraform/modules/common_infrastructure/main.tf
suhanselvan-wft: This has been added.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,141 @@
...
+  count                = var.infrastructure.vnets.sap.subnet_app.is_existing ? 0 : 1
+  name                 = var.infrastructure.vnets.sap.subnet_app.name == "" ? var.infra_default.vnets.sap.subnet_app.name : var.infrastructure.vnets.sap.subnet_app.name
+  resource_group_name  = var.infrastructure.resource_group.name == "" ? var.infra_default.resource_group : var.infrastructure.resource_group.name
+  virtual_network_name = !var.infrastructure.vnets.sap.is_existing ? azurerm_virtual_network.vnet-sap[0].name : var.infrastructure.vnets.sap.name == "" ? var.infra_default.vnets.sap.name : var.infrastructure.vnets.sap.name
+  address_prefix       = var.infrastructure.vnets.sap.subnet_app.prefix == "" ? var.infra_default.vnets.sap.subnet_app.prefix : var.infrastructure.vnets.sap.subnet_app.prefix
+}
+
+# NSGs ===========================================================================================================
+
+# Creates management default subnet nsg
```
persiaAziz: I see that this PR is creating or taking existing nsg as input but not really associating the nsg with the subnet. Is that going to happen in the next PR? I think since we have both nsg nd subnet in this PR, the association part should also be included here.

deploy/v2/terraform/modules/common_infrastructure/main.tf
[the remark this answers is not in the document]
suhanselvan-wft: Agree, `subnet-sap-app` is not required for single node HANA scenario. Updated code to skip deployment of this resource for this scenario. The HANA instance will have two vNICs, one residing in `subnet-sap-admin` and the other in `subnet-sap-client`.

deploy/v2/terraform/modules/common_infrastructure/main.tf
suhanselvan-wft: The NSG association was meant for this PR. It has been added now. Thanks for the catch.

deploy/v2/terraform/modules/single_node_hana/terraform.tfvars.json
suhanselvan-wft: As discussed over online meeting, The code has been updated such that if the user wants an existing resource to be used, the `is_existing` flag should be set to `true` and only the `name` argument needs to be declared. The other arguments have to be declared in the input JSON file only if the resource is not existing.

deploy/v2/terraform/modules/single_node_hana/input.auto.tfvars.json
```diff
@@ -0,0 +1,51 @@
...
+                            "subnet_admin" : {
+                                    "is_existing" : "false",
+                                    "name" : "subnet-admin",
+                                    "prefix" : "10.1.1.0/24",
+                                    "nsg" : {
+                                            "is_existing" : "false",
+                                            "name" : "nsg-admin"
+                                    }
+                            },
+                            "subnet_client" : {
```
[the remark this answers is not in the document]
suhanselvan-wft: Have made the replacement and updated on the diagram in README.md file as well.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -164,10 +164,10 @@ data "azurerm_network_security_group" "nsg-admin" {
   resource_group_name = var.infrastructure.resource_group.name
 }
 
-# Imports the SAP client subnet nsg data
-data "azurerm_network_security_group" "nsg-client" {
-  count               = var.infrastructure.vnets.sap.subnet_client.nsg.is_existing ? 1 : 0
-  name                = var.infrastructure.vnets.sap.subnet_client.nsg.name
+# Imports the SAP db subnet nsg data
```
rukawata: This is a question related to how to use the imported existing resources. 

**Data Source** provides read only access on one existing resource (in this case, Network Security Group) from the same or a different configuration file (*.tf). It does not work with an existing resource that get into tfstate by `terraform import`.

On the other hand, get an existing resource into the landscape using `terraform import` will require:
- A resource block for it in the configuration, establishing the name by which it will be known to Terraform.
- After the import, adjustments to the configuration to align with the current (or desired) state of the imported object.

I do not see how above procedures are handled here. Please let me know if I missed anything.

suhanselvan-wft: I thought it was decided during the call that if the user wants to use an existing resource they set the `is_existing`  flag to `true` and provide just the `name` of the resource in the input JSON file. `terraform import` will not be used until we find a way to call it within the modules and not require the user to manually import before execution of the scripts.

rukawata: Ok. Then for now we will not manage the existing resource. 

With this approach, there are still problems:
For example with `resource_group_name = var.infrastructure.resource_group.name`, the `resource_group` should be the one for the existing resource. And it may be different for different existing resources.

In order to pass that information easier, I would propose:

1. To pass ARM id in the JSON if `is_existing` is true.
2. Parse the ARM id to get the resource name and resource group name.

suhanselvan-wft: Yes, I thought about that. But during the initial discussions it was expected that the user has all the resources in a single resource group. I can implement this if @tniek is fine with using ARM ID as input. He was earlier concerned about user experience of having to find the ARM ID. Else we can ask the user to pass the resource group name.

tniek: +1 on having the user specify an ARM ID (rather than name) of existing resources.

While it is a reasonable assumption that all (new) resources that get deployed as part of a landscape will be in one (single) resource group, I agree that there could be existing resources (VNETs, subnets, NSGs) that are already part of other resource groups.

From a user experience perspective, it's probably easier to copy & paste the ARM ID into one single variable inside the JSON, as opposed to the name/resource_group combination into two separate fields.

@suhanselvan-wft - Please go ahead an implement this suggestion.

suhanselvan-wft: Sure. Will do

suhanselvan-wft: @tniek I just found that Azure portal does not show the ARM ID for subnets. The user will have to use powershell or Azure CLI to query and retrieve it. Is this okay?

suhanselvan-wft: This change has been implemented. The user can use existing resources by providing only the ARM ID using the `"arm_id"` key in the JSON. I just have the one question asked above.

tniek: @suhanselvan-wft - Yes, it's fine for the user (as a stop-gap solution, anyway) to obtain the ARM ID of an existing subnet via AZ CLI or PS.

deploy/v2/terraform/modules/single_node_hana/input.auto.tfvars.json
```diff
@@ -0,0 +1,60 @@
+{
+
+    "infrastructure" : {
+            "region" : "",
```
rukawata: Two questions:

1. Why all default values are empty now
2. `arm_id` should not always be there:

    - it would be the only required key when `is_existing` is set to true.
    - when `is_existing` is set to false, `arm_id` is not required, it does not add any value to the JSON file.

suhanselvan-wft: 1. Those were not default values. Just sample data I had entered to test. 
2. The current JSON is just a representation of the possible inputs. The user can modify it according to their requirements. We will be providing documentation on how to modify the JSON after it reaches its final structucture with all the resources. The code works based on the conditions you specified.

tniek: So you're saying that:
- if `use_existing = false`, then `name` and all other properties of the pertinent resource are required; and,
- if `use_existing = true`, the only required variable in that section is `arm_id`?
None of this would have an impact on any nested properties, e.g. a subnet inside a VNET, though.

suhanselvan-wft: Yes. That is correct. 
if `"use_existing" = "true"` for say management VNET, the only required value is `arm_id` for that resource. The user can use an existing subnet as `subnet_mgmt` by setting `"use_existing" = "true"` and providing its `arm_id` or create a new `subnet_mgmt` inside that existing VNET by setting `"use_existing" = "false "` and providing `"name"` and `"prefix"`.

tniek: Great, thanks for clarifying!

tniek: The following definition of the admin subnet (omitting `arm_id` since `is_existing = true`):

```
"subnet_admin" : {
        "is_existing" : "true",
        "name" : "subnet-admin",
        "prefix" : "10.0.1.0/25",
        "nsg" : {
                "is_existing" : "false",
                "name" : "nsg-sap"
        }
},
```

results in the following error:

```
Error: Unsupported attribute

  on ../common_infrastructure/main.tf line 103, in data "azurerm_subnet" "subnet-sap-admin":
 103:   name                 = split("/",var.infrastructure.vnets.sap.subnet_admin.arm_id)[10]
    |----------------
    | var.infrastructure.vnets.sap.subnet_admin is object with 4 attributes

This object does not have an attribute named "arm_id".
```

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,236 @@
...
+  network_security_group_id = var.infrastructure.vnets.sap.subnet_app.nsg.is_existing ? data.azurerm_network_security_group.nsg-app[0].id : azurerm_network_security_group.nsg-app[0].id
+}
+
+# VNET PEERINGs ==================================================================================================
+
+# Peers management VNET to SAP VNET
+resource "azurerm_virtual_network_peering" "peering-management-sap" {
+  count                        = signum((var.infrastructure.vnets.management.is_existing ? 0 : 1) + (var.infrastructure.vnets.sap.is_existing ? 0 : 1))
+  name                         = "${var.infrastructure.vnets.management.is_existing ? data.azurerm_virtual_network.vnet-management[0].name : azurerm_virtual_network.vnet-management[0].name}-${var.infrastructure.vnets.sap.is_existing ? data.azurerm_virtual_network.vnet-sap[0].name : azurerm_virtual_network.vnet-sap[0].name}"
+  resource_group_name          = var.infrastructure.resource_group.is_existing ? data.azurerm_resource_group.rg[0].name : azurerm_resource_group.rg[0].name
```
rukawata: As discussed, in the scenario of using existing vnet, the `resource_group_name` should be the resource group name of the existing vnet.

Thanks.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,236 @@
...
+  resource_group_name = split("/",var.infrastructure.vnets.sap.arm_id)[4]
+}
+
+# SUBNETs ========================================================================================================
+
+# Creates mgmt subnet of management VNET
+resource "azurerm_subnet" "subnet-mgmt" {
+  count                = var.infrastructure.vnets.management.subnet_mgmt.is_existing ? 0 : 1
+  name                 = var.infrastructure.vnets.management.subnet_mgmt.name
+  resource_group_name = var.infrastructure.resource_group.is_existing ? data.azurerm_resource_group.rg[0].name : azurerm_resource_group.rg[0].name
```
rukawata: As discussed, in the scenario where we use existing vnet, the subnet needs to be created under the same resource group as vnet.

suhanselvan-wft: This change has been implemented. Also I have made the changes to the directory structure as requested. The input JSON file is now in `sap-hana/deploy/v2/` directory. The root module for the single node hana scenario is now under `sap-hana/deploy/v2/terraform/scenarios/single_node_hana/`. While running `terraform plan` or `apply` the user will have to explicitly specify the input variable file. e.g `terraform plan -var-file="../../../input.tfvars.json"`

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,236 @@
...
+  subnet_id                 = var.infrastructure.vnets.sap.subnet_app.is_existing ? data.azurerm_subnet.subnet-sap-app[0].id : azurerm_subnet.subnet-sap-app[0].id
+  network_security_group_id = var.infrastructure.vnets.sap.subnet_app.nsg.is_existing ? data.azurerm_network_security_group.nsg-app[0].id : azurerm_network_security_group.nsg-app[0].id
+}
+
+# VNET PEERINGs ==================================================================================================
+
+# Peers management VNET to SAP VNET
+resource "azurerm_virtual_network_peering" "peering-management-sap" {
+  count                        = signum((var.infrastructure.vnets.management.is_existing ? 0 : 1) + (var.infrastructure.vnets.sap.is_existing ? 0 : 1))
+  name                         = "${var.infrastructure.vnets.management.is_existing ? data.azurerm_virtual_network.vnet-management[0].name : azurerm_virtual_network.vnet-management[0].name}-${var.infrastructure.vnets.sap.is_existing ? data.azurerm_virtual_network.vnet-sap[0].name : azurerm_virtual_network.vnet-sap[0].name}"
```
rukawata: @suhanselvan-wft The name for `azurerm_virtual_network_peering` should contain resource group names of both vnets - in this way the name would be definitely unique.

With the current naming conversion, it will be a problem in below scenario:

1. rg1.vnet1 is peering with rg1.vnet2
2. Reuse rg1.vnet1
3. Creating new vnet rg2.vnet2
4. With current naming conversion, peering vnet1vnet2 already exist - creating fails.
5. However if we introduce resource group names, eg. rg1_vnet1_rg2_vnet2 - creating succesful :)

Thanks!

suhanselvan-wft: This has been fixed. 
Just to understand, so in the scenario you specified there is already another vnet named vnet2 in a different resource group to which rg1.vnet1 has an existing peering?

deploy/v2/terraform/scenarios/single_node_hana/README.md
```diff
@@ -0,0 +1,5 @@
+# HANA Single-Node Instance
```
tniek: I'm aware we discussed this during our last sprint, but all things considered, why do we even need to distinguish between scenarios in the first place?

Would a file structure like the following be feasible:
```
v2/
  input.json
+ terraform/
    main.tf
  + modules/
    + common_infra/
    + hana_instance/
    + hana_ha/
+ ansible/
  playbook.yml
  + roles/
    + db_hana_install/
    + os_disk_layout/
    + os_ha_setup/
```

That way, there's only one single terraform script that gets called, and it makes the distinction what infrastructure to deploy based on the variables from the input JSON. 

Happy to get some more opinions on this, we can also discuss during our next sprint.

deploy/v2/terraform/modules/common_infrastructure/main.tf
rukawata: Another possible scenario which would fail with the current version:
Customer wants to use existing vnets for both management vnet and sap vnet.
The problem would be we don't know if they have peering created between the two existing vnets. 
We could:
-  Option 1: add a section for peering in the input JSON
-  Option 2: require customer to create peering themselves before they run the script. And our script would assume if is_exising holds true for both management vnet and sap vnet, peering exists.

We can discuss how to handle this in the sprint.

rukawata: Let's go with option 2.
And we should document this to make sure if end user wants to reuse both management vent and sap vent, they should make sure the peering is in place before running the script.

deploy/v2/terraform/scenarios/single_node_hana/README.md
suhanselvan-wft: Updated with the new file structure.

deploy/v2/input.tfvars.json
```diff
@@ -1,60 +1,60 @@
...
-                                    "nsg" : {
-                                            "is_existing" : "",
-                                            "arm_id" : "",
-                                            "name" : ""
-                                    }
-                            }
-                    }
-            }
-    }
+        "infrastructure" : {
```
tniek: Why are all of these indented one more tab?

suhanselvan-wft: Fixed. Looks like my editor decided to interpret a tab as 8 spaces instead of 4 earlier. Reverted back to original.

persiaAziz:
Please update you branch before merging

persiaAziz:
just retriggered the build. Let's see what happens now.

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -0,0 +1,236 @@
...
+  name                 = var.infrastructure.vnets.sap.subnet_app.name
+  resource_group_name  = var.infrastructure.vnets.sap.is_existing ? data.azurerm_virtual_network.vnet-sap[0].resource_group_name : azurerm_virtual_network.vnet-sap[0].resource_group_name
+  virtual_network_name = var.infrastructure.vnets.sap.is_existing ? data.azurerm_virtual_network.vnet-sap[0].name : azurerm_virtual_network.vnet-sap[0].name
+  address_prefix       = var.infrastructure.vnets.sap.subnet_app.prefix
+}
+
+# Imports data of existing mgmt subnet
+data "azurerm_subnet" "subnet-mgmt" {
+  count                = var.infrastructure.vnets.management.subnet_mgmt.is_existing ? 1 : 0
+  name                 = split("/", var.infrastructure.vnets.management.subnet_mgmt.arm_id)[10]
```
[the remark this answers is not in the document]
suhanselvan-wft: While I understand your concern and that was my initial approach too, It was changed based on @tniek 's suggestion [https://github.com/Azure/sap-hana/pull/187#discussion_r304153826](url) to not declare a secondary set of variables.

PR #191, opened 2019-07-25
[HOTFIX] adding timeout and retries for all content downloads

rukawata:
- The download of the SAP binaries from an Azure blob can be quite spotty some time, which would lead to failed deployment

- To reduce the amount of failures caused by the above reason, we introduce below 2 precautions to all `get_url` and `win_get_url` tasks:
  - adding parameter `timeout`, and set to 30s (can be changed by `url_timeout`).
  - if the download is not successful, the above tasks will retry 10 times (can be changed by variable `url_retries_cnt`) with 10s in between (can be changed by `url_retries_delay`).

Below is an example output when the download fails:
[sample_output.log](https://github.com/Azure/sap-hana/files/3429943/sample_output.log)

### DIFF
--- a/deploy/vm/ansible/roles/saphana-install/tasks/main.yml
+++ b/deploy/vm/ansible/roles/saphana-install/tasks/main.yml
@@ -20,7 +20,6 @@
     dest: /hana/shared/install/SAPCAR_LINUX.EXE
     mode: 0755
     timeout: "{{ url_timeout }}"
-  # 10 retries with 10 second in between
   register: result
   until: result is succeeded
   retries: "{{ url_retries_cnt }}"
--- a/deploy/vm/ansible/roles/saphana-install/tasks/main.yml
+++ b/deploy/vm/ansible/roles/saphana-install/tasks/main.yml
@@ -19,11 +19,22 @@
     url: "{{ url_sapcar }}"
     dest: /hana/shared/install/SAPCAR_LINUX.EXE
     mode: 0755
+    timeout: "{{ url_timeout }}"
+  # 10 retries with 10 second in between
+  register: result
+  until: result is succeeded
+  retries: "{{ url_retries_cnt }}"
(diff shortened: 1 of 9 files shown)


Caiyeon:
10 retries seems a little excessive. I have a few questions, none of which are blockers:

1. What's the failure rate of one call?
1. What's the time required for one download? (or rather, the download of the largest file)

tniek:
> 10 retries seems a little excessive. I have a few questions, none of which are blockers:
> 1. What's the failure rate of one call?

Discussed offline.

> 2. What's the time required for one download? (or rather, the download of the largest file)

Required download time obviously depends on file size and location. For a 3GB file (HANA DB server archive) ~3-5 minutes within the same region or ~15-30 minutes cross-region.

deploy/vm/ansible/roles/saphana-install/tasks/main.yml
```diff
@@ -19,11 +19,22 @@
     url: "{{ url_sapcar }}"
     dest: /hana/shared/install/SAPCAR_LINUX.EXE
     mode: 0755
+    timeout: "{{ url_timeout }}"
+  # 10 retries with 10 second in between
```
tniek: Why is the comment only in this file?

rukawata: Sorry. I forgot to remove that line. Since now the retries can be configured, what's in the comment is not accurate. I will remove this comment.

rukawata:
> Overall ok, although not a fan of making the same changes in 9 files. Going forward, there should be a more generic download role if we don't decide to change the design altogether.
Totally agree. Just this will require re-factoring of the code. I will make sure in the new design this is changed.
> 
> Finally, as this issue reportedly causes deployments to fail, the fix should be marked as [Hotfix].

PR #192, opened 2019-07-26
[HOTFIX]Inbound NSG rule adding host agent's IP

persiaAziz:


abhinababasu:
Why is this a hotfix?

persiaAziz:
> Why is this a hotfix?

This is blocking PRs from getting merged, as the build is failing. Still doing some trial and error to make the NSG work. Terraform has a bug where passing a list of strings(IP addresses in this case) as environment variables don't get translated properly

azure-pipelines.yml
```diff
@@ -92,5 +94,5 @@ jobs:
   steps:
   - script: |
       az login --service-principal --user $(hana-pipeline-spn-id) --password  $(hana-pipeline-spn-pw) --tenant $(landscape-tenant) --output none
-      az group delete -n single-node-$(Build.BuildId) --yes --subscription $(landscape-subscription)
+      #az group delete -n single-node-$(Build.BuildId) --yes --subscription $(landscape-subscription)
```
PakDLiu: why is this commented out?

persiaAziz: This PR is not fixing the unreachability problem yet. So looking in to the VM to investigate what is going wrong. Unfortunately , I couldn't find a way to mark this PR as draft

azure-pipelines.yml
```diff
@@ -32,8 +32,10 @@ jobs:
     displayName: 'Install python tools'
   - script: |
       ssh-keygen -b 4096 -t rsa -f /tmp/sshkey -q -N ""
+      export BUILD_MACHINE_IP=$(dig +short myip.opendns.com @resolver1.opendns.com)
```
tniek: Will this be a problem with IPv6 addresses?
In case it is, another option would be:
`curl ifconfig.co -v4`

persiaAziz: By default this command returns the A record (IPV4 address). So should be good

PR #194, opened 2019-07-31
[HOTFIX] Adapt KV name according to new naming convention

tniek:
- Rename KV from "sapmon1234567890" to "sapmon-kv-1234567890" (to be consistent with new naming convention for managed resources)

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -18,7 +18,7 @@ import re
 
 ###############################################################################
 
-PAYLOAD_VERSION              = "0.4"
+PAYLOAD_VERSION              = "0.4.1"
 PAYLOAD_DIRECTORY            = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -344,7 +345,8 @@ def setSecret(self, secretName, secretValue):
             data   = json.dumps({"value": secretValue})
             ) == secretValue
       except Exception as e:
-         logger.error("could not set KeyVault secret (%s)" % e)
+         logger.critical("could not set KeyVault secret (%s)" % e)
```
Caiyeon: Does `e` currently not include the http status code? Would it be easy enough to do in this PR? Seems like a very easy thing to add

lagalbra: Shouldn't be added in a hotfix though

Caiyeon: Agreed, hence not a blocker. Just curious

PR #195, opened 2019-08-02
Script to do basic system validation for known issues

jeffaco:
Create a script to perform basic system validation for LI/VLI
systems.

The script checks: Boot file configuration, NIC configuration
(for VLI systems only), EDAC modules, and system swap space.

### DIFF
--- a/tools/checksystem.sh
+++ b/tools/checksystem.sh
@@ -1,0 +1,178 @@
+#!/bin/bash
+
+# Set global exit code
+EXIT_STATUS=0
+
+# Boot configuration files
+BOOT_INSTALLDEVICE_FILENAME="/etc/default/grub_installdevice"
+BOOT_DEVICEMAP_FILENAME="/boot/grub2/device.map"
+
+checkBootConfig()
+{
+  # Note: Handling of these files are tricky, as it's not clear when the files exist.
+  #
+  # 1. We know that if the files don't exist in SP3, we can upgrade to SP4
+  #    without issue. But we can't depend on this without further testing,
+  # 2. The files do exist in Operations-generated images,
+  # 3. The files do exist in SuSE-generated SP3 images,
+  # 4. The files do NOT exist in SuSE-generated SP4 images.
+  #
+  # We're reaching out to SuSE support to more fully understand when the
+  # files are (or are not) created, and if deleting them is advised for
+  # mitigation.
+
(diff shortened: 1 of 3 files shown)


tools/checksystem.sh
```diff
@@ -0,0 +1,176 @@
...
+  fi
+
+  return
+}
+
+checkNicConfig()
+{
+  # Verify that flags are properly set to avoid port flapping on VLI systems
+
+  # If this isn't a HPE VLI system, then there's no NIC flags to be concerned of
```
pabowers: nit: "there are" instead of "there's"

tools/checksystem.sh
```diff
@@ -0,0 +1,176 @@
...
+    fi
+  done
+
+  return
+}
+
+checkEdacDisabled()
+{
+  # EDAC modules are comprised of Kernel modules "sb_edac" and "edac_core".
+  # Insure that neither of these modules are currently loaded as modules.
```
pabowers: nit: Ensure

tools/checksystem.sh
```diff
@@ -0,0 +1,176 @@
...
+
+  if lsmod | grep -q sb_edac; then
+    echo "EDAC enabled! Module sb_edac is loaded as kernel module" 1>&2
+    EXIT_STATUS=1
+  fi
+
+  if lsmod | grep -q edac_core; then
+    echo "EDAC enabled! Module edac_core is loaded as kernel module" 1>&2
```
pabowers: since this is an error, should it also be prefaced with "ERROR"?

tools/checksystem.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
```
persiaAziz: SInce this is a new folder, can you please add a README here with some description of this checksystem tool. Also I would recommend putthing the folders like "tools/systemvalidation/checksystem.sh" for a clear seperation in case if you have more tools like this in the future

tools/checksystem.sh
jeffaco: Yup, good catch. Doesn't affect operation, but not as "obvious". Fixed here and at line 93.

tools/checksystem.sh
jeffaco: Added, thanks for the suggestion.

PR #196, opened 2019-08-03
Set up CI with Azure Pipelines for wft

rukawata:


azure-pipelines-1.yml
```diff
@@ -0,0 +1,19 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- wft
```
tniek: What exactly triggers this pipeline? (Tag, certain folder, ...?)

Caiyeon:
Code LGTM but can we have descriptions for PRs?

e.g. what is wft, why are we adding this, etc.

rukawata:
> Code LGTM but can we have descriptions for PRs?
> 
> e.g. what is wft, why are we adding this, etc.

wft is the branch with new code for deployment.
In the future, we would have separated pipelines for the old code and the new code.
This yaml was generated when I tried to create a new pipeline for this Github repo.

Sorry, I didn't know that the yaml was actually generated - the page to design pipeline actually showed me an error: "Unable to configure a service on the selected Github repository." - which I wanted to discuss that with @persiaAziz on Tuesday...

hook_auth

But anyway, if the file does get generated, I will start working on this yaml directly.

azure-pipelines-1.yml
rukawata: > What exactly triggers this pipeline? (Tag, certain folder, ...?)

trigger would be any changes in wft branch.
But I don't think it would actually work - see my earlier [comments ](https://github.com/Azure/sap-hana/pull/196#issuecomment-518317025)

persiaAziz: Do you plan to grow this PR after wft's change gets merged? I personally don't think this needs to be a part of the repo yet.

azure-pipelines-1.yml
```diff
@@ -0,0 +1,19 @@
...
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- script: echo Hello, world!
```
persiaAziz: Let's wait for wft's change to get in. Then we can build a useful pipeline. Alternatively you can include "v2" as the includes path, so that this pipeline gets triggered for changes under the v2 directory

PR #202, opened 2019-08-09
Add Azure Pipeline for deploy/v2

rukawata:
Adding below test case for code submitted under `deploy/v2`:

1. allNew: create all resources from scratch
2. reuseRG: use existing resource group (generated inside the job)
3. reuseVNET: use existing vnet (management generated from allNew)
4. reuseNSG: use existing NSG (nsg geneareted from allNew)

Azure pipeline uses `azure-pipelines-v2.yaml`.
Input templates are stored under `templates/tempGen`.

All succeeded/failed/cancelled pipeline tests, will trigger `cleanUp` job to keep the subscription clean.

For PRs:

- That have code change(s) under `deploy/v2`
- Against master
- Contain new parameter in the JSON file

Please adjust  `templates/tempGen/template.json` accordingly in the PR to pass this test.


azure-pipelines-wft.yaml
```diff
@@ -0,0 +1,101 @@
...
+- stage: CreatingResources
+  jobs:
+  - job: createAllNew
+    pool:
+      vmImage: "ubuntu-16.04"
+    steps:        
+    - template: templates/terraform-deployment-steps.yaml
+      parameters:
+        inputJSONFile: '../pipelineInputJSON/input.auto.tfvars.allNew.json'
+        testCaseName: 'wft-allNew'
```
persiaAziz: Can we replace all the 'wft's with 'v2'

templates/terraform-deployment-steps.yaml
```diff
@@ -0,0 +1,48 @@
...
+          persistCredentials: true
+        - script: |
+            pip install packaging
+            pip install --ignore-installed pyyaml ansible[azure]
+            sudo -H pip install msrest==0.6.0
+          displayName: 'Install python tools'
+        - script: |
+            ssh-keygen -b 4096 -t rsa -f /tmp/sshkey -q -N ""
+            cd deploy/v2/terraform
+            sed -i "s|VAR_AZ_RG_NAME|${{parameters.testCaseName}}-$(Build.BuildId)|g" $VAR_FILE_NAME
```
persiaAziz: Can this be done using a template? It might be a little bit of code, but I think it will be easier to maintain.

rukawata: > Can this be done using a template? It might be a little bit of code, but I think it will be easier to maintain.

Which part? sed? or the whole script section? `terraform-deployment-steps.yaml` itself is a template.

persiaAziz: I mean instead of calling sed, can we use the template task of the pipeline. This way the pipeline script will be easier to read and maintain. That means instead of having nested templates, we have variables for RG, NSG etc, and we replace the value using template.

rukawata: Unfortunately to avoid sed or any type of string replacement is not possible with the current design of this pipeline. Below are the two dynamic elements that require string replacement:
- All resource groups' life span is within the test, therefore the resource group names are generated during the testing and is tight to specific buildId.
- All the existing resources (RG, VNET, NSG) are generated within the test, therefore it is only possible to retrieve their arm id during the test.

But you are absolutely right the names are not completely clear. Therefore I:
- Restructured the code
- Add a python script to update the template JSON
- Add a [README.MD](https://github.com/Azure/sap-hana/blob/deac9b2cc14930e0c27b41004e142aa6e2b4ca54/templates/README.MD) with the design as well as instructions on how to add new test scenarios.

Hope with the above change, it will be easier to read and maintain in the future :)

azure-pipelines-v2.yaml
```diff
@@ -0,0 +1,166 @@
...
+      parameters:
+        scenario: $(scenario)
+        testCaseName: $(testcase)
+        placeHolder: nsg-arm-id
+        value: $(arm_id)
+    - template: templates/terraform-deployment-steps.yaml
+      parameters:
+        testCaseName: $(testcase)
+#####################################################################
+# This stage destroy resources crated from all above test cases.    #
```
pabowers: typo in the comment. please fix to `created` in place of `crated`

rukawata: Thanks! Done.

rukawata:
Adding below test case for code submitted under `deploy/v2`:

1. allNew: create all resources from scratch
2. reuseRG: use existing resource group (generated inside the job)
3. reuseVNET: use existing vnet (management generated from allNew)
4. reuseNSG: use existing NSG (nsg geneareted from allNew)

Azure pipeline uses `azure-pipelines-v2.yaml`.
Input templates are stored under `templates/tempGen`.

All succeeded/failed/cancelled pipeline tests, will trigger `cleanUp` job to keep the subscription clean.

For PRs:

- That have code change(s) under `deploy/v2`
- Against master
- Contain new parameter in the JSON file

Please adjust  `templates/tempGen/template.json` accordingly in the PR to pass this test.

PR #204, opened 2019-08-15
v2: Added module to create jumpbox and RTI box

suhanselvan-wft:
Added new module to create the jumpboxes and RTI box.


deploy/v2/input.tfvars.json
```diff
@@ -56,5 +56,62 @@
                 }
             }
         }
+    },
+
+    "jumpboxes": {
+        "windows_jumpbox": {
```
tniek: I see that you've changed the structure of the jumpbox section slightly.
If you're now using a map (instead of a list) and store the jumpbox names in the keyss, why is there still a `name` key/value-pair?

deploy/v2/terraform/main.tf
```diff
@@ -9,3 +9,14 @@ module "common_infrastructure" {
...
+
+# Create Jumpboxes and RTI box
+module "jumpbox" {
+  source         = "./modules/jumpbox"
+  infrastructure = var.infrastructure
+  jumpboxes      = var.jumpboxes
+  rg             = module.common_infrastructure.rg
+  subnet-mgmt    = module.common_infrastructure.subnet-mgmt
+  nsg-mgmt       = module.common_infrastructure.nsg-mgmt
+
```
tniek: nit: remove blank line

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
+##################################################################################################################
+# JUMPBOXES
+##################################################################################################################
+
+# BOOT DIAGNOSTICS ===============================================================================================
+
+# Generates random text for boot diagnostics storage account name
```
tniek: Instead of just using a random string, would it make sense to associate this with something more descriptive (e.g. underlying VM name with timestamp), or have the customer specify it altogether?

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+  name                     = "diag${random_id.random-id.hex}"
+  resource_group_name      = var.rg[0].name
+  location                 = var.rg[0].location
+  account_replication_type = "LRS"
+  account_tier             = "Standard"
+}
+
+# NETWORK SECURITY RULES =========================================================================================
+
+# Creates jumpbox rdp network security rule
```
tniek: nit: "RDP"

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+resource "azurerm_network_security_rule" "nsr-rdp" {
```
tniek: This rule should only be created for Windows jumpboxes.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+  direction                   = "Inbound"
+  access                      = "allow"
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = 3389
+  source_address_prefix       = "*"
+  destination_address_prefix  = "*"
+}
+
+# Creates jumpbox ssh network security rule
```
tniek: nit: "SSH"

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+resource "azurerm_network_security_rule" "nsr-ssh" {
```
tniek: This rule should only be created for Linux jumpboxes (normal ones and RTI).

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+resource "azurerm_network_security_rule" "nsr-rdp" {
+  name                        = "rdp"
+  resource_group_name         = var.rg[0].name
+  network_security_group_name = var.nsg-mgmt[0].name
+  priority                    = 101
+  direction                   = "Inbound"
+  access                      = "allow"
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = 3389
```
tniek: Will we need a separate rule for winrm?

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = 22
+  source_address_prefix       = "*"
+  destination_address_prefix  = "*"
+}
+
+# NICS ============================================================================================================
+
+# Creates the jumpbox nic and ip
```
tniek: nit: "NIC and IP address"

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+resource "azurerm_network_interface" "nic-primary" {
+  for_each                      = var.jumpboxes
+  name                          = "${each.value.name}-nic1"
```
tniek: Not so nice that name is hard-coded; what if customers want to follow a different naming convention for their Azure resources?

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+    name                          = "${each.value.name}-nic1-ip"
+    subnet_id                     = var.subnet-mgmt[0].id
+    private_ip_address            = var.infrastructure.vnets.management.subnet_mgmt.is_existing ? each.value.private_ip_address : each.key == "linux_jumpbox" ? lookup(var.jumpboxes.linux_jumpbox, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 4) : each.key == "windows_jumpbox" ? lookup(var.jumpboxes.windows_jumpbox, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 5) : each.key == "rti_box" ? lookup(var.jumpboxes.rti_box, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 6) : null
+    private_ip_address_allocation = "static"
+  }
+}
+
+# VIRTUAL MACHINES ================================================================================================
+
+# Creates linux vm
```
tniek: nit: "Linux VM"

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+      path     = "/home/${each.value.authentication.username}/.ssh/authorized_keys"
+      key_data = file(each.value.authentication.path_to_public_key)
+    }
+  }
+
+  boot_diagnostics {
+    enabled     = true
+    storage_uri = azurerm_storage_account.storageaccount-bootdiagnostics.primary_blob_endpoint
+  }
+
```
tniek: nit: remove blank line

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+}
+
+# Creates windows vm
```
tniek: nit: "Windows VM"

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  os_profile_windows_config {
+  }
+
+  boot_diagnostics {
+    enabled     = true
+    storage_uri = azurerm_storage_account.storageaccount-bootdiagnostics.primary_blob_endpoint
+  }
+
```
tniek: nit: remove blank line

deploy/v2/terraform/modules/jumpbox/variables.tf
```diff
@@ -0,0 +1,19 @@
...
+variable "rg" {
+  description = "Details of the resource group"
+}
+
+variable "subnet-mgmt" {
+  description = "Details of the management subnet"
+}
+
+variable "nsg-mgmt" {
+  description = "Details of the management nsg"
```
tniek: nit: "NSG"

deploy/v2/input.tfvars.json
suhanselvan-wft: I tried my best to preserve the original structure. But Terraform requires different resource blocks for Windows and Linux VMs. So it became hard to loop with index numbers. Also I had to use for and foreach expressions which do not support list of objects. 
The key `"windows_jumpbox"` in this case is static and used for identifying the VM by Terraform. The `name` key can be used by the user to assign a custom name for the VM.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: The storage account will be common for all the VMs in the deployment. So yes, we can add another user input for this.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: RDP is blocked by default by an NSG. So we'll need it.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Then we can add another sub-section for NIC in the input file for the users to specify that. Is this fine?

deploy/v2/terraform/modules/jumpbox/variables.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/input.tfvars.json
```diff
@@ -56,5 +56,62 @@
...
+	    },
+	    "authentication": {
+	        "type": "",
+	        "username": "",
+	        "use_existing": "",
+	        "path_to_public_key": "",
+	        "path_to_private_key": ""
+	    },
+	    "components": []
+       	}
```
rukawata: Can you also include JSON sample templates for each scenario with default values in, therefore it's easier for anybody to test the solution - they don't have to fill out all variables from scratch and also can be used as a reference

deploy/v2/terraform/main.tf
suhanselvan-wft: Fixed.

deploy/v2/input.tfvars.json
suhanselvan-wft: Sure. Where would you want the location of these templates to be?

rukawata: Let's store them under `deploy/v2/templateSamples` for now.
We can always move it to a different location later on.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/input.tfvars.json
suhanselvan-wft: This has been added. The only variables that the user will have to change before running will be the ssh key location. I tried having a default location for them and creating the ssh key in that location using local-exec provisioner but Terraform requires the ssh key file to exist before `terraform apply`.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,160 @@
...
+    subnet_id                     = var.subnet-mgmt[0].id
+    private_ip_address            = var.infrastructure.vnets.management.subnet_mgmt.is_existing ? each.value.private_ip_address : each.key == "linux_jumpbox" ? lookup(var.jumpboxes.linux_jumpbox, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 4) : each.key == "windows_jumpbox" ? lookup(var.jumpboxes.windows_jumpbox, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 5) : each.key == "rti_box" ? lookup(var.jumpboxes.rti_box, "private_ip_address", false) != false ? each.value.private_ip_address : cidrhost(var.infrastructure.vnets.management.subnet_mgmt.prefix, 6) : null
+    private_ip_address_allocation = "static"
+  }
+}
+
+# VIRTUAL MACHINES ================================================================================================
+
+# Creates Linux VM
+resource "azurerm_virtual_machine" "vm-linux" {
```
rukawata: All the jumpboxes do not have public IP. Is it intended?
Without public IP, we can't really use the jumpbox.

suhanselvan-wft: Oops. Missed that. Will add them

suhanselvan-wft: Public IP has been added. Thanks for the catch.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Fixed.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,170 @@
...
+}
+
+# NETWORK SECURITY RULES =========================================================================================
+
+# Creates Windows jumpbox RDP network security rule
+resource "azurerm_network_security_rule" "nsr-rdp" {
+  count                       = lookup(var.jumpboxes, "windows_jumpbox", false) != false ? 1 : 0
+  name                        = "rdp"
+  resource_group_name         = var.rg[0].name
+  network_security_group_name = var.nsg-mgmt[0].name
```
rukawata: > resource_group_name         = var.rg[0].name
  network_security_group_name = var.nsg-mgmt[0].name

The above lines won't work if an existing NSG from a different RG is used.
Same applies to both `nsr-rdp`, `nsr-ssh`.

suhanselvan-wft: This has been fixed. They will use the same resource group as NSG.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,170 @@
...
+  name                        = "rdp"
+  resource_group_name         = var.nsg-mgmt[0].resource_group_name
+  network_security_group_name = var.nsg-mgmt[0].name
+  priority                    = 101
+  direction                   = "Inbound"
+  access                      = "allow"
+  protocol                    = "Tcp"
+  source_port_range           = "*"
+  destination_port_range      = 3389
+  source_address_prefix       = "*"
```
rukawata: Not sure if this has been discussed.
Do we want to open the jumpbox to all? Or should we add a list variable with IP ranges that are allowed to use the jumpbox?

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: There could potentially be another problem that we need to discuss how to handle:
If the existing NSG has already the same rule with same priority, the deployment will fail:
>Error Creating/Updating Network Security Rule "linux_jumpbox-ssh" (NSG "nsg-mgmt" / Resource Group "v2-allNew-321"): network.SecurityRulesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="SecurityRuleConflict" Message="Security rule Cleanuptool-Allow-102 conflicts with rule linux_jumpbox-ssh. Rules cannot have the same Priority and Direction." Details=

rukawata: As discussed, please add logic to not add security if an existing NSG is used.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: As discussed, please add variable "allowed_ips" under subnet_mgmt -> nsg.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,170 @@
...
+# VIRTUAL MACHINES ================================================================================================
+
+# Creates Linux VM
+resource "azurerm_virtual_machine" "vm-linux" {
+  for_each                      = { for k, v in var.jumpboxes : (k) => (v) if replace(v.os.publisher, "Windows", "") == v.os.publisher }
+  name                          = each.value.name
+  location                      = var.rg[0].location
+  resource_group_name           = var.rg[0].name
+  network_interface_ids         = [azurerm_network_interface.nic-primary[each.key].id]
+  vm_size                       = "Standard_D4s_v3"
```
rukawata: As discussed, please add a variable for vm_size.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: This has been added.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: This has been added.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: This has been added.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: The user will now have an option to give a custom name for the boot diagnostics storage account.

deploy/v2/terraform/modules/jumpbox/main.tf
tniek: What I meant was:
Will we need a separate rule for the winrm protocol, used to remotely execute commands on the Windows jumpbox (in the absence of a protocol like SSH)?

deploy/v2/terraform/modules/jumpbox/main.tf
tniek: Yes; and make it optional, so the user doesn't have to specify if they're okay with the default name.

deploy/v2/input.tfvars.json
```diff
@@ -23,7 +23,8 @@
                     "nsg" : {
                         "is_existing" : "",
                         "arm_id" : "",
-                        "name" : ""
+                        "name" : "",
+                        "allowed_ips" : ""
```
rukawata: Thanks for making the requested changes.
Please read [README](https://github.com/Azure/sap-hana/blob/master/templates/README.MD) for new azure pipeline test, and update [template](https://github.com/Azure/sap-hana/blob/master/templates/tempGen/template.json) to pass the test with your new PRs.
You can submit the changes of the template as part of this PR.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: I'm not sure we'll need that. Have to discuss.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: Okay. This has been added.

deploy/v2/input.tfvars.json
suhanselvan-wft: I've updated the input template file with the required variables. But the reuseVNET and reuseNSG test scenarios are failing because the respective place holder ARM IDs are not being populated in the input file.

rukawata: There seems to be some sporadic problem with clean up job of the Azure Pipeline. - I will look into that. After a manual clean up, I triggered your build again, it finishes successfully :)

deploy/v2/input.tfvars.json
suhanselvan-wft: The code has been updated based on the new data structure for jumpboxes. 
@rukawata since we are having two separate resource blocks for Windows and Linux NICs, I'm making an assumption that a max of five windows instances will be created by the user. This is to avoid overlapping of private IP addresses when the code is assigning them. Let me know if this threshold needs to be increased.

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+    },
+
+    "jumpboxes" : {
+        "windows" : [
+            {
+            "name" : "jumpbox-windows",
+            "destroy_after_deploy" : "false",
+            "size" : "Standard_D4s_v3",
```
rukawata: We have a quota (100) for cores in the test pipeline subscription.

With `"size" : "Standard_D4s_v3"`, 4 cores per VM, 3 VMs per terraform deployment job, 4 jobs in total:
We easily use 4 * 3 * 4 = 48 cores per build.

Of course we do have clean up jobs in the pipeline to take care of the resources that are been created. However, if we have more than one PRs submitted at the same time, meaning more than one build will be running, we easily exceed the quota.

This is also why your PRs didn't pass the recent tests:
>2019-08-21T03:48:48.3769799Z [1m[31mError: [0m[0m[1mcompute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="OperationNotAllowed" Message="Operation results in exceeding quota limits of Core. Maximum allowed: 100, Current in use: 100, Additional requested: 4. Please read more about quota increase at https://aka.ms/ProdportalCRP/?#create/Microsoft.Support/Parameters/{\"subId\":\"***\",\"pesId\":\"15621\",\"supportTopicId\":\"32447243\"}."[0m

Therefore, please change the size in the template to a smaller one if possible.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,195 @@
...
+
+# NICS ============================================================================================================
+
+# Creates the public IP addresses for Windows VMs
+resource "azurerm_public_ip" "public-ip-windows" {
+  count               = length(var.jumpboxes.windows)
+  name                = "${var.jumpboxes.windows[count.index].name}-public-ip"
+  location            = var.rg[0].location
+  resource_group_name = var.rg[0].name
+  allocation_method   = "Static"
```
rukawata: I see that you implement for all public IPs and private IPs as static.
The dynamic IPs has lower cost. Is there a specific reason for using static IPs?
Also, if you want to use static IPs, **SKUs are required**, see [here](https://www.terraform.io/docs/providers/azurerm/r/public_ip.html#sku).
>Note Public IP Standard SKUs require allocation_method to be set to Static.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -0,0 +1,195 @@
...
+  resource_group_name           = var.rg[0].name
+  network_interface_ids         = [azurerm_network_interface.nic-primary-linux[count.index].id]
+  vm_size                       = var.jumpboxes.linux[count.index].size
+  delete_os_disk_on_termination = "true"
+
+  storage_os_disk {
+    name              = "${var.jumpboxes.linux[count.index].name}-osdisk"
+    caching           = "ReadWrite"
+    create_option     = "FromImage"
+    managed_disk_type = "StandardSSD_LRS"
```
rukawata: Disk type should also be adjustable by the custom input JSON.
Please change that to all disk types for VMs.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: The IPs have to be static as per the requirement given. As for SKU, it defaults to basic which meets our requirements.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan-wft: The change has been made.

deploy/v2/template_samples/single_node_hana.tfvars.json
suhanselvan-wft: I've reduced the size to D2s_v3 for the jumpboxes. We'll also have to increase the quota because the HANA nodes will have minimum 32 cores each.

deploy/v2/input.tfvars.json
rukawata: The only place I see you have this 5 hard limit is when you assign destination_address_prefix.
This does not make too much sense to me for two reasons:

- It makes not much sense to have several rules for SSH to different IPs. I've sent this question to Marshal.
- With your logic, in RDP rule, only the last private IP is in the destination IP address (eg. 10.0.1.5). This will be a problem when we have a second windows VM (eg. 10.0.1.4) - this RDP rules will not allow access to this VM.

We can wait until Marshal's response, then change the way it's been implemented.

suhanselvan-wft: The RDP rule has been fixed. When the code assigns the private IPs, the Windows jumpboxes will have IPs assigned from the 4th IP of the subnet. The Linux boxes will have IPs assigned from the 9th IP of the subnet. We'll have to document it so that the user doesn't create more than 5 Windows jumpboxes. It will fail otherwise.

rukawata: Is there a reason why we need different security rules for different VMs? All VM will use the same NSG anyway, does it make more sense to just include a CIDR range instead? Then we will have only one rule for RDP, one rule for SSH.

I feel with the current way of creating rules that we have to limit the number of VMs does not make too much sense.

suhanselvan-wft: I agree. Initially the assumption was one Linux and Windows jumpbox. Considering the case now, I can change the destination prefix of the rules to the CIDR range of the management subnet.

Even if we change the rules, the limit on the number of Windows VMs will remain the same.

suhanselvan-wft: @rukawata  I've changed the destination address of the rules to the management subnet prefix. Let me know if this is fine.

rukawata: > I agree. Initially the assumption was one Linux and Windows jumpbox. Considering the case now, I can change the destination prefix of the rules to the CIDR range of the management subnet.
> 
> Even if we change the rules, the limit on the number of Windows VMs will remain the same.

I do not see why we have to have this limit of 5 windows VM.
You could simply add the number of windows VM's by length of windows VM list.
Something like below:

`cidrhost(azurerm_subnet.subnet.address_prefix, count.index + 4 + length(var.jumpboxes.windows))`

suhanselvan-wft: I'm already doing that by adding count.index to 4. As I said in a previous comment above, the Windows NIC resource block will assign private IPs starting from 4th IP of the subnet. The Linux NIC resource block will assign private IPs starting from 9th IP of the subnet. So if we consider the case where the user wants 6 Windows jumpboxes and 2 Linux jumpboxes, the 6th Windows jumpbox and the 1st Linux jumpbox will request the 9th IP of the subnet and Terraform apply will fail.

rukawata: > I'm already doing that by adding count.index to 4. As I said in a previous comment above, the Windows NIC resource block will assign private IPs starting from 4th IP of the subnet. The Linux NIC resource block will assign private IPs starting from 9th IP of the subnet. So if we consider the case where the user wants 6 Windows jumpboxes and 2 Linux jumpboxes, the 6th Windows jumpbox and the 1st Linux jumpbox will request the 9th IP of the subnet and Terraform apply will fail.

I think add 4 is to avoid using the reserved IP [link](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-faq)
Let's say we have n windows and m linux VMs:

- windows VM will get 4 to 4+n-1
- and linux VM will get from 4+n to 4+m-1

*n = length(var.jumpboxes.windows)
*m = length(var.jumpboxes.linux)

This means, we will have something like below for windows VM private IP:
`cidrhost(azurerm_subnet.subnet.address_prefix, count.index + 4)`

And something like below for linux VM private IP:
`cidrhost(azurerm_subnet.subnet.address_prefix, count.index + 4 + length(var.jumpboxes.windows))`

I've tested this approach with 6 windows VMs and 3 linux VMs and worked.

suhanselvan-wft: That'll work. But it will also increase the possibility for error if the subnet is small. Its a choice to make.

rukawata: If the end users plan to have that many VMs, they are the ones should make sure the provided subnet has sufficient IP address. So it's not a problem.

Please go ahead and apply this change.

suhanselvan-wft: Okay. This change has been made.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: As discussed, we do not need WinRM. Only two ports are required to open.

PR #205, opened 2019-08-17
Add workflow graph to README.MD

rukawata:
Add workflow graph to README.MD for better understanding of the new pipeline.


rukawata:
> Thanks for working on this.
> 
> Is there no scenario reuseSubnet? Otherwise, looks good

Subnet is kind of property of vnet. If they bring existing subnet, they will have to use existing vnet as well.
In all scenarios, we only consider using ONE existing resource for the time being, therefore reuseSubnet is not included at the moment.

PR #207, opened 2019-08-19
Exclude code paths that do not required v2 check

rukawata:
Explicitly exclude paths that do not need to be checked by azure-pipeline-v2.

### DIFF
--- a/azure-pipelines-v2.yaml
+++ b/azure-pipelines-v2.yaml
@@ -10,6 +10,10 @@ pr:
   paths:
     include:
     - deploy/v2/*
+    exclude:
+    - README.md
+    - monitor/*
+    - deploy/*
 variables:
   - group: azure-config-variables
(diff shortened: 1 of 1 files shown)


azure-pipelines-v2.yaml
```diff
@@ -10,6 +10,10 @@ pr:
   paths:
     include:
     - deploy/v2/*
+    exclude:
```
tniek: Not quite sure on the logic here:
So we're explicitly including v2 only - but at the same time excluding others that are already mutually exclusive anyway? Why is this needed?

rukawata: > Not quite sure on the logic here:
> So we're explicitly including v2 only - but at the same time excluding others that are already mutually exclusive anyway? Why is this needed?

I agree this is confusing. But simply include deploy/v2/* does not exclude other folders...that is why this PR is needed.

PR #210, opened 2019-08-21
Exclude folders from azure pipeline testing

rukawata:
The code under folders `tools` and `templates` are irrelevant to single HANA deployment.
Therefore exclude them from azure pipeline testing.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,6 +13,8 @@ pr:
     - README.md
     - monitor/*
     - deploy/v2/*
+    - templates/*
+    - tools/*
 jobs:
 - job: single_node_hana
   variables:


azure-pipelines.yml
```diff
@@ -13,6 +13,8 @@ pr:
     - README.md
     - monitor/*
     - deploy/v2/*
+    - templates/*
```
lagalbra: According to the README, the templates folder holds the information to run Azure pipeline testing, so if it changes, the testing should be run again to make sure there are no regressions. Is my understanding of what that folder contains correct?

rukawata: Sorry, I know this is a little confusing. 
We have two pipelines for testing two sets of code:

- `azure-pipelines.yml` => **Azure.sap-hana**
- `azure-pipelines-v2.yaml` => **Azure.sap-hana.v2**

`templates/*` is only used for azure-pipelines-v2.yaml, therefore I exclude it from azure-pipelines.yml.

PR #217, opened 2019-08-31
v2:Added module to create HANA database nodes

suhanselvan:
This PR pertains to the v2 implementation and includes module to create the HANA database nodes.

### DIFF
--- a/deploy/v2/hdb_sizes.json
+++ b/deploy/v2/hdb_sizes.json
@@ -1,7 +1,7 @@
 {
     "S" : {
         "compute" : {
-            "vm_size" : "Standard_M32ls"
+            "vm_size" : "Standard_D4s_v3"
         },
         "storage" : [
             {
--- a/deploy/v2/hdb_sizes.json
+++ b/deploy/v2/hdb_sizes.json
@@ -1,7 +1,8 @@
 {
     "S" : {
         "compute" : {
-            "vm_size" : "Standard_D4s_v3"
+            "vm_size" : "Standard_D4s_v3",
+	    "swap_size_gb": 2
         },
         "storage" : [
             {
--- a/deploy/v2/hdb_sizes.json
+++ b/deploy/v2/hdb_sizes.json
@@ -3,204 +3,54 @@
         "compute" : {
             "vm_size" : "Standard_M32ls"
         },
-        "storage" : {
-            "os" : {
+        "storage" : [
+            {
+                "name" : "os",
+                "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
                 "size_gb" : 64,
                 "caching" : "ReadWrite"
             },
-            "data_log" : {
-                "count" : 3,
+            {
+                "name" : "data",
+                "count" : 1,
                 "disk_type" : "Premium_LRS",
                 "size_gb" : 512,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
-            "shared" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 512,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "sap" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "backup" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 512,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            }
-        }
-    },
-
-    "M" : {
-        "compute" : {
-            "vm_size" : "Standard_M64ls"
-        },
-        "storage" : {
-            "os" : {
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadWrite"
-            },
-            "data_log" : {
-                "count" : 3,
-                "disk_type" : "Premium_LRS",
-                "size_gb" : 512,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "shared" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 512,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "sap" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "backup" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 512,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            }
-        }
-    },
-
-    "L" : {
-        "compute" : {
-            "vm_size" : "Standard_M64s"
-        },
-        "storage" : {
-            "os" : {
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadWrite"
-            },
-            "data_log" : {
+            {
+                "name" : "log",
                 "count" : 2,
                 "disk_type" : "Premium_LRS",
-                "size_gb" : 1024,
+                "size_gb" : 512,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
-            "shared" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "sap" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "backup" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            }
-        }
-    },
-
-        "XL" : {
-        "compute" : {
-            "vm_size" : "Standard_M128s"
-        },
-        "storage" : {
-            "os" : {
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 128,
-                "caching" : "ReadWrite"
-            },
-            "data_log" : {
+            {
+                "name" : "shared",
                 "count" : 3,
-                "disk_type" : "Premium_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "shared" : {
-                "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
+                "size_gb" : 512,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
-            "sap" : {
+            {
+                "name" : "logbackup",
                 "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
                 "size_gb" : 64,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
-            "backup" : {
+            {
+                "name" : "databackup",
                 "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
+                "size_gb" : 512,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             }
-        }
-    },
-
-    "XXL" : {
-        "compute" : {
-            "vm_size" : "Standard_M128ms"
-        },
-        "storage" : {
-            "os" : {
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 128,
-                "caching" : "ReadWrite"
-            },
-            "data_log" : {
-                "count" : 5,
-                "disk_type" : "Premium_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "shared" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "sap" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 64,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            },
-            "backup" : {
-                "count" : 1,
-                "disk_type" : "StandardSSD_LRS",
-                "size_gb" : 1024,
-                "caching" : "ReadOnly",
-                "write_accelerator" : false
-            }
-        }
+        ]
     }
 }
--- a/deploy/v2/hdb_sizes.json
+++ b/deploy/v2/hdb_sizes.json
@@ -1,0 +1,216 @@
+{
+    "sizes" : {
+        "S" : {
+            "compute" : {
+		"vm_size" : "Standard_M32ls"
+            },
+            "storage" : {
+                "os" : {
+	            "size_gb" : 64
+    		},
+		"data_log" : {
+		    "count" : 3,
+    		    "disk_type" : "Premium_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"shared" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"sap" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 64,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"backup" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		}
+            },
+	    "network" : {
+                "accelerated_networking" : true
+	    }
+        },
+
+        "M" : {
+            "compute" : {
+		"vm_size" : "Standard_M64ls"
+            },
+            "storage" : {
+                "os" : {
+	            "size_gb" : 64
+    		},
+		"data_log" : {
+		    "count" : 3,
+    		    "disk_type" : "Premium_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"shared" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"sap" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 64,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"backup" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 512,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		}
+            },
+	    "network" : {
+                "accelerated_networking" : true
+	    }
+        },
+
+        "L" : {
+            "compute" : {
+		"vm_size" : "Standard_M64s"
+            },
+            "storage" : {
+                "os" : {
+	            "size_gb" : 64
+    		},
+		"data_log" : {
+		    "count" : 2,
+    		    "disk_type" : "Premium_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"shared" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"sap" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 64,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"backup" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		}
+            },
+	    "network" : {
+                "accelerated_networking" : true
+	    }
+        },
+
+        "XL" : {
+            "compute" : {
+		"vm_size" : "Standard_M128s"
+            },
+            "storage" : {
+                "os" : {
+	            "size_gb" : 128
+    		},
+		"data_log" : {
+		    "count" : 3,
+    		    "disk_type" : "Premium_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"shared" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"sap" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 64,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"backup" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		}
+            },
+	    "network" : {
+                "accelerated_networking" : true
+	    }
+        },
+
+        "XXL" : {
+            "compute" : {
+		"vm_size" : "Standard_M128ms"
+            },
+            "storage" : {
+                "os" : {
+	            "size_gb" : 128
+    		},
+		"data_log" : {
+		    "count" : 5,
+    		    "disk_type" : "Premium_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"shared" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"sap" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 64,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		},
+		"backup" : {
+		    "count" : 1,
+    		    "disk_type" : "StandardSSD_LRS",
+		    "size_gb" : 1024,
+		    "caching" : "ReadOnly",
+		    "write_accelerator" : false
+		}
+            },
+	    "network" : {
+                "accelerated_networking" : true
+	    }
+        }
+    }
+}
+
+	
+		
--- a/deploy/v2/hdb_sizes.json
+++ b/deploy/v2/hdb_sizes.json
@@ -1,216 +1,206 @@
 {
-    "sizes" : {
-        "S" : {
-            "compute" : {
-		"vm_size" : "Standard_M32ls"
-            },
-            "storage" : {
-                "os" : {
-	            "size_gb" : 64
-    		},
-		"data_log" : {
-		    "count" : 3,
-    		    "disk_type" : "Premium_LRS",
-		    "size_gb" : 512,
-		    "caching" : "ReadOnly",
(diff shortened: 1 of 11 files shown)


deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
+{
+    "S" : {
+        "compute" : {
+            "vm_size" : "Standard_M32ls"
```
rukawata: In the template that we designed, I see `D8s_v3`.
Is there any reason to change to `Standard_M32ls`?
Plus, the reason why v2 pipeline failed is because we don't have quota for **standardMSFamily** Cores.
We need to decide which are the correct vm_size, then I can request quota from forward.
If you have double/questions, please check with Marschal on this.

deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
...
+        },
+        "storage" : {
+            "os" : {
+                "disk_type" : "StandardSSD_LRS",
+                "size_gb" : 64,
+                "caching" : "ReadWrite"
+            },
+            "data_log" : {
```
rukawata: Please do not merge data and log sections.
There is a reason we have tags for data, log, shared.
This is due to the nature of HANA installation.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  storage_image_reference {
+    publisher = var.databases[count.index].os.publisher
+    offer     = var.databases[count.index].os.offer
+    sku       = var.databases[count.index].os.sku
+    version   = "latest"
+  }
+
+  dynamic "storage_data_disk" {
```
rukawata: It is not good to have storage_data_disk 4 times for different disk types.
Please change this maybe after the definition of disks is set.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
+##################################################################################################################
+# HANA Database Node
+##################################################################################################################
+
+# NETWORK SECURITY RULES =========================================================================================
+
+# Creates network security rule to deny external traffic for SAP admin subnet
```
rukawata: Why would the database subnet deny traffic from admin subnet?
Isn't the whole purpose is to only grant access to database subnet from admin subnet?

deploy/v2/hdb_sizes.json
suhanselvan: The template we designed was to define the structure of the input file. The actual values are based on the configuration provided for demo systems in the reference architecture document shared with us. I'm attaching it here for your reference.

https://github.com/Azure/sap-hana/files/3562985/S4HANA.Reference.Architecture.docx

deploy/v2/hdb_sizes.json
suhanselvan: Again, this was done based on the disk configuration provided in the reference architecture document.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: I'm not denying any traffic to the SAP DB subnet.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: I'm sorry, can you please elaborate on the issue with this implementation? I do not see any issue in Terraform for having multiple `storage_data_disk` blocks. Let me know if I'm missing something.

rukawata: As discussed, please reduce the redundant code.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
...
+
+# Creates HANA database VM
+resource "azurerm_virtual_machine" "vm-hdb" {
+  count                         = length(var.databases)
+  name                          = var.databases[count.index].name
+  location                      = var.resource-group[0].location
+  resource_group_name           = var.resource-group[0].name
+  primary_network_interface_id  = azurerm_network_interface.nic-hdb-db[count.index].id
+  network_interface_ids         = [azurerm_network_interface.nic-hdb-admin[count.index].id, azurerm_network_interface.nic-hdb-db[count.index].id]
+  vm_size                       = var.databases[count.index].size == "S" ? local.sizes.S.compute.vm_size : var.databases[count.index].size == "M" ? local.sizes.M.compute.vm_size : var.databases[count.index].size == "L" ? local.sizes.L.compute.vm_size : var.databases[count.index].size == "XL" ? local.sizes.XL.compute.vm_size : var.databases[count.index].size == "XXL" ? local.sizes.XXL.compute.vm_size : null
```
rukawata: Instead of hard-coded logic using the sizes S, M, L, have you try to use lookup function with the local variable sizes.

I've tested in a small example and worked. In your code, it would be something like:
`lookup(local.sizes, "${var.databases[count.index].size}").compute.vm_size`

By doing this, we make the code more readable and also easier to extend when we have add more to the size matrix.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
...
+    private_ip_address            = var.infrastructure.vnets.sap.subnet_db.is_existing ? var.databases[count.index].db_nic.private_ip_address : lookup(var.databases[count.index].db_nic, "private_ip_address", false) != false ? var.databases[count.index].db_nic.private_ip_address : cidrhost(var.infrastructure.vnets.sap.subnet_db.prefix, (count.index + 4))
+    private_ip_address_allocation = "static"
+  }
+}
+
+# VIRTUAL MACHINES ================================================================================================
+
+# Creates HANA database VM
+resource "azurerm_virtual_machine" "vm-hdb" {
+  count                         = length(var.databases)
```
rukawata: What's the scenario we implement here?
In a HA or multi-node scenario, we would need more than one VMs for one database. This is not implemented here.
I see in the input JSON template @tniek provided, there is a section `"nodes"` - this defines how many VMs in the database. 

```
"nodes": [
			{
				"private_ip": "10.0.0.6",
				"role": "worker",
			},
			{
				"private_ip": "10.0.0.7",
				"role": "standby",
			},
	]
```

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
...
+  location                      = var.resource-group[0].location
+  resource_group_name           = var.resource-group[0].name
+  network_security_group_id     = var.nsg-db[0].id
+  enable_accelerated_networking = true
+
+  ip_configuration {
+    primary                       = true
+    name                          = "${var.databases[count.index].name}-db-nic-ip"
+    subnet_id                     = var.subnet-sap-db[0].id
+    private_ip_address            = var.infrastructure.vnets.sap.subnet_db.is_existing ? var.databases[count.index].db_nic.private_ip_address : lookup(var.databases[count.index].db_nic, "private_ip_address", false) != false ? var.databases[count.index].db_nic.private_ip_address : cidrhost(var.infrastructure.vnets.sap.subnet_db.prefix, (count.index + 4))
```
rukawata: From the input JSON @tniek provided, the private IP address are pre-assigned.
Is it still the case? Need to check with Marshal.

```
"nodes": [
			{
				"private_ip": "10.0.0.6",
				"role": "worker",
			},
			{
				"private_ip": "10.0.0.7",
				"role": "standby",
			},
	]
```

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: Interesting... I actually tried something similar by having an interpolated string like this,
`local.sizes."${var.databases[count.index].size}".compute.vm_size` and it didn't work. But now it works if its the result of a function. Anyways, thanks for the find. I'll update the code.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,181 @@
...
+    private_ip_address_allocation = "static"
+  }
+}
+
+# VIRTUAL MACHINES ================================================================================================
+
+# Creates HANA database VM
+resource "azurerm_virtual_machine" "vm-hdb" {
```
rukawata: The scenario will expend to other scenarios as well in the future, eg. SQL DB.
That's why we have `"platform": "HANA"` defined.
Therefore, the VM should not have **hdb** in the name.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: We are only implementing a single node HANA scenario right now.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: Isn't S/4 supposed to only run on HANA DB?

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: I see. Then this code is only for a single HANA node module, and we should have the code to deploy single HANA instance in `sap-hana/deploy/v2/terraform/main.tf`.
This way we can reuse the module when we deploy multiple nodes.

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: Yes, S4 only runs on HANA. But we may extend the scenario to traditional NW installation as well, which can run on other DBs. 
@mwhatleyjr can you comment on this? Or we will only have S4.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: I looked into the option of calling the data disk resource block as a module. But Terraform doesn't support loops for modules right now. Also resource blocks don't support nested loops either. I'm investigating 'unconventional' ways to achieve the result but, it doesn't seem promising to me.

deploy/v2/hdb_sizes.json
tniek: +1 on @rukawata's comment: yes, data, log and shared need to be separated.

I understand that in the original reference architecture, this was not called specifically. However, the sizing templates we built in our architecture workshop specifically call out the separated disks (see the hana-sizing.json file inside 20190711-templates.zip):
```
				"data": {
					"number_disks": 1,
					"disk_type": "Premium_LRS",
					"size_gb": 200,
					"caching": "ReadOnly",
					"throughput_mbps": 100,
					"iops": 1000,
					"write_accelerator": false,
				},
				"log": {
					"number_disks": 1,
					"disk_type": "Premium_LRS",
					"size_gb": 200,
					"caching": "ReadOnly",
					"throughput_mbps": 100,
					"iops": 1000,
					"write_accelerator": true,
				},
				"shared": {
					"number_disks": 1,
					"disk_type": "Premium_LRS",
					"size_gb": 500,
					"caching": "ReadOnly",
					"throughput_mbps": 100,
					"iops": 1000,
					"write_accelerator": true,
				},
				"logbackup": {
					"number_disks": 1,
					"disk_type": "Premium_LRS",
					"size_gb": 500,
					"caching": "ReadOnly",
					"throughput_mbps": 100,
					"iops": 1000,
					"write_accelerator": false,
				},
				"databackup": {
					"number_disks": 1,
					"disk_type": "Premium_LRS",
					"size_gb": 500,
					"caching": "ReadOnly",
					"throughput_mbps": 100,
					"iops": 1000,
					"write_accelerator": false,
				},
```
Please adjust your code accordingly (if not already done).

deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
...
+                "write_accelerator" : false
+            },
+            "sap" : {
+                "count" : 1,
+                "disk_type" : "StandardSSD_LRS",
+                "size_gb" : 64,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false
+            },
+            "backup" : {
```
tniek: See above template we designed in July; there should be a separation between data backup and log backup.

deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
...
+                "write_accelerator" : false
+            },
+            "shared" : {
+                "count" : 1,
+                "disk_type" : "StandardSSD_LRS",
+                "size_gb" : 512,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false
+            },
+            "sap" : {
```
tniek: What is this folder for in the context of HANA? SAP bits/download media?

deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
...
+                "count" : 1,
+                "disk_type" : "StandardSSD_LRS",
+                "size_gb" : 1024,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false
+            }
+        }
+    },
+
+        "XL" : {
```
tniek: nit: wrong indentation

deploy/v2/hdb_sizes.json
```diff
@@ -0,0 +1,206 @@
+{
```
tniek: Apart from the "t-shirt" sizes (S, M, L, XL, XXL), there should be (at least) one "demo" size as well, suitable for quick PoC's or demos. Compute-wise this would be a 32/64 GB D-series, storage-wise it could be a single disk for data/log/shared (assuming single-node). In this case, the Terraform modules and Ansible roles would need to be flexible enough to accommodate this scenario.

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,133 @@
...
+                "path_to_private_key" : ""
+            },
+            "components" : []
+            }
+        ]
+    },
+
+    "databases" : [
+        {
+            "name" : "",
```
tniek: In the input.json template we designed, I don't see a `name` key; what is this supposed to be used for?
(Cannot be VM name, since that would change in a HA or scale-out scenario.)

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -116,5 +116,64 @@
 	    ]
             }
         ]
-    }   
+    },
+   
+    "databases" : [
+        {
+            "name" : "hdb1",
```
tniek: See above - unclear what this refers to

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: Good catch.
@mwhatleyjr please clarify.

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: The emphasis is on "right now."
However, we need to engineer the code in a flexible way, so we can accommodate scenarios with more than one node (namely, HA pair and scale-out).

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: Again, the code needs to be flexible.
In this specific case, the potential sizes should not be hard-coded in the TF module; instead, you would determine the list of all possible values (i.e. keys in your sizing JSON).

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: While multiple instantiation of modules is (still!) not supported, you could always call the module and loop over the count inside the module.

templates/tempGen/template.json
```diff
@@ -113,5 +113,60 @@
          ]
          }
       ]   
-   }
+   },
+
+    "databases" : [
```
tniek: What exactly are we storing (redundantly?) in the `tempGen` directory?
This file seems identical to `deploy/v2/template_samples/single_node_hana.tfvars.json`.

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: This configuration file `hdb_node` should focus on creating only one server for HANA database.
Then, we will have another configuration (eg. `hdb_database`) call this `hdb_node`, pass a count (length(nodes)), therefore it creates a HANA database with several nodes.
Then, we can expend the scope to for example HA pair (eg. `hdb_ha`), where we would call `hdb_database` twice to have two sets of resources.

_*namings are subject to change._

@mwhatleyjr, @suhanselvan has a question regarding the scope - how many databases would we have? Currently the databases is defined as a list.

deploy/v2/hdb_sizes.json
rukawata: @mwhatleyjr, we have a question for you. The sap section is for /usr/sap. Do we need to have a separate disk for that?

deploy/v2/terraform/modules/hdb_node/main.tf
mwhatleyjr: The scope is for 1 db, but in v2, we need 2 hana db's, one active, the other standby, and then we will need to deploy a scale out deployment (with and w/o a standby node).  It would be best plan appropriately for these scenarios.

deploy/v2/terraform/modules/hdb_node/main.tf
mwhatleyjr: This is fine as long as it can be changed....

deploy/v2/terraform/modules/hdb_node/main.tf
mwhatleyjr: Most customers have a name scheme in place usually something like <sid>hdb01, <sid>hdb02, or something like <sid>erphdb01

deploy/v2/hdb_sizes.json
mwhatleyjr: We should - our current S4 documentation has a single P10 disk for /usr/sap

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: I think the question is more centered around having HANA + anyDB instances in parallel (think S/4 where you'd have one HANA + one ASE or MaxDB for Fiori)?

deploy/v2/terraform/modules/hdb_node/main.tf
tniek: That's why all resources names should be customizable as well.

deploy/v2/hdb_sizes.json
tniek: But we wouldn't need that for a HANA-only installation, correct? In /usr/sap, there would only be symlinks to /hana/shared/SID...

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: > I think the question is more centered around having HANA + anyDB instances in parallel (think S/4 where you'd have one HANA + one ASE or MaxDB for Fiori)?

@mwhatleyjr , can you comment on this?

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: As discussed, we will keep it this way for now.

deploy/v2/hdb_sizes.json
rukawata: > Apart from the "t-shirt" sizes (S, M, L, XL, XXL), there should be (at least) one "demo" size as well, suitable for quick PoC's or demos. Compute-wise this would be a 32/64 GB D-series, storage-wise it could be a single disk for data/log/shared (assuming single-node). In this case, the Terraform modules and Ansible roles would need to be flexible enough to accommodate this scenario.

@mwhatleyjr , can you comment on the demo size? So we can add it to the sizing json.

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: @suhanselvan please implement the logic based on whether there is value in "private_ip". Thanks!

deploy/v2/terraform/modules/hdb_node/main.tf
mwhatleyjr: Yes, I missed the context of the question, yes there will be a need to have at least a ASE or MaxDB instance in the landscape somewhere.

deploy/v2/hdb_sizes.json
mwhatleyjr: Yes, this is a need.  We should let the user select either a D8 v3, D16 v3, or D32 v3

deploy/v2/hdb_sizes.json
suhanselvan: The sizing template has been updated.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: The code has been updated to accommodate any number of disk types with any number of disks per type.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: This has been changed.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: The code has been updated to create multiple nodes for multiple databases. 
@rukawata I was able to change the map key to a number and resolve the IP address assigning issue. Thanks.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: This has been implemented.

deploy/v2/hdb_sizes.json
rukawata: Thanks! Please also add `swap_size_gb` accoding to the new hana-sizing.json we provided on 09/13/2019

suhanselvan: updated.

rukawata: We only have size S in the sizing json. Can we have the complete sizing json that includes M, L, XL...? Thanks!

suhanselvan: The latest hana-sizing.json shared does not reflect the disk type specification specified by @tniek . It still has data and log sections merged and no separate disks for data and log backup. The file provided previously didn't have info for sizes other than 'S' either. I can update once that is provided.

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -116,5 +116,64 @@
...
+                    "credentials" : {
+                    "cockpit_admin_password" : ""
+                    }
+                }
+            ],
+            "nodes" : [
+                {
+                    "private_ip" : "",
+                    "role" : ""
+                },
```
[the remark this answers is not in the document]
rukawata: for single node hana should be `worker`.
@suhanselvan please add that as well. Thanks.

deploy/v2/terraform/main.tf
```diff
@@ -19,3 +19,16 @@ module "jumpbox" {
   subnet-mgmt    = module.common_infrastructure.subnet-mgmt
   nsg-mgmt       = module.common_infrastructure.nsg-mgmt
 }
+
+# Create HANA database nodes
+module "hdb" {
+  source                         = "./modules/hdb_node"
```
rukawata: I tested and do not see any hdb node vm created.
The problem could be something is missing here.
@suhanselvan can you please check?
no_hdb_vm

suhanselvan: Screenshot (21)

I tested and it created all the VMs for me. I used the `../template_samples/single_node_hana.tfvars.json` as input file.

rukawata: I used `templates/tempGen/template.json` which looks quick similar as `single_node_hana.tfvars.json`, and I triggered terraform from `deploy/v2/terraform`.

This is how the Azure pipeline got triggered.

Can you test the same and let me know if you see the hdbx VMs? Thanks!

templates/tempGen/template.json
```diff
@@ -113,5 +113,60 @@
          ]
          }
       ]   
-   }
+   },
+
+    "databases":[
+        {
+            "name":"hdb1",
```
rukawata: The database's name is normally the SID (eg. HN1), which is defined in section "instance":

```
"instance": {
        "sid": "HN1",
        "instance_number": "00"
      },
```
Please check the latest template I sent out last Friday. Attach here the templates.
[20190913-templates.zip](https://github.com/Azure/sap-hana/files/3618462/20190913-templates.zip)

deploy/v2/terraform/main.tf
suhanselvan: The `templates/tempGen/template.json` was having the `platform` key set to blank. I had set a condition to create the nodes only for HANA databases. I've updated the template.json file.

templates/tempGen/template.json
suhanselvan: removed the name key and updated other keys as per the latest JSON template.

deploy/v2/template_samples/single_node_hana.tfvars.json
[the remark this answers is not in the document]
suhanselvan: Added.

deploy/v2/terraform/modules/hdb_node/main.tf
rukawata: Just FYI, now for each node, we have two attributes for private ip addresses:
`admin_nic_ip` and `db_nic_ip`

suhanselvan:
> LGTM. I assume the Terraform code to connect with Ansible will come in the next PR, correct? Thanks!

That is correct.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,143 @@
+##################################################################################################################
+# Database Node
```
tniek: "HANA DB node"

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  storage_image_reference {
+    publisher = each.value.os.publisher
+    offer     = each.value.os.offer
+    sku       = each.value.os.sku
+    version   = "latest"
+  }
+
+   dynamic "storage_data_disk" {
```
tniek: wrong indentation

deploy/v2/terraform/modules/hdb_node/variables.tf
```diff
@@ -0,0 +1,42 @@
...
+}
+
+variable "subnet-sap-admin" {
+  description = "Details of the SAP admin subnet"
+}
+
+variable "subnet-sap-db" {
+  description = "Details of the SAP DB subnet"
+}
+
```
tniek: nit: remove blank line

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,124 @@
...
-	    {
-	    "name" : "",
+            "authentication" : {
+                "type" : "",
+                "username" : "",
+                "use_existing" : "",
+                "path_to_public_key" : "",
+                "path_to_private_key" : ""
+            },
+            "components" : []
```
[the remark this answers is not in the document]
suhanselvan: looks fine for me. Moved the closing square bracket to next line.

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,124 @@
...
+            },
+
```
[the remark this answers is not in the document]
suhanselvan: removed.

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,124 @@
...
+                "username" : "",
+                "use_existing" : "",
+                "path_to_public_key" : "",
+                "path_to_private_key" : ""
+            },
+            "components" : []
+            }
+        ]
+    },
+
```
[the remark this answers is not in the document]
suhanselvan: removed.

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,124 @@
...
 	    },
-	    "components" : []
-       	    }
-        ]     
-    }
+	    "components" : [
+		{
+		    "name" : "",
+		    "credentials" : {
+		    "xsa_admin_password" : ""
```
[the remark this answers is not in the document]
suhanselvan: fixed.

deploy/v2/input.tfvars.json
```diff
@@ -61,69 +61,124 @@
...
+		    },
+		    "routing" : ""
+		},
+		{
+	            "name" : "",
+		    "credentials" : {
+		    "cockpit_admin_password" : ""
```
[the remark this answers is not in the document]
suhanselvan: fixed.

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -116,5 +116,57 @@
...
+                    "name" : "XSA",
+                    "credentials" : {
+                    "xsa_admin_password" : "Manager1"
+                    },
+                    "routing" : "port"
+                },
+                {
+                    "name" : "Cockpit",
+                    "credentials" : {
+                    "cockpit_admin_password" : "Manager1"
```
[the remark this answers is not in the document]
suhanselvan: fixed.

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -116,5 +116,57 @@
...
+                "db_systemdb_password" : "Manager1",
+                "db_tenant_password" : "Manager1",
+                "os_sidadm_password" : "Help4you",
+                "os_sapadm_password" : "Help4you"
+            },
+            "components" : [
+                {
+                    "name" : "XSA",
+                    "credentials" : {
+                    "xsa_admin_password" : "Manager1"
```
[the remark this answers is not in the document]
suhanselvan: fixed.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: changed.

deploy/v2/terraform/modules/hdb_node/main.tf
suhanselvan: fixed.

deploy/v2/terraform/modules/hdb_node/variables.tf
suhanselvan: removed.

PR #219, opened 2019-09-04
Map Log Analytics TimeGenerated field to UTC_TIMESTAMP column

tniek:
- This PR maps the TimeGenerated field in Log Analytics to the UTC_TIMESTAMP column of the SQL result set as described [here](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/data-collector-api).
- This was requested internally (MSIT), as KQL relies on the TimeGenerated field for predictive analytics/anomaly detection.

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -18,7 +18,7 @@ import re
 
 ###############################################################################
 
-PAYLOAD_VERSION              = "0.4.4"
+PAYLOAD_VERSION              = "0.4.5"
 PAYLOAD_DIRECTORY            = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -437,6 +437,7 @@ def buildSig(content, timestamp):
          "Authorization": buildSig(jsonData, timestamp),
          "Log-Type":      logType,
          "x-ms-date":     timestamp,
+         "time-generated-field": "UTC_TIMESTAMP",
```
persiaAziz: Do you need to change the format of the timestamp being generated? or it already complies with UTC?

tniek: Not needed, as the format of both fields matches.

PR #221, opened 2019-09-05
Remove logic to cover multiple KeyVault naming conventions

tniek:
- We've now closed on the naming convention for all required resources inside the managed resource group; specifically, the KeyVault is named `sapmon-kv-1234567890abcd` (where 1234567890abcd denotes the unique sapmonId).
  - Since there are no more SAP Monitor instances using the legacy naming convention (`sapmon1234567890abcd`), it is unnecessary to check for it.
- Therefore, the logic introduced in PR #197 to identify the correct KeyVault name (`function _Context.identifyKeyVault()`) is no longer needed and has been removed in this PR.

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -18,7 +18,7 @@ import re
 
 ###############################################################################
 
-PAYLOAD_VERSION              = "0.4.5"
+PAYLOAD_VERSION              = "0.4.6"
 PAYLOAD_DIRECTORY            = os.path.dirname(os.path.realpath(__file__))
 STATE_FILE                   = "%s/sapmon.state" % PAYLOAD_DIRECTORY
 INITIAL_LOADHISTORY_TIMESPAN = -(60 * 1)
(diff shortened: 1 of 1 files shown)


monitor/sapmon.py
```diff
@@ -18,7 +18,7 @@
 
 ###############################################################################
 
-PAYLOAD_VERSION              = "0.4.4"
+PAYLOAD_VERSION              = "0.4.6"
```
pabowers: are we skipping 0.4.5 ?

monitor/sapmon.py
```diff
@@ -437,6 +440,7 @@ def buildSig(content, timestamp):
          "Authorization": buildSig(jsonData, timestamp),
          "Log-Type":      logType,
          "x-ms-date":     timestamp,
+         "time-generated-field": "UTC_TIMESTAMP",
```
pabowers: I thought this change was in a different PR

PR #226, opened 2019-09-17
  Add storage log handler to the collector script

persiaAziz:


monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
+INITIAL_LOADHISTORY_TIMESPAN      = -(60 * 1)
+TIME_FORMAT_HANA                  = "%Y-%m-%d %H:%M:%S.%f"
+TIME_FORMAT_LOG_ANALYTICS         = "%a, %d %b %Y %H:%M:%S GMT"
+TIMEOUT_HANA                      = 5
+DEFAULT_CONSOLE_LOG_LEVEL         = logging.INFO
+DEFAULT_FILE_LOG_LEVEL            = logging.INFO
+DEFAULT_QUEUE_LOG_LEVEL           = logging.INFO
+LOG_FILENAME                      = "%s/sapmon.log" % PAYLOAD_DIRECTORY
+KEYVAULT_NAMING_CONVENTION        = "sapmon-kv-%s"
+STORAGE_ACCOUNT_NAMING_CONVENTION = "sapmonsto%s"
```
PakDLiu: different convention for storage accounts?

monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
+        "file": {
+            "class": "logging.handlers.RotatingFileHandler",
+            "formatter": "detailed",
+            "level": DEFAULT_FILE_LOG_LEVEL,
+            "filename": LOG_FILENAME,
+            "maxBytes": 10000000,
+            "backupCount": 10,
+        },
+        'queue': {
+            'account_name': 'mystorageaccountname',
```
PakDLiu: looks like these are overwritten later, should we put these here? might cause confusion?

monitor/sapmon.py
```diff
@@ -477,6 +552,8 @@ def __init__(self, operation):
       self.lastPull = None
       self.lastResultHashes = {}
       self.readStateFile()
+      self.storageQueue = QueueStorage(sapmonId=self.sapmonId, msiClientID=vmTags.get("SapMonMsiClientId", None),subscriptionId=self.vmInstance["subscriptionId"],resourceGroup=self.vmInstance["resourceGroupName"])
```
PakDLiu: is this test code? we are initializing storageQueue here and initializeLogger?

monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
-STATE_FILE                   = "%s/sapmon.state" % PAYLOAD_DIRECTORY
-INITIAL_LOADHISTORY_TIMESPAN = -(60 * 1)
-TIME_FORMAT_HANA             = "%Y-%m-%d %H:%M:%S.%f"
-TIME_FORMAT_LOG_ANALYTICS    = "%a, %d %b %Y %H:%M:%S GMT"
-TIMEOUT_HANA                 = 5
-DEFAULT_CONSOLE_LOG_LEVEL    = logging.INFO
-DEFAULT_FILE_LOG_LEVEL       = logging.INFO
-LOG_FILENAME                 = "%s/sapmon.log" % PAYLOAD_DIRECTORY
-KEYVAULT_NAMING_CONVENTION   = "sapmon-kv-%s"
+PAYLOAD_VERSION                   = "0.4.6"
```
tniek: Please increment the version to 0.5.0 (major feature)

monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
-         "filename": LOG_FILENAME,
-         "maxBytes": 10000000,
-         "backupCount": 10,
-      },
-   },
-   "root": {
-      "level": logging.DEBUG,
-      "handlers": ["console", "file"],
-   }
+    "version": 1,
```
tniek: Please increment this to version 2 of the logging config.

monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
+        },
+        "file": {
+            "class": "logging.handlers.RotatingFileHandler",
+            "formatter": "detailed",
+            "level": DEFAULT_FILE_LOG_LEVEL,
+            "filename": LOG_FILENAME,
+            "maxBytes": 10000000,
+            "backupCount": 10,
+        },
+        'queue': {
```
tniek: Double-quotes for strings (to be consistent)?

monitor/sapmon.py
tniek: Could be empty, will be overwritten

monitor/sapmon.py
```diff
@@ -288,6 +319,21 @@ def getComputeInstance(operation):
       except Exception as e:
          logging.error("could not obtain instance metadata (%s)" % e)
       return computeInstance
+   
+   @staticmethod
+   def getAuthTokenNoLogging(resource, msiClientId = None):
+      """
+      Get an authentication token via IMDS
```
tniek: What exactly does the "NoLogging" in the function name refer to?

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
 
 ###############################################################################
 
+class QueueStorage():
```
tniek: Probably AzureStorageQueue (to be consistent with the naming)

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+    storageAccountName = None
+    queueName = None
+    token = {}
+    subscriptionId = None
+    resourceGroup = None
+    def __init__(self, sapmonId,msiClientID,subscriptionId, resourceGroup):
+        """
+        Get the Queue Name
```
tniek: "Get the name of the storage queue"

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+        self.token["access_token"] = tokenResponse
+        self.subscriptionId = subscriptionId
+        self.resourceGroup = resourceGroup
+
+    def getAccessKey(self):
+        """
+        Get access key to the storage queue
+        """
+        storageclient = StorageManagementClient(credentials=BasicTokenAuthentication(self.token),subscription_id=self.subscriptionId)
+        storageKeys = storageclient.storage_accounts.list_keys(resource_group_name=self.resourceGroup,account_name=self.storageAccountName)
```
tniek: nit: spacing between parameters

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+        if storageKeys == None or len(storageKeys.keys) <= 0 :
```
tniek: Should be `if storageKeys is None` (or `if not storageKeys`)

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+           print("Could not retrive storage keys of the storage account{0}".format(self.storageAccountName))
```
tniek: Why `print`? Wouldn't we be able to log to console/file at least?

monitor/sapmon.py
tniek: Since logging-specific, maybe refactor into a separate logging class?

monitor/sapmon.py
```diff
@@ -735,7 +812,23 @@ def monitor(args):
 
    logger.info("monitor payload successfully completed")
    return
-      
+
+def initializeLogger(args):
```
tniek: nit: "initLogger" (to be consistent)

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
 
 ###############################################################################
 
+class QueueStorage():
+    storageAccountName = None
```
tniek: I would rename into `accountName`

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+    queueName = None
```
tniek: I would rename into just `AzureStorageQueue.name`

monitor/sapmon.py
tniek: I think this was discussed in her PR to deploy the storage account; no hyphens and limit of 24 characters resulted in "sapmonsto". If we're really unhappy, we could rename into "sapmonque"

persiaAziz: yes.. storage account name can't have hyphens. But the storage queue name follows the convention.

monitor/sapmon.py
persiaAziz: I will put a dummy value

monitor/sapmon.py
persiaAziz: This function can't have any logging (as this is being used to create the logger). Since function overloading is not supported in this scenario, I can't create a no logging version of the same function getAuthToken. Hence the name

monitor/sapmon.py
persiaAziz: I can't use the logger here as it has not been initialized yet

monitor/sapmon.py
persiaAziz: sorry yes.. this is test code. I meant to get rid of this

persiaAziz: It will be just one function to initialize the global logger with dependency on the other classes . If you feel strongly about having a separate class then I can make it.

monitor/sapmon.py
persiaAziz: I got an error by changing it to 2. Looks like this version is internal to the logging package.

monitor/sapmon.py
```diff
@@ -15,53 +15,68 @@
...
+        "file": {
+            "class": "logging.handlers.RotatingFileHandler",
+            "formatter": "detailed",
+            "level": DEFAULT_FILE_LOG_LEVEL,
+            "filename": LOG_FILENAME,
+            "maxBytes": 10000000,
+            "backupCount": 10,
+        },
+        "queue": {
+            "account_name": "---",
```
PakDLiu: i guess what i was saying is can we remove this field outright?

monitor/sapmon.py
persiaAziz: Done

monitor/sapmon.py
persiaAziz: removed

monitor/sapmon.py
persiaAziz: done

monitor/sapmon.py
persiaAziz: done

monitor/sapmon.py
persiaAziz: replaced with ---

monitor/sapmon.py
persiaAziz: I would keep it for clarity of what the dictionary looks like

monitor/sapmon.py
tniek: Not really a fan; is there a fan to initialize the logger with file/console and (programmatically) add queue later on once we have the token? Maybe a TODO item.

monitor/sapmon.py
```diff
@@ -272,6 +287,22 @@ def _sendRequest(endpoint, params = {}, headers = {}):
...
+   @staticmethod
+   def getMetadata():
+      """
+      Get the compute instance for the current VM via IMS
+      """     
+      computeInstance = None
+      try:
+         computeInstance = AzureInstanceMetadataService._sendRequest(
+            "instance",
+            headers = {"User-Agent": "SAP Monitor/%s (%s)" % (PAYLOAD_VERSION, "--")}
```
tniek: I really don't like that we removed `operation` from the IMDS headers here, as this removes our ability to distinguish onboarding from monitoring payloads. I.e., we won't know be able to tell via KQL anymore how many customers actually deployed.

Any options to encode the operation back into the HTTP header?

monitor/sapmon.py
```diff
@@ -458,6 +504,35 @@ def buildSig(content, timestamp):
...
+        """
+        Get the access key to the storage queue
+        """
+        storageclient = StorageManagementClient(credentials=BasicTokenAuthentication(self.token), subscription_id=self.subscriptionId)
+        storageKeys = storageclient.storage_accounts.list_keys(resource_group_name=self.resourceGroup, account_name=self.accountName)
+        if storageKeys is None or len(storageKeys.keys) <= 0 :
+           print("Could not retrive storage keys of the storage account{0}".format(self.accountName))
+           return None
+        return storageKeys.keys[0].value
+################################################################################
```
tniek: nit: add blank line before line 534

PR #227, opened 2019-09-18
Add SapMonitor query framework

tniek:
- This PR refactors the monitoring payload, so that content (i.e. checks, which in the case of SAP HANA, consist of SQL queries) is separated from the execution logic.
  - This allows for much easier extension of the monitoring content, without bloating the payload code or introducing redundancy.
- For this new framework to work, we need to adjust the deployment logic, such that *.json (content) files are pulled together with the `sapmon.py` payload script (separate PR in VSO).

### DIFF
--- a/monitor/SapHana.json
+++ b/monitor/SapHana.json
@@ -1,0 +1,26 @@
+{
+        "contentType": "HANA",
+        "contentVersion": "0.1.0",
+        "checks": [
+                {
+                        "name": "HostConfig",
+                        "description": "SAP HANA Host Config",
+                        "customLog": "SapHana_HostConfig",
+                        "frequencySecs": 60,
+                        "hanaOptions": {
+                                "query": "SELECT * FROM SYS.M_LANDSCAPE_HOST_CONFIGURATION"
+                        }
+                },
+                {
+                        "name": "LoadHistory",
+                        "description": "SAP HANA Load History",
+                        "customLog": "SapHana_LoadHistory",
+                        "frequencySecs": 60,
+                        "hanaOptions": {
+                                "isTimeSeries": true,
+                                "initialTimespanSecs": 30,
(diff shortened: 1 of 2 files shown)


monitor/SapHana.json
```diff
@@ -0,0 +1,26 @@
+{
+        "contentType": "HANA",
```
persiaAziz: DO you mean database type? Content type sounds like something that would contain say "sql" or "json"

monitor/sapmon.py
```diff
@@ -81,13 +84,151 @@
 
 ###############################################################################
 
+sapmonContentTypes = {
+   "HANA": "SapHanaCheck"
+}
+
+class SapmonCheck:
```
persiaAziz: I would make this class an abstract or interface so all the future query classes can implement this. This class should contain member functions like prepareSql, calculateResultHash etc as abstract methods, so the derived classes have to implement this method.

monitor/sapmon.py
```diff
@@ -498,6 +546,7 @@ class _Context(object):
    Internal context handler
    """
    hanaInstances = []
+   availChecks   = []
```
persiaAziz: availchecks? you mean available checks?

monitor/sapmon.py
```diff
@@ -81,13 +84,151 @@
...
+      self.name          = name
+      self.description   = description
+      self.customLog     = customLog
+      self.frequencySecs = frequencySecs
+      self.state         = {
+         "isEnabled":    enabled,
+         "lastRunLocal": None,
+      }
+
+class SapHanaCheck(SapmonCheck):
```
persiaAziz: SInce we already need multiple script files anyway, I think this class should go in a separate file, just to keep the files cleaner

monitor/sapmon.py
```diff
@@ -81,13 +84,151 @@
...
+         logger.info("SQL result is empty")
+         return False
+      self.state["lastRunServer"] = self.lastResult[0][self.colIndex[self.COL_SERVER_UTC]]
+      self.state["lastResultHash"] = self.calculateResultHash()
+      logger.info("internal state successfully updated")
+      return True
+
+###############################################################################
+
 class SapHana:
```
persiaAziz: Sorry If I am missing something. Where are we using this SapHana class now?

monitor/SapHana.json
tniek: It's really monitoring content, so could be "HANA", "Pacemaker". "NetApp" etc. I used to call it `checksType`, but that seemed to misleading.

monitor/sapmon.py
tniek: Actually, none of the ones you mentioned (`prepareSql()`, `calculateResultHash()`) would be implemented by other check types (except non-HANA databases, maybe). Really the only methods that would be overloaded are `run()` and `updateState()`.
That being said, I agree that making `SapmonCheck` an abstract class is a good idea!

monitor/sapmon.py
tniek: Yes, to make it shorter. Prefer `availableChecks`?

monitor/sapmon.py
tniek: +1 on that, the whole payload script needs to be refactored. I'd address that in a separate PR, though.

monitor/sapmon.py
tniek: `SapHana` represents the HANA DB instance as such, whereas `SapHanaCheck` is a just a check. The reason for separating them is that there may be other operations we need to do on a HANA instance that are not checks.

Going forward, as we will support multiple source systems, there could be a common base class (something like `SourceSystem`) that all source system types inherit from.

monitor/sapmon.py
persiaAziz: yes. availableChecks sounds good :)

monitor/sapmon.py
```diff
@@ -81,13 +84,151 @@
...
+   COL_CONTENT_VERSION = "CONTENT_VERSION"
+   COL_SAPMON_VERSION  = "SAPMON_VERSION"
+   TIME_FORMAT_HANA    = "%Y-%m-%d %H:%M:%S.%f"
+
+   prefix             = "HANA"
+   isTimeSeries       = False
+   def __init__(self, hanaOptions, **kwargs):
+      super().__init__(**kwargs)
+      self.query                  = hanaOptions["query"]
+      self.isTimeSeries           = hanaOptions.get("isTimeSeries", False)
```
rukawata: isTimeSeries is False by default, do we need to do that twice? `isTimeSeries       = False`

tniek: Yes, it is. See `sapmon.py`, class `SapHanaCheck` (last line before the constructor)

monitor/SapHana.json
```diff
@@ -0,0 +1,26 @@
...
+                        "hanaOptions": {
+                                "query": "SELECT * FROM SYS.M_LANDSCAPE_HOST_CONFIGURATION"
+                        }
+                },
+                {
+                        "name": "LoadHistory",
+                        "description": "SAP HANA Load History",
+                        "customLog": "SapHana_LoadHistory",
+                        "frequencySecs": 60,
+                        "hanaOptions": {
```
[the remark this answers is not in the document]
tniek: Per @PakDLiu's suggestion, we want to highlight that these properties are specific to HANA. If we use a more generic name, this wouldn't be the case anymore.

monitor/sapmon.py
```diff
@@ -81,13 +85,159 @@
...
+            logger.info("time series query for check %s_%s has never been run, applying initalTimespanSecs=%d" % \
+               (self.prefix, self.name, self.initialTimespanSecs))
+            sqlUntilNow = " WHERE h.TIME > ADD_SECONDS(NOW(), %d) AND" % (self.initialTimespanSecs * (-1))
+         else:
+            logger.info("time series query for check %s_%s has been run at %s, filter out only new records since then" % \
+               (self.prefix, self.name, lastRunServer.strftime(self.TIME_FORMAT_HANA)))
+            try:
+               sqlUntilNow = " WHERE ADD_SECONDS(h.TIME, i.VALUE*(-1)) > '%s' AND" % lastRunServer.strftime(self.TIME_FORMAT_HANA)
+            except Exception as e:
+               logger.error("could not format lastRunServer=%s (%s)" % (str(lastRunServer), e))
```
PakDLiu: if the code gets to this line, `sqlUntilNow` would be null right? Then the rest of the code runs? Should we return after this `logger.error` statement?

monitor/sapmon.py
```diff
@@ -81,13 +85,159 @@
...
+      logger.info("converting result set into JSON")
+      logData = []
+      for r in self.lastResult:
+         logItem = {}
+         for c in self.colIndex.keys():
+            if c != self.colTimeGenerated and (c.startswith("_") or c == "DUMMY"): # remove internal fields
+               continue
+            logItem[c] = r[self.colIndex[c]]
+         try:
+            jsonData = json.dumps(logItem, sort_keys=True, indent=4, cls=_JsonEncoder)
```
PakDLiu: Where is `jsonData `used?

monitor/sapmon.py
tniek: Yep, good catch. Will fix.

monitor/sapmon.py
tniek: Ideally in line 215 - will fix.

PR #229, opened 2019-09-20
Add file to collect stats from sar and iostat

jeffaco:


tools/sar-iostat.sh
```diff
@@ -0,0 +1,101 @@
...
+  # Delete the collection processes
+
+  killall -w iostat mpstat sar
+}
+
+function ctrl_c() {
+  # Delete the collection processes and exit
+
+  stop_jobs
+  echo "Exiting due to interrupt ..."
```
hrushib: Shouldn't this be logged before calling stop_jobs?

jeffaco: I'm not 100% sure what you mean here. The background jobs constantly collect information, and are written to the logs constantly. Stopping the jobs just stops that from happening, no logging should be necessary. What kind of "logging" did you have in mind here?

tools/sar-iostat.sh
```diff
@@ -0,0 +1,117 @@
...
+
+  iostat -d -k -x -y 1 > "${IOSTAT_OUTPUT}" &
+  # mpstat collects averages, so pull that less often
+  mpstat -I ALL 5 > "${MPSTAT_OUTPUT}" &
+  sar -A -o "${SAR_OUTPUT}" 1 > /dev/null &
+
+  /usr/sbin/tcpdump -tttt -i eth1 ether proto 0x8808 -w "${TCPDUMP_OUTPUT_ETH1}" 2>/dev/null &
+  /usr/sbin/tcpdump -tttt -i eth2 ether proto 0x8808 -w "${TCPDUMP_OUTPUT_ETH2}" 2>/dev/null &
+  /usr/sbin/tcpdump -tttt -i eth3 ether proto 0x8808 -w "${TCPDUMP_OUTPUT_ETH3}" 2>/dev/null &
+  /usr/sbin/tcpdump -tttt -i eth4 ether proto 0x8808 -w "${TCPDUMP_OUTPUT_ETH4}" 2>/dev/null &
```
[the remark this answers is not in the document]
jeffaco: Predictable network names are a ways off, and this script is temporary in nature (at least for now). We may not even need this any longer if/when we move active/backup. For now, I won't consider this change. May make this change, though, if we still need this script down the road and we have vendor generated VLI images.

PR #230, opened 2019-09-24
lastRunServer is not of datetime type

persiaAziz:
The wrong type was causing the script to crash and thus log injection was not happening regularly


lagalbra:
Can you add more details in the PR description? From the title, I'm not sure if this is a bug, what is impacted, or what the new functionality is with this change.

PR #231, opened 2019-09-24
Remove logging of passwords from onboarding payload

tniek:


lagalbra:
Did you also check where the password could be logged as a single field inside of a larger object that's getting logged?

lagalbra:
> Did you also check where the password could be logged as a single field inside of a larger object that's getting logged?

For example, line 716, 731 log  `hanaDetails`

tniek:
> Make sure no objects containing the password are logged as a whole

Good point - thank you!

PR #233, opened 2019-09-25
replace sys.exc_info() with traceback

persiaAziz:
* sys.exc_info does not take any parameter- which is the first problem
* traceback format will give a nicer formatting

### DIFF
--- a/monitor/sapmon.py
+++ b/monitor/sapmon.py
@@ -25,6 +25,7 @@ import pyhdb
 import re
 import requests
 import sys
+import traceback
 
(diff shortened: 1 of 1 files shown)


persiaAziz:
> PAYLOAD_VERSION should be increased (0.6.3)
> Do we want to show a traceback for all errors we're logging?

Payload version updated. We should use traceback for all errors. This PR is sort of a warmfix as it breaks the deployment if there is a storage queue error. I will make a new PR to put traceback in all the errors

PR #234, opened 2019-09-26
Wft terraform handoff

suhanselvan:
This PR pertains to the v2 implementation.
-> Added resource block in `common_infrastructure` module to create storage account for SAP bits.
-> Moved boot diagnostics storage account resource block from `jumpbox` module to `common_infrastructre` module as it is common for jumpboxes as well as database VMs.
-> Added module `output_json` that generates a new `output.json` file. This module merges the private IP address of the VMs created to the input JSON file provided by the user.
-> Added Provisioner blocks to copy ssh keys and `output.json` from local system to RTI box.


deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -153,9 +133,48 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+      user        = var.jumpboxes.linux[count.index].authentication.username
+      private_key = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", false) == false ? null : file(var.jumpboxes.linux[count.index].authentication.path_to_private_key)
+      password    = lookup(var.jumpboxes.linux[count.index].authentication, "password", null)
+    }
+    source      = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", null)
+    destination = "/home/${var.jumpboxes.linux[count.index].authentication.username}/.ssh/id_rsa"
+    on_failure  = "continue"
+  }
+
+  provisioner "file" {
```
rukawata: I assume the above provisioned suppose to put ssh key pair and output json into RTI, however I do not see them on the RTI:

```
root@rti:/home/azureadm/.ssh# ls -ltr
total 4
-rw------- 1 azureadm azureadm 725 Sep 30 17:05 authorized_keys
```

Could you please check that?

deploy/v2/terraform/modules/output_json/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "output-json" {
+  value = local_file.output-json
```
rukawata: Can you change the name of the output file as a variable (taken from read_tf_output_from_file in 20190930-templates\input.json) ?

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -234,3 +234,32 @@ resource "azurerm_virtual_network_peering" "peering-sap-management" {
...
+resource "random_id" "random-id" {
+  keepers = {
+    # Generate a new id only when a new resource group is defined
+    resource_group = var.infrastructure.resource_group.is_existing ? data.azurerm_resource_group.resource-group[0].name : azurerm_resource_group.resource-group[0].name
+  }
+  byte_length = 4
+}
+
+# Creates storage account for storing SAP Bits
+resource "azurerm_storage_account" "storageaccount-sapbits" {
```
rukawata: The storage for SAP Bits would be need also a container.
I added another section in the input JSON to provide the type of containers, please adjust the code for storage account and add code for creating a container. Thanks!
[20190930-templates.zip](https://github.com/Azure/sap-hana/files/3672653/20190930-templates.zip)

deploy/v2/terraform/modules/common_infrastructure/outputs.tf
```diff
@@ -25,3 +25,11 @@ output "subnet-sap-db" {
...
   value = var.infrastructure.vnets.sap.subnet_db.nsg.is_existing ? data.azurerm_network_security_group.nsg-db : azurerm_network_security_group.nsg-db
 }
+
+output "storageaccount-bootdiagnostics" {
+  value = azurerm_storage_account.storageaccount-bootdiagnostics
+}
+
+output "storageaccount-sapbits" {
+  value = azurerm_storage_account.storageaccount-sapbits
+}
```
rukawata: We would need 3 things from the storage account for SAP bits to be able to mount the container to Jumpboxes as well as HDB nodes:

- Storage account name
- Storage access key
- Container name

You an find details why those are needed in the link below:

- Instruction to mount [Blobfuse](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-how-to-mount-container-linux)
- Instruction to mount [AzureFiles](https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-use-files-linux)

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: It worked fine when I tested. May I know which input file you used for testing?

deploy/v2/terraform/modules/output_json/outputs.tf
suhanselvan: Will do.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: Sure, I sent the input JSON to your email. Thanks!

suhanselvan: Its not working because the private key is in .ppk format and Terraform doesn't support it. You'll have to convert it into .pem

deploy/v2/terraform/modules/output_json/outputs.tf
suhanselvan: Wondering why we would need to make the output file name a variable as it is only going to be used by Ansible and will be destroyed along with the RTI after installation.

deploy/v2/terraform/modules/common_infrastructure/main.tf
suhanselvan: Apart from the name of the storage account and container can we make the rest of the parameters static? Because letting the user change parameters like `"account_kind": "BlockBlobStorage"` would actually cause issues.

rukawata: As discussed, we would like to keep those configurable. Thanks!

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: Thanks. Now it works.

deploy/v2/terraform/modules/output_json/main.tf
```diff
@@ -0,0 +1,9 @@
...
...windows-jumpbox-ips[index(var.jumpboxes.windows, windows-jumpbox)] }], "linux" = [for linux-jumpbox in var.jumpboxes.linux : { name = linux-jumpbox.name, destroy_after_deploy = linux-jumpbox.destroy_after_deploy, size = linux-jumpbox.size, disk_type = linux-jumpbox.disk_type, os = linux-jumpbox.os, authentication = linux-jumpbox.authentication, components = linux-jumpbox.components, private_ip_address = local.linux-jumpbox-ips[index(var.jumpboxes.linux, linux-jumpbox)] }] }, "databases" = [for database in var.databases : { platform = database.platform, version = database.version, os = database.os, size = database.size, filesystem = database.filesystem, high_availability = database.high_availability, instance = database.instance, authentication = database.authentication, credentials = database.credentials, components = database.components, nodes = [for dbnode-admin-ip in local.dbnode-admin-ips : { admin_nic_ip = dbnode-admin-ip, db_nic_ip = local.dbnode-db-ips[index(local.dbnode-admin-ips, dbnode-admin-ip)], role = local.dbnodes[index(local.dbnode-admin-ips, dbnode-admin-ip)].role } if local.dbnodes[index(local.dbnode-admin-ips, dbnode-admin-ip)].platform == database.platform] }] })
```
rukawata: Can you write the string for the content in multiple line for better readability?

deploy/v2/terraform/modules/common_infrastructure/outputs.tf
suhanselvan: Added.

deploy/v2/terraform/modules/output_json/main.tf
suhanselvan: Wish I could. But HCL does not support multi-line expressions.

deploy/v2/terraform/main.tf
```diff
@@ -30,5 +33,19 @@ module "hdb" {
...
+# Generate output JSON file
+module "output_json" {
+  source                 = "./modules/output_json"
+  infrastructure         = var.infrastructure
+  jumpboxes              = var.jumpboxes
+  databases              = var.databases
+  software               = var.software
+  storageaccount-sapbits = module.common_infrastructure.storageaccount-sapbits
+  nic-windows            = module.jumpbox.nic-windows
+  nic-linux              = module.jumpbox.nic-linux
```
tniek: I'm a bit confused why we're returning the individual NICs - wouldn't they be properties of `var.jumpboxes`?
Also, what happens if there are several Windows or Linux jumpboxes - do they all share the same NIC?

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -234,3 +234,49 @@ resource "azurerm_virtual_network_peering" "peering-sap-management" {
...
+
+# STORAGE ACCOUNTS ===============================================================================================
+
+# Generates random text for boot diagnostics storage account name
+resource "random_id" "random-id" {
+  keepers = {
+    # Generate a new id only when a new resource group is defined
+    resource_group = var.infrastructure.resource_group.is_existing ? data.azurerm_resource_group.resource-group[0].name : azurerm_resource_group.resource-group[0].name
+  }
+  byte_length = 4
```
tniek: Is length 4 sufficient here?
@rukawata Is a random storage account name even the right approach?

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -234,3 +234,49 @@ resource "azurerm_virtual_network_peering" "peering-sap-management" {
...
+}
+
+# Creates storage account for storing SAP Bits
+resource "azurerm_storage_account" "storageaccount-sapbits" {
+  count                    = var.software.storage_account_sapbits.is_existing ? 0 : 1
+  name                     = lookup(var.software.storage_account_sapbits, "name", false) ? var.software.storage_account_sapbits.name : "sapbits${random_id.random-id.hex}"
+  resource_group_name      = var.infrastructure.resource_group.is_existing ? data.azurerm_resource_group.resource-group[0].name : azurerm_resource_group.resource-group[0].name
+  location                 = var.infrastructure.region
```
tniek: Region limitation for premium blobs?

deploy/v2/terraform/modules/common_infrastructure/main.tf
```diff
@@ -234,3 +234,49 @@ resource "azurerm_virtual_network_peering" "peering-sap-management" {
...
+
+# Imports existing storage account to use for SAP bits
+data "azurerm_storage_account" "storageaccount-sapbits" {
+  count               = var.software.storage_account_sapbits.is_existing ? 1 : 0
+  name                = split("/", var.software.storage_account_sapbits.arm_id)[8]
+  resource_group_name = split("/", var.software.storage_account_sapbits.arm_id)[4]
+}
+
+# Creates boot diagnostics storage account
+resource "azurerm_storage_account" "storageaccount-bootdiagnostics" {
```
tniek: See above, suggest `storage-bootdiag`

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -153,9 +133,48 @@ resource "azurerm_virtual_machine" "vm-linux" {
     }
   }
 
+  provisioner "file" {
+    connection {
+      type        = "ssh"
+      host        = var.jumpboxes.linux[count.index].name == "rti" ? azurerm_public_ip.public-ip-linux[count.index].ip_address : null
+      user        = var.jumpboxes.linux[count.index].authentication.username
+      private_key = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", false) == false ? null : file(var.jumpboxes.linux[count.index].authentication.path_to_private_key)
+      password    = lookup(var.jumpboxes.linux[count.index].authentication, "password", null)
```
tniek: Does this actually work (providing both, `private_key` and `password` even if one is null)?

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -153,9 +133,48 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+      host        = var.jumpboxes.linux[count.index].name == "rti" ? azurerm_public_ip.public-ip-linux[count.index].ip_address : null
+      user        = var.jumpboxes.linux[count.index].authentication.username
+      private_key = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", false) == false ? null : file(var.jumpboxes.linux[count.index].authentication.path_to_private_key)
+      password    = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", null)
+    }
+    source      = var.output-json.filename
+    destination = "/home/${var.jumpboxes.linux[count.index].authentication.username}/output.json"
+    on_failure  = "continue"
+  }
+
```
tniek: A lot of redundant code here. Is there a way to combine these and not having to repeat the `connection` block every single time?

deploy/v2/terraform/modules/output_json/main.tf
```diff
@@ -0,0 +1,9 @@
...
+  filename = "${path.module}/../../../output.json"
```
tniek: No relative path magic please. What if the directory structure changes?

deploy/v2/terraform/modules/output_json/main.tf
```diff
@@ -0,0 +1,9 @@
+##################################################################################################################
+# OUTPUT JSON
+##################################################################################################################
+
+# Generates the output JSON with IP address details
```
tniek: Not only IP address details though, right?

deploy/v2/terraform/modules/output_json/variables.tf
```diff
@@ -0,0 +1,43 @@
...
+
+variable "storageaccount-sapbits" {
+  description = "Details of the storage account for SAP bits"
+}
+
+locals {
+  windows-jumpbox-ips = var.nic-windows[*].private_ip_address
+  linux-jumpbox-ips   = var.nic-linux[*].private_ip_address
+  dbnode-admin-ips    = [for k, v in var.nic-dbnode-admin : v.private_ip_address]
+  dbnode-db-ips       = [for k, v in var.nic-dbnode-db : v.private_ip_address]
```
tniek: For readability, I would prefix `ips-` (instead of putting it at the end).

deploy/v2/terraform/modules/output_json/variables.tf
```diff
@@ -0,0 +1,43 @@
...
+  dbnodes             = flatten([for database in var.databases : [for node in database.nodes : { role = node.role, platform = database.platform, name = node.name }]])
```
tniek: If we already have a list of dbnodes, wouldn't the IP addresses be properties of them? If yes, why are we creating `dbnode-admin-ips` and `dbnode-db-ips` redundantly?

deploy/v2/terraform/modules/output_json/main.tf
rukawata: I tried something like below and worked for me:

```
resource "local_file" "output-json" {
  content  = jsonencode (
  {"infrastructure": var.infrastructure,
  "databases": var.databases,
  "software": var.software,
  "private_ip": azurerm_network_interface.pip-nic-linux.private_ip_address,
  "fqdn": azurerm_public_ip.pip-linux.fqdn,
  "sap_hostname": var.databases[0].domain_name,
  "storage_account_id": azurerm_storage_account.sa.id,
  "storage_key": azurerm_storage_account.sa.primary_access_key,
  "storage_container": azurerm_storage_container.sc.id
  })
  filename = "./output.json"
}
```

deploy/v2/template_samples/single_node_hana.tfvars.json
```diff
@@ -117,18 +117,18 @@
             "platform" : "HANA",
             "version" : "2.00.043",
             "os" : {
-                "publisher" : "SUSE",
-                "offer" : "SLES-SAP",
-                "sku" : "12-sp2"
+                "publisher" : "Redhat",
+                "offer" : "RHEL-SAP-HANA",
+                "sku" : "7.3"
```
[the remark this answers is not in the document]
rukawata: Good catch, should be SLES.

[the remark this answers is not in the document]
rukawata: @tniek, as discussed with @suhanselvan, this is temporary set to RHEL since ansible part starts with RHEL, then SLES. After ansible part for RHEL is done, it will be changed back to SLES.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: It seems with file provisioner, you can only define one source and one destination.
@suhanselvan, maybe it's possible to define a variable which is a list of source and the same for destination, and iterate through the list? Can you explore that option? Thanks!

rukawata: BTW, @tniek is correct, we can define one connection (to RTI) in the resource block that can be used for all provisioner, then you don't have to define it multiple times. [see doc](https://www.terraform.io/docs/provisioners/connection.html)

> A connection block nested directly within a resource affects all of that resource's provisioners.

@suhanselvan, please adjust this part. Thanks!

deploy/v2/terraform/modules/common_infrastructure/main.tf
suhanselvan: Kept as is.

deploy/v2/terraform/modules/output_json/main.tf
suhanselvan: Added.

deploy/v2/terraform/main.tf
```diff
@@ -30,5 +33,19 @@ module "hdb" {
   nsg-admin                      = module.common_infrastructure.nsg-admin
   subnet-sap-db                  = module.common_infrastructure.subnet-sap-db
   nsg-db                         = module.common_infrastructure.nsg-db
-  storageaccount-bootdiagnostics = module.jumpbox.storageaccount-bootdiagnostics
+  storageaccount-bootdiagnostics = module.common_infrastructure.storageaccount-bootdiagnostics
```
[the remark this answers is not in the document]
suhanselvan: changed.

deploy/v2/terraform/main.tf
```diff
@@ -30,5 +33,19 @@ module "hdb" {
...
+}
+
+# Generate output JSON file
+module "output_json" {
+  source                 = "./modules/output_json"
+  infrastructure         = var.infrastructure
+  jumpboxes              = var.jumpboxes
+  databases              = var.databases
+  software               = var.software
+  storageaccount-sapbits = module.common_infrastructure.storageaccount-sapbits
```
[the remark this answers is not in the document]
suhanselvan: changed.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: changed.

deploy/v2/terraform/modules/output_json/variables.tf
suhanselvan: changed.

deploy/v2/terraform/modules/output_json/variables.tf
suhanselvan: Its because if the user does not specify an IP address for the nodes in the input file, Terraform assigns one. This is why I'm finding the IPs from the output of the hdb_node module.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: Yes. Setting an argument to null is equivalent to omitting it completely.

deploy/v2/terraform/main.tf
suhanselvan: Since Terraform assigns the IPs to jumpboxes in some cases. I'm getting the IPs from the output of jumpbox module. It returns the NIC info of all the Windows and Linux jumpboxes. I've made the variable names plural for better clarity.

deploy/v2/terraform/modules/output_json/main.tf
rukawata: @suhanselvan , please use the path from read_tf_output_from_file as the dest for output json.

suhanselvan: Changed. I'm also using a relative file path for locating the sizing JSON. Would you like that to be an user input as well?

rukawata: Thx! For the sizing JSON, we can make it a user input but gives it default value like we currently have. Only use the input value when it's specified.

deploy/v2/template_samples/single_node_hana.json
```diff
@@ -115,20 +115,20 @@
     "databases" : [
         {
             "platform" : "HANA",
-            "version" : "2.00.043",
+            "db_version" : "2.00.043",
             "os" : {
-                "publisher" : "SUSE",
-                "offer" : "SLES-SAP",
-                "sku" : "12-sp2"
+                "publisher" : "Redhat",
```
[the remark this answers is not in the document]
suhanselvan: Can you please look at the following discussion? https://github.com/Azure/sap-hana/pull/234#discussion_r331639432

deploy/v2/template_samples/single_node_hana.json
```diff
@@ -155,12 +155,22 @@
                     }
                 }
             ],
-            "nodes" : [
+            "dbnodes" : [
                 {
 	            "name" : "hdb1",
```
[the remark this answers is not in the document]
suhanselvan: The code does support multiple standalone nodes if the user needs. Its a similar naming convention being followed by the input file for the Azure Pipelines test.

deploy/v2/terraform/main.tf
```diff
@@ -1,34 +1,51 @@
...
 
 # Create Jumpboxes and RTI box
 module "jumpbox" {
-  source         = "./modules/jumpbox"
-  infrastructure = var.infrastructure
-  jumpboxes      = var.jumpboxes
-  resource-group = module.common_infrastructure.resource-group
-  subnet-mgmt    = module.common_infrastructure.subnet-mgmt
-  nsg-mgmt       = module.common_infrastructure.nsg-mgmt
+  source           = "./modules/jumpbox"
```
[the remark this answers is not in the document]
suhanselvan: Yes. This was already agreed upon way back.

deploy/v2/terraform/modules/common_infrastructure/main.tf
suhanselvan: @persiaAziz the 4 byte random id is prefixed by a fixed text depending on the storage account being created and was deemed sufficient during the last sprint sync call.

suhanselvan:
@rukawata The pipeline is expecting the status for a test and I'm not sure what it is. Can you please look into it?

rukawata:
> @rukawata The pipeline is expecting the status for a test and I'm not sure what it is. Can you please look into it?

That test does not run against code under v2. It's okay that it shows as expected.
The thing blocks the merge is change requested from @persiaAziz - once she approves your PR, you would be able to merge.

PR #237, opened 2019-10-03
Add HANA System Availability check

tniek:
- This PR adds a new check to the HANA monitoring content: [M_SYSTEM_AVAILABILITY](https://help.sap.com/viewer/4fe29514fd584807ac9f2a04f6754767/2.0.04/en-US/1ef9723a03214bd889c4fb8947765aa4.html) is a time-series query that provides important information on individual HANA services (or an entire instance) starting/stopping/failing over.
- Since the underlying HANA monitoring view uses different column names than our other checks (e.g. EVENT_TIME instead of TIME), a few modifications were necessary to the `prepareSql()` method.

### DIFF
--- a/monitor/SapHana.json
+++ b/monitor/SapHana.json
@@ -1,6 +1,6 @@
 {
         "contentType": "HANA",
-        "contentVersion": "0.1.0",
+        "contentVersion": "0.2.0",
         "checks": [
                 {
                         "name": "HostConfig",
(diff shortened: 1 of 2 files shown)


monitor/SapHana.json
```diff
@@ -18,9 +18,20 @@
...
+                                "initialTimespanSecs": 3600,
```
persiaAziz: Any reason we jumped to an hour on the initial time span?

tniek: 30 seconds was actually way too short for Load History (that's just three rows).
I had played around with different thresholds, and 360 rows (for the very first execution) seems reasonable.

PR #238, opened 2019-10-04
V2: Ansible - RHEL OS and Disk configurations 

saddi7087:
Ansible code to configure the OS disks and installs the pre-requisites for SAP HANA Install.


deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
+- name: 
+  include_vars: 
+    file: ../hdb_sizes.json
+    name: hdb_sizes
+   
+- name:
+  include_vars:
```
rukawata: Instead of putting everything into the` tasks/main.yml`, can we use` vars/main.yml` for all the variables? Same apply to all tasks.

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+    lv: "lv_hana_{{item}}"
+    size: 100%FREE
+  with_items: 
+   - "{{ devices }}"
+  ignore_errors: true
+
+- set_fact:
+   mountdirs: "{{devices| reject('search','backup','sap') | list }}+['backup']"
+
+- debug: var=mountdirs
```
rukawata: What's this debug info for? There is not a output msg...

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+    - "{{ mountdirs }}"
+
+- name: creating filesystem on LVMs
+  filesystem:
+     fstype: "{{ output.databases[0].filesystem |default('xfs')}}"
+     dev: "/dev/vg{{ item }}/lv_hana_{{ item }}"
+     opts: "su=64k,sw=2 "
+  with_items: 
+    - "{{ mountdirs }}"
+  ignore_errors: True
```
rukawata: Is there a reason to ignore the error and continue with the playbook?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
...
+    name: "/hana/{{ item }"
+    fstype: "{{ output.databases[0].filesystem  | default('xfs')}}"
+    opts: defaults
+    src: "/dev/vg{{ item }}/lv_hana_{{ item }}"
+    state: mounted
+  with_items:
+    - "{{ mountdirs }}"
+  ignore_errors: True
+
+- name: create sap  mount point directories
```
rukawata: Please put all mount point creation together.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
+---
+
+- name: Installing OS Packages
+  yum:
+    name:
```
rukawata: The name should be a list variable and use with_items here to loop though the list.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+      - bzip2-devel
+      - libffi-devel
+    state: present
+  when: ansible_os_family == "RedHat"
+
+
+- name: Installing OS Packages
+  zypper:
+    name:
+      - libyui-qt-pkg7
```
rukawata: Same as above

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+      - tuned-profiles-sap-hana
+      - compat-sap-c++-7
+      - libatomic
+      - numad
+      - gcc 
+      - openssl-devel
+      - bzip2-devel
+      - libffi-devel
+    state: present
+  when: ansible_os_family == "RedHat"
```
rukawata: Where is ansible_os_family  defined?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+  systemd:
+     name: chronyd
+     state: stopped
+
+- name: Add ntp servers to chrony.conf
+  blockinfile:
+    path: /etc/chrony.conf
+    block: |
+      server ntp.example.com
+      server ntp1.example.com}
```
rukawata: a single '}'? Typo?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+      server ntp2.example.com
+
+- name: reload chrony service
```
rukawata: Not sure why we need chrony here...@mwhatleyjr can you confirm whether SAP on Azure uses chrony?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+
+- name: Disable SELinux
+  selinux:
+    state: disabled
+
+- name: SAP HANA NUMA configuration 
+  copy:
+    dest: /etc/sysctl.d/sap_hana.conf
+    content: |
+       kernel.numa_balancing=0
```
rukawata: This is OS recommendation for **RHEL** only (SAP Note 2292690 - SAP HANA DB: Recommended OS settings for RHEL 7).

Also are we targeting to have all those OS optimization in our script? It may become a challenge when SAP changes the recommendation - how do we keep track and update our code? I would think those recommendations are already considered when we build the SAP specific image in our gallery.

There are a couple of similar things below, eg regarding transparent_hugepage.

@mwhatleyjr, what's your take on this?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/src/Python-3.7
```
rukawata: Python3 is pre-installed on the VM (at least for the SLES image we use).
There is no need to add  deploy/v2/ansible/softwares/Python-3.7.3.tgz here.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
mwhatleyjr: We do need a ntp services - nothing wrong with chrony, i'll check to see what's been used in the past.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
mwhatleyjr: we might have to keep a dictionary or table to make it easier to identify what is in or what needs to be added for newer OS versions

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,105 @@
+- name: 
+  include_vars: 
+    file: ../hdb_sizes.json
+    name: hdb_sizes
+   
+- name:
+  include_vars:
+    file: "/home/{{ ansible_user}}/output.json"
```
persiaAziz: Can we call it something meaningful other than output

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+import json
+import os
+import sys
+size=sys.argv[1]
```
octavianhornoiu: Arguments should go into a parser such as: https://docs.python.org/3/howto/argparse.html

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+cmd="ls -ld /sys/block/sd*/device|grep 5.0|awk '{print $9,$11}'|awk -F '/' '{print $4,$8}'|awk -F ':' '{print $1 $4}'|sort -u -k2|awk '{print $1}'"
```
octavianhornoiu: This is the kind of thing we used to do in 1990.  Today it is 2019, please update this to real code.

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+jsonFile=open("/tmp/hdb_sizes.json")
+jsonString=jsonFile.read()
+data=json.loads(jsonString)
+disks=data[size]['storage']
+y=0;
+x=1;
+for dict in disks:
+     dk_name=dict['name']
+     dk_count=dict['count']
+     if dict['name'] != 'os':
```
pabowers: you should use the variable `dk_name` here since it is already created

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/src/Python-3.7
+import json
+import os
+import sys
+size=sys.argv[1]
+cmd="ls -ld /sys/block/sd*/device|grep 5.0|awk '{print $9,$11}'|awk -F '/' '{print $4,$8}'|awk -F ':' '{print $1 $4}'|sort -u -k2|awk '{print $1}'"
+stream = os.popen(cmd)
```
octavianhornoiu: You are using raw python calls instead of methods to read in responses from the shell.  This is unacceptable in 2019.

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
pabowers: I agree, we shouldn't ignore the error

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+disk_list = stream.read()
+disklist=disk_list.splitlines( )
```
octavianhornoiu: You have an actual space character inside splitlines() call which is completely unnecessary and perhaps even syntactically wrong.  Did you use an IDE?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+jsonFile=open("/tmp/hdb_sizes.json")
+jsonString=jsonFile.read()
+data=json.loads(jsonString)
```
octavianhornoiu: This should be a method/function

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+disks=data[size]['storage']
```
octavianhornoiu: What is "disks"?  There are no comments of any kind.  What is this thing?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+y=0;
+x=1;
+for dict in disks:
```
octavianhornoiu: Overloading the default 'dict' namespace is bad coding practice and no one does this... ever.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,233 @@
...
+     state: stopped
+     enabled: no
+
+- name: Disable and stop firewall
+  systemd:
+     name: firewalld
+     state: stopped
+     enabled: no
+
+- name: download  Python3
```
pabowers: Why can't we just use the package manager to install python instead of unarchiving it?

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+- hosts: "hdb0"
+  become: true
+  become_user: root
+  vars:
+    devices: []
+  roles:
+#    - { role: OS_lib_install }
```
pabowers: Please remove commented out code

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+size=sys.argv[1]
+cmd="ls -ld /sys/block/sd*/device|grep 5.0|awk '{print $9,$11}'|awk -F '/' '{print $4,$8}'|awk -F ':' '{print $1 $4}'|sort -u -k2|awk '{print $1}'"
+stream = os.popen(cmd)
+disk_list = stream.read()
+disklist=disk_list.splitlines( )
+jsonFile=open("/tmp/hdb_sizes.json")
+jsonString=jsonFile.read()
+data=json.loads(jsonString)
+disks=data[size]['storage']
+y=0;
```
octavianhornoiu: What does 'y' stand for?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+x=1;
```
octavianhornoiu: What does 'x' stand for?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+for dict in disks:
+     dk_name=dict['name']
+     dk_count=dict['count']
```
octavianhornoiu: What does 'dk' mean?
What are we naming?
What are we counting?
Why?

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+     if dict['name'] != 'os':
+        diskstring=''
```
octavianhornoiu: This violates pep8 standards as there is no space before and after '=' assignment.

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+        for x in range(0, dk_count):
+            diskstring = diskstring +' '+ '/dev/'+disklist[y]+'1'
```
octavianhornoiu: No one uses "+" string formatters anymore unless they are programming in PHP.  Please use normal string formatting methods.

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+            y += 1;
+        vgcmd='vgcreate ' +'vg' + dk_name  + ' ' + diskstring
```
octavianhornoiu: PEP8 violation in assignment.
String formatting using '+'
Why do some things have quotes around them as literals and some are passed in as just vars?  Why aren't the first two literals combined since they are both literals?  How does this even happen in code?

Another PEP8 violation in '+'vg''

deploy/v2/ansible/disks.py
```diff
@@ -0,0 +1,27 @@
...
+        stream = os.popen(vgcmd)
+        print (stream)
```
octavianhornoiu: Why do we not have a logging function?  Print is used for debugging and not for actual production code.  Time to create a logging method perhaps?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: I have moved the var files to main playbook, so it apply to all tasks

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: removed the debug from the code

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: removed the ignore_errors from the code

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: All the mounts are kept together, this requires the mount path variable in the hdb_sizes json file. assuming mount point path variable in hdb_sizes json, I have kept all mount point together

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: Ansible deprecated use of with_items to loop on package modules like yum, i have options to use name, pkg and package, so i have used name.

Using a loop on a package module via squash_actions
The use of squash_actions to invoke a package module, such as “yum”, to only invoke the module once is deprecated, and will be removed in Ansible 2.11.

Instead of relying on implicit squashing, tasks should instead supply the list directly to the name, pkg or package parameter of the module. This functionality has been supported in most modules since Ansible 2.3.

OLD In Ansible 2.6 (and earlier) the following task would invoke the “yum” module only 1 time to install multiple packages

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: ansible_os_family is from host variables. it is ansible global variable set from the gather facts.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: Yes, It is a typo and now it is fixed.

deploy/v2/ansible/disks.py
saddi7087: Python3 is not installed on RHEL images, if it already installed, please provide the path I will start using it.

saddi7087: Updated the python script

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: This output json is coming from terraform, please let me know what variable to be used.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: The point is to use a list variable instead of listing all packages in the task.
You can use with_items or loop or whatever...
Same rules for yum as well as zypper.

Something like below would be more acceptable.
```yaml
- name: Install packages
  yum:
    name: "{{ packages }}"
    state: present
```

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+     state: stopped
+     enabled: no
+
+- name: Disable and stop abrtd
+  systemd:
+     name: abrtd
+     state: stopped
+     enabled: no
+
+- name: Add always and Never to transparent Hugepages
```
persiaAziz: I don't understand the name of the play here. The name suggests to configure hugepages , but it is actually setting the corefile size to 0
Also any reason the corefile size is set to 0?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+
+- name: copy the disks script
+  copy: src="disks.py" dest="/tmp/disks.py"
+
+- name: copy the hdbs sizes json script
+  copy: src="../hdb_sizes.json" dest="/tmp/hdb_sizes.json"
+
+- name: Physical volumes
+  shell: "pvcreate /dev/{{item}}1"
+  with_items: "{{ raw_disks }}"
```
rukawata: As pointed out yourself in [comment](https://github.com/Azure/sap-hana/pull/238#discussion_r332502166), with_items  will be deprecated. There are a couple of other places where you use with_items. Please replace them with loop.

saddi7087: with_items will be deprecated only for package modules, with_items is available for all other modules.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+  systemd:
+     name: chronyd
+     state: started
+
+- name: restart chronyd service
+  systemd:
+     name: chronyd
+     state: restarted   
+
+- name: stop chronyd service
```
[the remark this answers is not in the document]
saddi7087: It is typo, name was wrong, I will fix and push the code

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: Yes, Name was wrong. I have fixed and committed the code. To disable the core dumps for users as per standard SAP documents.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: Updated the code with lists of packages

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+
+- name: Physical volumes
+  shell: "pvcreate /dev/{{ item }}1"
+  with_items: "{{ raw_disks }}"
+
+- name: Get disks information and physical volume groups
+  shell: "/usr/local/bin/python3.7 /tmp/disks.py {{ output.databases[0].size }}"
+  register: disk_info
+
+- name: Create logical volumes
```
rukawata: Hi @saddi7087, in the interest of time, I've looked into this myself. And here is my engineered solution for your problem. You will need to make adjustment so it works inside your task. Below is just to demonstrate how this can be achieved.

```yaml
- hosts: all
  become: true

  vars:
    disk_dict: {}
    start: 0

  tasks:
  - name: create device list
    shell: ls -l /dev/disk/azure/scsi1/* | awk '{split($0,disks," "); system("readlink -f " disks[9])}'
    register: device_list

  - name: create dictionary with mount point and devices list
    set_fact:
      start: "{{ start|int+item.value.number_disks }}"
      disk_dict: "{{ disk_dict | combine({ item.key: {'count': item.value.number_disks, 'devices': device_list.stdout_lines[start|int:start|int+item.value.number_disks]} }) }}"
    loop: "{{ S.storage|dict2items }}"
    when: item.key != 'os'

  - debug:
      msg: "{{ disk_dict }}"
```

And output looks like this, to show the created dictionary with list of devices:

```
TASK [debug] *************************************************************************************************************
ok: [xxxxxx] => {
    "msg": {
        "data": {
            "count": 2,
            "devices": [
                "/dev/sdc",
                "/dev/sdd"
            ]
        },
        "log": {
            "count": 2,
            "devices": [
                "/dev/sdf",
                "/dev/sdg"
            ]
        },
        "shared": {
            "count": 1,
            "devices": [
                "/dev/sde"
            ]
        }
    }
}
```

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,51 @@
...
+
+- name: Create device list
+  shell: ls -l /dev/disk/azure/scsi1/* | awk '{split($0,disks," "); system("readlink -f " disks[9])}'
+  register: device_list
+
+- name: Create dictionary with disk name and devices list
+  set_fact:
+    start: "{{ start|int+item.count }}"
+    disk_dict: "{{ disk_dict | combine({ item.name: {'count': item.count, 'devices': device_list.stdout_lines[start|int:start|int+item.count]} }) }}"
+  loop: "{{ hdb_sizes[output.databases[0].size].storage }}"
```
rukawata: The T-shirt size should not come from a fixed item from database list.
The way it is now it will always try to search for the first element from the list.
We should pull the T-shirt size from the element from the list with `platform` key values `HANA`.

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+- hosts: "hdb0"
```
rukawata: The host should be in an inventory list rather than hard coded in the playbook.

saddi7087: currently host will taken from the inventory list but I have named inventory list as hdb0 and I renamed to "hanadbnodes".

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: Modified code to get the T-shirt size from database list.

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,21 @@
+---
+
```
rukawata: I think ansible is not installed by Suhan's [PR #234/Azure/sap-hana/pull/234).
@suhanselvan can you comment on that?

deploy/v2/ansible/roles/SAP_HANA_server_Install/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
+---
```
rukawata: As discussed, please remove the installation from this PR.
The installation should be a independent task and please work on a different PR for this in parallel. Thanks!

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,142 @@
+---
+
+- name: Installing OS packages
+  yum:
+    name: "{{ packages }}"
+  vars:
+    packages:
+      - "@base"
```
rukawata: As discussed, this should be moved to ../var/main.yml.
This way the code looks cleaner and easier to understand.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: As discussed, Matt will schedule a follow up meeting with @mwhatleyjr on this.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,142 @@
...
+      - tuned-profiles-sap-hana
+      - compat-sap-c++-7
+      - libatomic
+      - numad
+      - gcc 
+      - openssl-devel
+      - bzip2-devel
+      - libffi-devel
+    state: present
+  when: ansible_os_family == "RedHat"
```
rukawata: Can we use include_tasks to create a dynamic task list?
It means, for example, we will have two yml files:

- RedHat_os_config.yml
- SUSE_os_config.yml

In the main.yml, it looks like something:
```yaml
- name: config RedHat
  include_tasks: "{{ ansible_os_family }}_os_config.yml"
  when:  ansible_os_family == "RedHat"

- name: config SUSE
  include_tasks: "{{ ansible_os_family }}_os_config.yml"
  when:  ansible_os_family == "SUSE"

```

rukawata:
> [CLA assistant check](https://cla.opensource.microsoft.com/Azure/sap-hana?pullRequest=238)
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.microsoft.com/Azure/sap-hana?pullRequest=238) before we can accept your contribution.
> 
> ❌ saddi7087 [sign now](https://cla.opensource.microsoft.com/Azure/sap-hana?pullRequest=238)
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.microsoft.com/check/Azure/sap-hana?pullRequest=238) it.

@saddi7087 please make sure you sign the CLA.

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  set_fact:
+    start: "{{ start|int+item.count }}"
+    disk_dict: "{{ disk_dict | combine({ item.name: {'count': item.count, 'devices': device_list.stdout_lines[start|int:start|int+item.count]} }) }}"
+  loop: "{{ hdb_sizes[output.databases | map(attribute='size')| join(',')].storage }}"
+  when: item.name != 'os'
+
+- name: Create physical volume group
+  lvg: 
+    vg: "vg{{ item.key }}"
+    pvs: "{{ item.value.devices | join (',') }}"
```
rukawata: The value for key 'devices' should be a list already, it should be okay without join them with ','.
Please check that.

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: Create dictionary with disk name and devices list
+  set_fact:
+    start: "{{ start|int+item.count }}"
+    disk_dict: "{{ disk_dict | combine({ item.name: {'count': item.count, 'devices': device_list.stdout_lines[start|int:start|int+item.count]} }) }}"
+  loop: "{{ hdb_sizes[output.databases | map(attribute='size')| join(',')].storage }}"
+  when: item.name != 'os'
+
+- name: Create physical volume group
+  lvg: 
+    vg: "vg{{ item.key }}"
```
rukawata: Can this be `vg_{{ item.key }}` to be consistent with the formatting of `lv_hana_{{ item.key }}`?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    lv: "lv_hana_{{ item.key }}"
+    size: 100%FREE
+  with_dict: "{{ disk_dict }}"
+  register: logical_vg
+  failed_when: logical_vg is failed
+
+- name: Create mount point directories
+  file:
+    path: "{% if item.key=='sap' %}/usr/{{ item.key }}{% else %}/hana/{{ item.key }}{% endif %}"
```
rukawata: Should the mount point include SID?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,142 @@
...
+
+- name: Disable and stop firewall
+  systemd:
+     name: firewalld
+     state: stopped
+     enabled: no
+
+- name: Reboot the machine if SELinux is disabled
+  reboot:
+  when: se_state.changed and ansible_os_family == 'RedHat'
```
rukawata: I think we are missing the additional lib installation for HANA 2 SPS04: see [SAP Note 2593824](https://launchpad.support.sap.com/#/notes/2593824)
The version is part of the input JSON: `"db_version": "2.00.043"`

rukawata: I think we also miss the SWAP space configuration.
You can pull the swap information from hana-sizing.json `"swap_size_gb": 2`

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+
+- name: Creating filesystem on LVMs
+  filesystem:
+     fstype: "{{ output.databases|map(attribute='filesystem')| join('\n') | default('xfs') }}"
+     dev: "/dev/vg{{ item.key }}/lv_hana_{{ item.key }}"
+  with_dict: "{{ disk_dict }}"
+  register: file_system
+  failed_when: file_system is failed
+
+- name: Add fstab entries and mount volumes
```
rukawata: Where is the `nofail` added?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: Yes, modified with vg_{{ item.key }}

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: I have removed the join

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: All the packages can be kept in one list with the ansible_os_family.
- name: Installing OS packages
  package:
    name: "{{ item }}"
    state: present
  with_items: "{{ packages[ansible_os_family] }}"

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: I don't have access to the SAP Note, can you please provide the missing lib's.

saddi7087: Swap space is configured by default on the machine while terraform provisioning, so i did not configured the SWAP.
[azureadm@hdb1 ~]$ free -m
              total        used        free      shared  buff/cache   available
Mem:          16031         561       12616          48        2852       15035
Swap:          2047           0        2047
[azureadm@hdb1 ~]$ swapon -s
Filename				Type		Size	Used	Priority
/mnt/resource/swapfile                 	file	2097148	0	-1

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: Disk configuration has to fail, if any task in disk configuration is failed, so i didn't add any nofail in the disk configuration role.

deploy/v2/ansible/roles/SAP_HANA_server_Install/tasks/main.yml
saddi7087: Yes, HANA installation is removed from the PR.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: moved the packages to the var files

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
saddi7087: Marshall confirmed SID is not required in mount point.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,123 @@
+---
```
rukawata: @saddi7087 this is sample code I have to install beautifulsoup for python 2.7. Please adjust so it works for python3.
Here is also a link so we install pip in an OS-independent fashion: [pip install](https://pip.pypa.io/en/stable/installing/)

```yaml
- name: Download pip
  get_url:
    url: https://bootstrap.pypa.io/get-pip.py
    dest: /usr/bin/get-pip.py
  become: true

- name: Install Pip
  command: |
    python /usr/bin/get-pip.py
  become: true

- name: Install beautifulsoup4
  pip:
    name: beautifulsoup4
    state: latest
  become: true

```

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: I used the latest code from master to provision a new HANA server, however I don't see swap space configure - can you double check on your side?

```
hdb-node1:~ # swapon -s
hdb-node1:~ # cat /proc/swaps
Filename                                Type            Size    Used    Priority
hdb-node1:~ # file /mnt/resource/swapfile
/mnt/resource/swapfile: cannot open `/mnt/resource/swapfile' (No such file or directory)
```

To add swap space on azure linux VM, you can use [waagent](https://support.microsoft.com/en-us/help/4010058/how-to-add-a-swap-file-in-linux-azure-virtual-machines)

Sample code as below (this is the swap space I configured for **S4** scenario):
```yaml
- name: set SWAP space
  ini_file:
    path: dest=/etc/waagent.conf
    section: null
    option: "{{ item.option }}"
    value: "{{ item.value }}"
  with_items:
    - {option: "ResourceDisk.Format", value: "y"}
    - {option: "ResourceDisk.EnableSwap", value: "y"}
    - {option: "ResourceDisk.SwapSizeMB", value: "20480"}
  become: true

- name: restart waagnet
  command: service waagent restart
  become: true

```

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: As discussed, since RHEL out of box image does not have many things pre-configured as SLES, we will incorporate the specific configuration.

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
rukawata: I see it's now added. Thanks.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: @saddi7087 the above steps work for me on a SLES 12 SP4:

```
hdb-node1:~ # vi /etc/waagent.conf
hdb-node1:~ # service waagent restart
hdb-node1:~ # swapon -s
Filename                                Type            Size    Used    Priority
/mnt/resource/swapfile                  file            2097148 0       -1
hdb-node1:~ # cat /proc/swaps
Filename                                Type            Size    Used    Priority
/mnt/resource/swapfile                  file            2097148 0       -1

```
And according to this [link](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/suse-create-upload-vhd), it has the same steps for configure SWAP space on SUSE. 
Make sure you restart the waagent after changing the parameter.

saddi7087: Yes, it is working for SUSE. the problem was with the ini_file module was modifying the /etc/waagent.conf file so i replaced the code with lineinfile and tested it working with both Suse and RHEL.

rukawata: > I think we are missing the additional lib installation for HANA 2 SPS04: see [SAP Note 2593824](https://launchpad.support.sap.com/#/notes/2593824)
> The version is part of the input JSON: `"db_version": "2.00.043"`

I see you add the task for SWAP space. How about his one?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: Hi @saddi7087, I see you use the above code for the package install. I was more referring to not only the package installation, but also the OS specific and non-OS specific configuration changes. To make it more complete, we should also have one yml for common configurations for both SUSE and RedHat, things like "SAP HANA NUMA configuration", "Disable & numad service", etc...

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: I have added the additional libs to packages yaml. 
Rhel
   - compat-sap-c++-7
   - libatomic

Suse
   - libgcc_s1
   - libstdc++6
   - libatomic1

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: modified the code to address the OS specific and non-OS specific configuration changes.

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: @saddi7087 , those libs are not always required (only required for HANA 2.0 SP04). We should check the HANA version before we install them, which can be done by checking input JSON: `"db_version": "2.00.043"`.

deploy/v2/ansible/sap_playbook.yml
rukawata: @saddi7087 can you check the above questions + below one additional quesiton with Suhan?

 3. When will the terraform create inventory file for Ansible?

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
rukawata: As discussed, this will be part of the downloader role.

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
+---
```
rukawata: @saddi7087 question regarding the path. 
According to [this fixed bug](https://github.com/ansible/ansible/issues/33784), `include_tasks` should look for configurations under task of the role, but not root. Therefore, I think those dynamic yml files should be under `deploy/v2/ansible/roles/OS_lib_install/tasks`. Can you please check? Also please use structure like:

```
OS_lib_install
└── tasks
    ├── common
    │   └── common_os_config.yml
    ├── main.yml
    ├── RedHat
    │   └── RedHat_os_config.yml
    └── Suse
        └── Suse_os_config.yml
```

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
...
+  with_items: "{{ packages[ansible_os_family] }}"
+
+- name: Installing Hana 2.0 SPS04  packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+      - compat-sap-c++-7
+      - libatomic
+  when: "{{ output.databases|map(attribute='version')| join(',')}} == '2.00.043'"
```
rukawata: Why fixed to 2.00.043?

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
...
+    content: |
+       kernel.numa_balancing=0
+
+- name: Save sysctl.conf config
+  command: sysctl -p /etc/sysctl.d/sap_hana.conf 
+
+- name: Save sysctl.conf config
+  command: sysctl -p /etc/sysctl.d/sap_hana.conf 
+
+- name: Disable & numad service
```
rukawata: 1. Let's keep the name of the task the same as in SAP Note (eg. SAP Note 2292690 - SAP HANA DB: Recommended OS settings for RHEL 7), as well the order of all optimization items.
2. This belongs to "Turn off auto-numa balancing" and should be together with what you wrote in  SAP HANA NUMA configuration. It's awkward to have sysctl.conf in between to separate them.

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
...
+  service:
+     name: numad
+     state: stopped
+     enabled: no
+
+- name: Create a symbolic link
```
rukawata: What is this sym link for? I don't see it in the RHEL optimization note.

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
...
+  file:
+    src: "{{ item.prop }}"
+    dest: "{{ item.value }}"
+    state: link
+  with_items:
+    - { prop: '/usr/lib64/libssl.so.10', value: '/usr/lib64/libssl.so.1.0.1' }
+    - { prop: '/usr/lib64/libcrypto.so.10', value: '/usr/lib64/libcrypto.so.1.0.1' }
+
+- name: Disable and stop abrt-ccpp
```
rukawata: Looks like this is for RHEL-8...Let's not mix different OS versions.

deploy/v2/ansible/dynamic/RedHat_os_config.yml
```diff
@@ -0,0 +1,85 @@
...
+     enabled: no
+
+- name: Disable and stop abrtd
+  systemd:
+     name: abrtd
+     state: stopped
+     enabled: no
+
+- name: Disable and stop kdump
+  systemd:
```
rukawata: Same as https://github.com/Azure/sap-hana/pull/238/files#r337332407

deploy/v2/ansible/dynamic/common_os_config.yml
```diff
@@ -0,0 +1,43 @@
...
+- name: Transparent_hugepage update to never and configure processor C-States
+  lineinfile: 
+     dest: /etc/default/grub 
+     regexp: "GRUB_CMDLINE_LINUX=" 
+     line: "GRUB_CMDLINE_LINUX='crashkernel=auto  @ vconsole.keymap=us @kernel_arguments: transparent_hugepage=never intel_idle.max_cstate=1 processor.max_cstate=1'"
+
+- name: Modify transparent_hugepage
+  command: grub2-mkconfig -o /boot/grub2/grub.cfg
+
+- name: Processes for the SAPSYS
```
rukawata: I'm not sure if this applies for both SUSE and RHEL. Please check.

deploy/v2/ansible/dynamic/Suse_os_config.yml
```diff
@@ -0,0 +1,24 @@
...
+  with_items: "{{ packages[ansible_os_family] }}"
+
+- name: Installing Hana 2.0 SPS04  packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - libgcc_s1
+    - libstdc++6
+    - libatomic1
```
rukawata: As discussed, please add condition here (only required for HANA 2.0 SP4).

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
saddi7087: modified the code to install libs based on the db version

deploy/v2/ansible/dynamic/RedHat_os_config.yml
saddi7087: modified the os_lib_install structure as above.

deploy/v2/ansible/dynamic/RedHat_os_config.yml
saddi7087: db version parameter is modified to support any HANA2 SPS40

deploy/v2/ansible/roles/OS_lib_install/tasks/Suse/SUSE-12/os_config.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Installing Hana 2.0 SPS04  packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - libgcc_s1
+    - libstdc++6
+    - libatomic1
+  when: output.databases|map(attribute='version')|select('match', ('2.[0-9][0-9].0[4-9][0-9]'))|list
```
rukawata: Two things regarding the version check:

1. The attribute should be called `db_version`.
2. `databases` is a list. We should loop and find `platform: HANA`, then check the version.

Below is a sample you could reference to. Since we may use version information in several places, the task to find HANA version should only run once, then register the version as an Ansible variable.

```yml
  tasks:
    - name: get dictionary for HANA
      set_fact: hana_database="{{ item }}"
      loop: "{{ output.databases|flatten(levels=1) }}"
      when: item.platform == 'HANA'

    - name: check db_version > HANA 2.0 SP04
      debug:
        msg: "{{ hana_database.db_version }}"
      when: hana_database.db_version is regex("2.\d{2}.0([4-9]\d|[0-9]\d{2,})")

```

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,82 @@
...
+  with_items: "{{ packages[ansible_os_family] }}"
+
+- name: Installing Hana 2.0 SPS04  packages
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+      - compat-sap-c++-7
+      - libatomic
+  when: output.databases|map(attribute='version')|select('match', ('2.[0-9][0-9].0[4-9][0-9]'))|list
```
rukawata: See [comment ](https://github.com/Azure/sap-hana/pull/238/files#r337777440).

deploy/v2/ansible/sap_playbook.yml
rukawata: add one more items for @suhanselvan @saddi7087 

4. hdb_sizes is not on the jumpbox with the current terraform code.

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,43 @@
...
+
+- name: Set SWAP space
+  lineinfile:
+    dest: /etc/waagent.conf
+    regexp: "^{{ item.prop }}="
+    line: "{{ item.prop }}={{ item.value }}"
+  with_items:
+    - { prop: "ResourceDisk.Format", value: "y" }
+    - { prop: "ResourceDisk.EnableSwap", value: "y" }
+    - { prop: "ResourceDisk.SwapSizeMB", value: "{{ (hdb_sizes[output.databases | map(attribute='size')| join(',')].compute.swap_size_gb) * 1024|int }}" }
```
rukawata: Please change `output.databases | map(attribute='size')| join(',')` to implementation like below:

```yml
  tasks:
    - name: get dictionary for HANA
      set_fact: hana_database="{{ item }}"
      loop: "{{ output.databases|flatten(levels=1) }}"
      when: item.platform == 'HANA'

    - name: check hdb size
      debug:
        msg: "{{ hana_database.size }}"

```

deploy/v2/ansible/sap_playbook.yml
suhanselvan: @rukawata My latest PR addresses these requirements.

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
saddi7087: Modified the code to HANA DB dictonary

deploy/v2/ansible/roles/OS_lib_install/tasks/Suse/SUSE-12/os_config.yml
saddi7087: Modified the code to use HANA DB dictionary.

deploy/v2/ansible/dynamic/Suse_os_config.yml
saddi7087: Modified the code to use condition (only required for HANA 2.0 SP4).

deploy/v2/ansible/dynamic/common_os_config.yml
saddi7087: Yes, It applies to both OS

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
rukawata: Thx. I see most of the places have been updated except for fetching the size. Please check.

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
+---
+
+- name: Packages3.2 - Installing OS package
+  package:
+    name: "{{ item }}" 
+    state: present
+  with_items: 
+    - "{{ packages[ansible_os_family] }}"
+
+- name: 2593824 - Additional Packages HANA 2 SPS04
```
pabowers: Why is the name 2593824?

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
...
+- name: SELinux3.7 - Disable SELinux1
+  selinux:
+    state: disabled
+  register: se_state
+
+- name: SELinux3.7 - Disable SELinux2
+  reboot:
+  when: se_state.changed
+
+- name: Disable3.8 - Disable Automatci NUMA Balancing1
```
pabowers: Typo, should be `Automatic`. Also why 1, 2, and 3?

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
...
+- name: 2593824 - Additional Packages HANA 2 SPS04
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+      - compat-sap-c++-7
+      - libatomic
+  when: hana_database.db_version is regex("2.\d{2}.0([4-9]\d|[0-9]\d{2,})")
+
+- name: SELinux3.7 - Disable SELinux1
```
pabowers: These names are confusing, why name them by part number? Can you instead name them according to what the job does?

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: restart waagnet
```
Caiyeon: Typo: waagent*

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
...
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States1
+  command: "echo never > /sys/kernel/mm/transparent_hugepage/enabled"
+
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States2
+  lineinfile: 
+     dest: /etc/default/grub 
+     regexp: "GRUB_CMDLINE_LINUX=" 
+     line: "GRUB_CMDLINE_LINUX='crashkernel=auto  @ vconsole.keymap=us @kernel_arguments: transparent_hugepage=never intel_idle.max_cstate=1 processor.max_cstate=1'"
+
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States3
```
pabowers: This name is super long

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
saddi7087: Fixed typo

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
saddi7087: This is the SAP NOTE no, I have added the SAP Note to the name

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
saddi7087: fixed the Automatic typo. I was asked to keep the naming with section no and task title of SAP document, if there are multiple tasks suggested to keep 1,2,3 at the end of task.

Section - Task title and no
3.8 - Disable Automatic NUMA Balancing 1
3.8 - Disable Automatic NUMA Balancing 2
3.8 - Disable Automatic NUMA Balancing 3

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
saddi7087: I was asked to keep the naming with section no and task title of SAP document, if there are multiple tasks suggested to keep 1,2,3 at the end of task.

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
saddi7087: Yes, this is super long. because this task is for THP and C-Processors states and for both RHEL and SUSE OS's.

deploy/v2/ansible/roles/OS_lib_install/tasks/Suse/SUSE-12/os_config.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+- name: Install OS packages
+  package: 
+    name: "{{ item }}" 
+    state: present
+  with_items: "{{ packages[ansible_os_family] }}"
```
persiaAziz: should we pin the versions of these packages??

persiaAziz:
Looks okay overall. My only concern is if we need to pin the versions?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  register: device_list
+
+- name: Create dictionary with disk name and devices list
+  set_fact:
+    start: "{{ start|int+item.count }}"
+    disk_dict: "{{ disk_dict | combine({ item.name: {'count': item.count, 'devices': device_list.stdout_lines[start|int:start|int+item.count]} }) }}"
+  loop: "{{ hdb_sizes[hana_database.size].storage }}"
+  when: item.name != 'os'
+
+- name: Physical volume group
```
tniek: "Create physical volume group"

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  lvg: 
+    vg: "vg_{{ item.key }}"
+    pvs: "{{ item.value.devices }}"
+  with_dict: "{{ disk_dict }}"
+  register: physical_vg
+  failed_when: physical_vg is failed
+
+- name: Logical volume group
```
tniek: "Create logical volume group"

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
+---
+
+- name: Create device list
+  shell: ls -l /dev/disk/azure/scsi1/* | awk '{split($0,disks," "); system("readlink -f " disks[9])}'
```
tniek: What happens if the list provided by `ls -l` is empty?

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  lvol:
+    state: present
+    vg: "vg_{{ item.key }}"
+    lv: "lv_hana_{{ item.key }}"
+    size: 100%FREE
+  with_dict: "{{ disk_dict }}"
+  register: logical_vg
+  failed_when: logical_vg is failed
+ 
+- name: Mount point directories
```
tniek: "Create mount points"

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  file:
+    path: "{% if item.key=='sap' %}/usr/{{ item.key }}{% else %}/hana/{{ item.key }}{% endif %}"
```
tniek: This should not be handled here!
Instead, we could define a mount point in the sizing.json

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    state: directory
+  with_dict: "{{ disk_dict }}"
+
+- name: Filesystems
```
tniek: "Create file systems"

deploy/v2/ansible/roles/OS_disks_preparation/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  filesystem:
+     fstype: "{{ hana_database.filesystem | default('xfs') }}"
+     dev: "/dev/vg_{{ item.key }}/lv_hana_{{ item.key }}"
+  with_dict: "{{ disk_dict }}"
+  register: file_system
+  failed_when: file_system is failed
+
+- name: Mount volumes
+  mount:
+    path: "{% if item.key=='sap' %}/usr/{{ item.key }}{% else %}/hana/{{ item.key }}{% endif %}"
```
tniek: See above, no logic should be handled here

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
+---
+
+- name: Packages3.2 - Installing OS package
```
tniek: "Step 3.2 - Install OS package"

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
...
+- name: SAP Note 2593824 - Additional Packages HANA 2 SPS04
+  package:
+    name: "{{ item }}"
+    state: present
+  with_items:
+      - compat-sap-c++-7
+      - libatomic
+  when: hana_database.db_version is regex("2.\d{2}.0([4-9]\d|[0-9]\d{2,})")
+
+- name: 3.7 - Disable SELinux 1
```
tniek: "Step 3.7 - Disable SELinux 1"

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
...
+  selinux:
+    state: disabled
+  register: se_state
+
+- name: 3.7 - Disable SELinux 2
```
tniek: From here on, I'll not comment on every single `name`...
Just add "Step" in the beginning of each name, so it's clear what we're referring to.

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
```diff
@@ -0,0 +1,99 @@
...
+
+- name: 3.9 - Add Symbolic Links
+  file:
+    src: "{{ item.prop }}"
+    dest: "{{ item.value }}"
+    state: link
+  with_items:
+    - { prop: '/usr/lib64/libssl.so.10', value: '/usr/lib64/libssl.so.1.0.1' }
+    - { prop: '/usr/lib64/libcrypto.so.10', value: '/usr/lib64/libcrypto.so.1.0.1' }
+
```
tniek: nit: remove blank line

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
+---
+
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States 1
```
tniek: The format of the naming is confusing; not sure what these numbers are

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
...
+  command: "echo never > /sys/kernel/mm/transparent_hugepage/enabled"
+
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States 2
```
tniek: See above

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
...
+  lineinfile: 
+     dest: /etc/default/grub 
+     regexp: "GRUB_CMDLINE_LINUX=" 
+     line: "GRUB_CMDLINE_LINUX='crashkernel=auto  @ vconsole.keymap=us @kernel_arguments: transparent_hugepage=never intel_idle.max_cstate=1 processor.max_cstate=1'"
+
+- name: HugePages & C-States(RHEL-3.10,3.12, SUSE-2205917) - Disable Transparent Hugepages & Configure Processor C-States 3
```
tniek: See above

deploy/v2/ansible/roles/OS_lib_install/tasks/common/os_config.yml
```diff
@@ -0,0 +1,28 @@
...
+  command: grub2-mkconfig -o /boot/grub2/grub.cfg
+
+- name: SWAP space
```
tniek: "Create swap space"

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+
+- name: config common OS libs
```
tniek: "Config"

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  include_tasks: "common/os_config.yml"
+
+- name: config RedHat DB node
```
tniek: "Config"

deploy/v2/ansible/roles/OS_lib_install/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  include_tasks: "{{ ansible_os_family }}/RHEL-{{ ansible_distribution_major_version }}/os_config.yml"
+  when:  ansible_os_family == "RedHat"
+
+- name: config SUSE DB node
```
tniek: "Config"

deploy/v2/ansible/roles/OS_lib_install/tasks/Suse/SUSE-12/os_config.yml
rukawata: I think there is no specific version for package is required.

saddi7087: this is already explained in the last call. version are specified at the OS repo's.

deploy/v2/ansible/roles/OS_lib_install/tasks/RedHat/RHEL-7/os_config.yml
saddi7087: naming issues are addressed.

PR #242, opened 2019-10-17
Add possibility to specify download path as input parameter

rukawata:
Add optional input dir to specify the location to download the bits:

```
>python3 basket.py -h
usage: basket.py [-h] --config CONFIG [--dir DIR] [--basket] [--dryrun]

Downloader

optional arguments:
  -h, --help       show this help message and exit
  --config CONFIG  The configuration file
  --dir DIR        Location to place the download file
  --basket         To include item in the basket, default False
  --dryrun         Set dryrun will not actually download the bits
```

### DIFF
--- a/deploy/v2/python_scripts/downloader/README.MD
+++ b/deploy/v2/python_scripts/downloader/README.MD
@@ -11,6 +11,7 @@ DLM stands for Download Manager
 - add SMP credentials to JSON files `(hdb.json`, `s4.json` are examples for different scenarios).
 - `python3 basket.py --help` : check required inputs
 - `--config <name of the json>` : Mandatory. It triggers download of bits according to SAP_Scenarios.py and JSON input files
+- `--dir <absolute path to download the bits`: Optional. The downloader will use this dir to store downloaded bits instead of default path
 - `--basket`: If set, it triggers download of bits according to SAP_Scenarios.py and JSON input files + the latest files (add in one shot into the basket) from SMP download basket.
 - `--dryrun`: If set, it triggers the download without actually download the bits.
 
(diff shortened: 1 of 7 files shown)


deploy/v2/python_scripts/downloader/README.MD
```diff
@@ -11,6 +11,7 @@ DLM stands for Download Manager
 - add SMP credentials to JSON files `(hdb.json`, `s4.json` are examples for different scenarios).
 - `python3 basket.py --help` : check required inputs
 - `--config <name of the json>` : Mandatory. It triggers download of bits according to SAP_Scenarios.py and JSON input files
+- `--dir <absolute path to download the bits`: Optional. The downloader will use this dir to store downloaded bits instead of default path
```
pabowers: what is the default path that is used?

deploy/v2/python_scripts/downloader/README.MD
```diff
@@ -11,6 +11,7 @@ DLM stands for Download Manager
...
 - `--basket`: If set, it triggers download of bits according to SAP_Scenarios.py and JSON input files + the latest files (add in one shot into the basket) from SMP download basket.
 - `--dryrun`: If set, it triggers the download without actually download the bits.
```
tniek: Your PR description says:
`Dryrun set to True will not actually download the bits`

Which of the two is it? `--dryrun` or `--Dryrun=True`?

deploy/v2/python_scripts/downloader/README.MD
rukawata: > what is the default path that is used?

Default path would be the same folder as the python script.

deploy/v2/python_scripts/downloader/README.MD
rukawata: Should be just add flag Dryrun. Thanks for catching that!

PR #243, opened 2019-10-18
Add conversion from bytes to string

rukawata:
To insure the scripts is compatible with lower python version (eg. 3.4), add explicit conversion for the response content from bytes to string.

### DIFF
--- a/deploy/v2/input.json
+++ b/deploy/v2/input.json
@@ -21,7 +21,7 @@
                         "is_existing" : "",
                         "arm_id" : "",
                         "name" : "",
-                        "allowed_ips" :[] 
+                        "allowed_ips" : ["0.0.0.0/0"] 
                     }
                 }
             },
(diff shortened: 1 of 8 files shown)


lagalbra:
Does the change also need to be applied to the other uses of `json.loads()` in this repo?
- deploy/vm/python-scripts/redeploy-vm.py
- deploy/vm/python-scripts/migrate-disk.py
- monitor/sapmon.py

rukawata:
> Does the change also need to be applied to the other uses of `json.loads()` in this repo?
> 
> * deploy/vm/python-scripts/redeploy-vm.py
> * deploy/vm/python-scripts/migrate-disk.py
> * monitor/sapmon.py

No, this only applies to the downloader which is under deploy/v2/python_scripts.
This change is required since we are parsing the response content from SAP service market place, which is bytes not str.

PR #244, opened 2019-10-18
Copy SSH key pair to all linux jumpboxes with correct permission

rukawata:
Changes in this PR:

1. All linux jumpboxes will now have SSH key pair stored. This way all of them will be able to logon HANA servers.
2. Wide open permission on private key will prevent logon to HANA server with the private key. With the change, we will have `644` for public key and `600` for private key.
3. The determine factor for being RTI should not be the name but key `destroy_after_deploy` which has value true or false.
4. Changed the structure of the input.json so it only has one section with path to ssh key pair. And this ssh key pair will be shared among database servers and jumpboxes if ssh key is used as authentication method.
5. Format all input.json with two spaces indention.


deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -135,26 +135,42 @@ resource "azurerm_virtual_machine" "vm-linux" {
 
   connection {
     type        = "ssh"
-    host        = var.jumpboxes.linux[count.index].name == "rti" ? azurerm_public_ip.public-ip-linux[count.index].ip_address : null
```
tniek: Sidenote - I really hope this is not how we distinguish an RTI from a regular Linux jumpbox; didn't we introduce a `destroy_after_deploy` parameter?

rukawata: Yes, that line has been removed and new code in this PR looks like this:
`host        = var.jumpboxes.linux[count.index].destroy_after_deploy == "true" ? azurerm_public_ip.public-ip-linux[count.index].ip_address : null`

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -135,26 +135,42 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
     user        = var.jumpboxes.linux[count.index].authentication.username
     private_key = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_private_key", false) == false ? null : file(var.jumpboxes.linux[count.index].authentication.path_to_private_key)
     password    = lookup(var.jumpboxes.linux[count.index].authentication, "password", null)
   }
 
+  # Copy ssh keypair over to jumpboxes and set permission
   provisioner "file" {
     source      = lookup(var.jumpboxes.linux[count.index].authentication, "path_to_public_key", null)
     destination = "/home/${var.jumpboxes.linux[count.index].authentication.username}/.ssh/id_rsa.pub"
-    on_failure  = "continue"
```
suhanselvan: Removing the `on_failure = "continue"` would halt the Terraform apply process if a linux jumpbox uses password for authentication.  One suggestion would be to not have the path for the keys specified for each jumpbox or database inside its own block in input.json. Instead we can have a separate section in input.json for specifying the SSH key paths. All the VMs that use key based authentication can refer to this path. This is assuming they all use the same SSH key pair.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -135,26 +135,42 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+    inline = [
+      "chmod 644 /home/${var.jumpboxes.linux[count.index].authentication.username}/.ssh/id_rsa.pub",
+      "chmod 600 /home/${var.jumpboxes.linux[count.index].authentication.username}/.ssh/id_rsa",
+    ]
+  }
+
+  # Copy output.json for ansbile only on RTI. 
+  # Use on_failure = "continue" current workaround to use conditional provisioner
   provisioner "file" {
-    source      = var.output-json.filename
```
suhanselvan: As I said during the call, specifying `var.output-json.filename` as source was to create an implicit dependency on the output_json module. Because during some tests the output.json file was not created before this provisioner started and led to failure.

rukawata: Ack. Revert this part.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: For now I will just copy the key pairs from the first database as a temporary workaround.
In the situation when we have multiple databases which uses different key pairs, we need to find a better solution than this.

rukawata: @suhanselvan after a discussion, we will go with the approach you proposed. Also it makes it hard to manage if they use different keys for databases (eg. how do we name them on the jumpboxes?). 
I will implement the new structure in this PR.

PR #245, opened 2019-10-23
v2: Create/copy Ansible inventory file and install required packages on RTI

suhanselvan:
This PR includes the following additions/changes:

->Creates the Ansible inventory file with host details and copies it to RTI.

->Installs Git, Ansible and clones the project repository on the RTI.

->Output JSON now includes the public IP information of the jumpboxes.

->Fixed some inconsistent variable names.

-> Changed hdb sizing JSON to default T-shirt sizes.


deploy/v2/template_samples/single_node_hana.json
```diff
@@ -78,9 +78,9 @@
         "size": "Standard_D2s_v3",
         "disk_type": "StandardSSD_LRS",
         "os": {
-          "publisher": "Redhat",
-          "offer": "RHEL-SAP",
-          "sku": "7.5"
+          "publisher": "Canonical",
```
rukawata: Any specific reason why we change the OS for linux jumpbox?

suhanselvan: It was changed to be consistent with the values in the latest template.json file.

suhanselvan: Can the single_node_hana.json file be removed altogether since its redundant?

rukawata: No. As we go alone, we may have sample also for other scenarios.

deploy/v2/hdb_sizes.json
```diff
@@ -1,8 +1,8 @@
 {
     "S" : {
```
rukawata: 1. In order to work with v2 pipeline test, please add demo size as below.
2. In [template](https://github.com/Azure/sap-hana/blob/master/templates/tempGen/template.json), set database size to `Demo`.

```JSON
 "Demo": {
    "compute": {
      "vm_size": "Standard_D8s_v3",
      "swap_size_gb": 2
    },
    "storage": {
      "os": {
        "disk_type": "StandardSSD_LRS",
        "disk_size_gb": 64,
        "caching": "ReadWrite"
      },
      "data": {
        "count": 1,
        "disk_type": "Premium_LRS",
        "disk_size_gb": 512,
        "caching": "ReadOnly",
        "write_accelerator": false
      },
      "log": {
        "count": 1,
        "disk_type": "Premium_LRS",
        "disk_size_gb": 512,
        "caching": "ReadOnly",
        "write_accelerator": false
      },
      "shared": {
        "count": 1,
        "disk_type": "StandardSSD_LRS",
        "disk_size_gb": 512,
        "caching": "ReadOnly",
        "write_accelerator": false
      },
      "sap": {
        "count": 1,
        "disk_type": "StandardSSD_LRS",
        "disk_size_gb": 64,
        "caching": "ReadOnly",
        "write_accelerator": false
      },
      "backup": {
        "count": 1,
        "disk_type": "StandardSSD_LRS",
        "disk_size_gb": 512,
        "caching": "ReadOnly",
        "write_accelerator": false
      }
    }
  },
```

deploy/v2/hdb_sizes.json
```diff
@@ -13,45 +13,225 @@
                 "caching" : "ReadWrite"
             },
             {
-                "name" : "data",
-                "count" : 1,
+                "name" : "data_log",
```
rukawata: @suhanselvan please check with @saddi7087 about merge data log together or not. This will impact his Ansible code.

deploy/v2/hdb_sizes.json
suhanselvan: I see that the tests are working fine with the current size "S". Do you still want to add this demo size?

deploy/v2/hdb_sizes.json
suhanselvan: The data was updated based on the latest sizing JSON provided on 10/01. If you could provide me the latest one with separate log and data disks, I can change it.

deploy/v2/hdb_sizes.json
rukawata: Yes, please.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -177,6 +177,42 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+      private_key = var.jumpboxes.linux[count.index].authentication.type == "key" ? file(var.sshkey.path_to_private_key) : null
+      password    = lookup(var.jumpboxes.linux[count.index].authentication, "password", null)
+    }
+
+    source      = var.ansible-inventory.filename
+    destination = "/home/${var.jumpboxes.linux[count.index].authentication.username}/hosts"
+    on_failure  = "continue"
+  }
+
+  # Installs Git, Ansible and clones repository on RTI
```
rukawata: Now we have several files copied over to RTI.
Therefore we have multiple failed attempts when it is not RTI (host: null).
Let's put all files for RTI into a separate folder `ansible_config_files` under `v2`, then just initialize ssh connection once to copy those files over.

suhanselvan: Sure. Will do.

deploy/v2/hdb_sizes.json
suhanselvan: Added.

deploy/v2/hdb_sizes.json
suhanselvan: Updated the JSON.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: This has been implemented. I've also added a `depends_on` argument for the linux jumpbox resource block to ensure the files are created in the `ansible_config_files` folder before the file provisioner  starts.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -172,11 +173,32 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+    connection {
+      type        = "ssh"
+      host        = var.jumpboxes.linux[count.index].destroy_after_deploy == "true" ? azurerm_public_ip.public-ip-linux[count.index].ip_address : null
+      user        = var.jumpboxes.linux[count.index].authentication.username
+      private_key = var.jumpboxes.linux[count.index].authentication.type == "key" ? file(var.sshkey.path_to_private_key) : null
+      password    = lookup(var.jumpboxes.linux[count.index].authentication, "password", null)
+    }
+
+    inline = [
+      "sudo apt update",
```
persiaAziz: Can we organize the installation of the packages in a better way rather than listing everything inline? Also please pin the version of ansible and other packages.

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -172,11 +173,32 @@ resource "azurerm_virtual_machine" "vm-linux" {
...
+      "sudo apt-get install git",
+      "sudo apt install software-properties-common",
+      "sudo apt-add-repository --yes --update ppa:ansible/ansible",
+      "sudo apt -y install ansible",
+      "git clone https://github.com/Azure/sap-hana.git"
+    ]
+    on_failure = "continue"
```
persiaAziz: why continue on failure?

deploy/v2/terraform/modules/output_files/main.tf
```diff
@@ -60,5 +62,11 @@ resource "local_file" "output-json" {
     }
     }
   )
-  filename = "${path.root}/../output.json"
+  filename = "${path.root}/../ansible_config_files/output.json"
+}
+
+# Generates the Ansible Inventory file
+resource "local_file" "ansible-inventory" {
+  content  = templatefile("${path.module}/ansible_inventory.tmpl", { jumpboxes-windows = var.jumpboxes.windows, jumpboxes-linux = var.jumpboxes.linux, ips-jumpboxes-windows = local.ips-jumpboxes-windows, ips-jumpboxes-linux = local.ips-jumpboxes-linux, ips-dbnodes-admin = local.ips-dbnodes-admin, ips-dbnodes-db = local.ips-dbnodes-db, dbnodes = local.dbnodes })
```
persiaAziz: is it possible to make this more readable ?.. something like
content  = templatefile("${path.module}/ansible_inventory.tmpl",
                                           { jumpboxes-windows = var.jumpboxes.windows,
                                            jumpboxes-linux =…..})

deploy/v2/hdb_sizes.json
```diff
@@ -20,6 +20,61 @@
...
                 "write_accelerator" : false
             },
+            {
+                "name" : "log",
+                "count" : 1,
+                "disk_type" : "Premium_LRS",
+                "size_gb" : 512,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false
+            },
```
tniek: Per my comment in PR #238, we should add the corresponding mount point for each shared folder.

deploy/v2/hdb_sizes.json
```diff
@@ -30,28 +85,248 @@
...
+                "name" : "shared",
+                "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
                 "size_gb" : 512,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
             {
-                "name" : "logbackup",
+                "name" : "sap",
```
tniek: Why is log backup now renamed into sap?

deploy/v2/hdb_sizes.json
```diff
@@ -30,28 +85,248 @@
...
                 "count" : 1,
                 "disk_type" : "StandardSSD_LRS",
                 "size_gb" : 64,
                 "caching" : "ReadOnly",
                 "write_accelerator" : false
             },
             {
-                "name" : "databackup",
+                "name" : "backup",
```
tniek: Why is databackup now renamed into backup?

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: It was done inline based on @rukawata 's suggestion. I've added comments for better clarity. The packages have also been pinned to a specific version.

deploy/v2/terraform/modules/jumpbox/main.tf
suhanselvan: Because we want the packages to be installed only on RTI and the provisioner will fail if its not RTI.

deploy/v2/terraform/modules/output_files/main.tf
suhanselvan: Changed the formatting.

deploy/v2/hdb_sizes.json
suhanselvan: Changed based on latest sizing JSON provided.

deploy/v2/hdb_sizes.json
suhanselvan: Changed based on latest sizing JSON provided.

deploy/v2/hdb_sizes.json
suhanselvan: Added the mount point info to sizing JSON.

deploy/v2/terraform/modules/jumpbox/main.tf
persiaAziz: provisioner will fail if its not RTI --- relying on failure sounds like a bad design. What is stopping us from calling this particular  provisioner inside the RTI resource?

deploy/v2/hdb_sizes.json
tniek: Which JSON? Who provided it and when?
Still trying to understand the motivation of renaming the mount points; this doesn't appear to be in-line with [Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/hana-backup-restore) (current as of 10/15/19).

suhanselvan: @rukawata Can you please keep him in loop?

rukawata: @tniek this is modified based on the table for different sizes from Marshal (just fwd you the email thread). Thanks!

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: My latest [commit ](https://github.com/Azure/sap-hana/pull/245/commits/60f7ff4a13dca3d993a2ad4c59368925803e2608) should address this concern :)

With that, we would first identify the RTI by it's tag, then using null provisioner to perform remote execution only on RTI.

suhanselvan: @rukawata Actually now that we have the output JSON, I had created a less complex solution by distinguishing the RTI box and getting the Public IP information from that. Shall I push that or keep the tag based solution you have?

deploy/v2/hdb_sizes.json
```diff
@@ -18,39 +18,344 @@
...
+                "size_gb" : 64,
+                "caching" : "ReadWrite"
+            },
+            {
+                "name" : "data",
+                "count" : 4,
+                "disk_type" : "Premium_LRS",
+                "size_gb" : 1024,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false,
```
pabowers: Is there a particular reason we aren't leveraging the write accelerator on the M-series VMs? https://docs.microsoft.com/en-us/azure/virtual-machines/windows/how-to-enable-write-accelerator

deploy/v2/hdb_sizes.json
```diff
@@ -18,39 +18,344 @@
...
+                "size_gb" : 128,
+                "caching" : "ReadWrite"
+            },
+            {
+                "name" : "data",
+                "count" : 5,
+                "disk_type" : "Premium_LRS",
+                "size_gb" : 1024,
+                "caching" : "ReadOnly",
+                "write_accelerator" : false,
```
pabowers: Same here. Why aren't we using the write accelerator?

deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -94,6 +98,7 @@ resource "azurerm_network_interface" "nic-linux" {
 
 # Creates Linux VM
 resource "azurerm_virtual_machine" "vm-linux" {
+  #depends_on                    = [var.output-json, var.ansible-inventory]
```
pabowers: please remove commented out code

deploy/v2/terraform/modules/output_files/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+variable "jumpboxes" {
+  description = "Details of the jumpboxes"
+}
+
+variable "databases" {
+  description = "Details of the HANA database nodes"
+}
+
+variable "software" {
+  description = "Details of the infrastructure components required for SAP installation"
```
pabowers: the description makes this seem more like hardware than "software" as the name indicates

deploy/v2/terraform/modules/output_files/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+}
+
+variable "nics-jumpboxes-linux" {
+  description = "Details of the Linux jumpbox NICs"
```
pabowers: What are the "details"?

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: @suhanselvan it depends on the solution. Ideally we want to keep the staff related to jumpbox within jumpbox as much as possible, rather than relying on a different module eg. output module. Can you elaborate your solution with more details?

suhanselvan: By using a `for_each` loop inside the null resource `prepare-rti` that loops through the linux jumpboxes in output JSON that have `destroy_after_deploy` set to `true`. This would return only the RTI. And then the `remote-exec` will be executed only on the RTI. The output JSON has the public IP information as well unlike the input JSON.

deploy/v2/hdb_sizes.json
rukawata: Thanks for pointing that out! Apparently write_accelerator only should apply to `/hana/log`.
I adjust the `hdb_size.json` accoding to [this document](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/hana-vm-operations-storage)

deploy/v2/hdb_sizes.json
rukawata: See above comments.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: Removed.

deploy/v2/terraform/modules/output_files/variables.tf
rukawata: "software" is to store information to be able to download SAP media.
Changed accordingly.

deploy/v2/terraform/modules/output_files/variables.tf
rukawata: Modified. Also some other places where there is no details.

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: @suhanselvan please include the null resource `prepare-rti` into output module with your solution. Let's take a look at the solution. Thanks.

suhanselvan: I have updated the code with new method in the latest commit. let me know if its fine. Thanks.

rukawata: Thanks. The overall solution works. And I'm fine keeping it in jumpbox module since the remote execution has dependency on linux jumpbox.

I made a couple of small changes:

- Add timeout settings back to ssh connection
- Add dependency between the null resource and linux VMs
- Add some comments in the code
- Change the format of the for loop so more readable
- Keep the tag name for jumpboxes - I like that we can see that it is LINUX-JUMPBOX/WINDOWS-JUMPBOX in the tag

PR #248, opened 2019-10-26
Change downloader to take input.json

rukawata:
1. Change downloader to take configuration from input.json.
2. Change folder name from python_scripts to python.


deploy/v2/python/downloader/helper.py
```diff
@@ -77,7 +77,7 @@ def load(filename):
             "Config file %s does not exist" % filename
 
         with open(filename) as f:
-            data = json.load(f)
+            data = json.load(f)["software"]["downloader"]
```
Caiyeon: I don't see a "downloader" field in input.json["software"]. Am I missing something?

Also, above this, we assert that the file exists. And below, we assert that credentials exists. Why not assert here that software and downloader exists?

rukawata: This new field has been added from a different [PR](https://github.com/Azure/sap-hana/blob/3b2281540fd59de89945b6e8af05317912cf8a78/deploy/v2/input.json).
Good point, I will add assert to check if software and downloader exists.

PR #250, opened 2019-10-31
V2: Ansible - OS and Disk configurations

saddi7087:
V2: Ansible - OS and Disk configurations. 

This is the same code as [PR #238]( https://github.com/Azure/sap-hana/pull/238) - we closed that one due to not able to resolve conflicts between that branch and the latest master.

Below code are added to test the Ansible code (since playbook is not yet in master).
@rukawata will create a new PR to remove this after this PR got merged.

```
     ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" << EOF
        cd sap-hana
        git checkout wft-ansible-os
      EOF
```

PR #251, opened 2019-11-01
Extend v2 pipeline to test different OS versions

rukawata:
To be able to run the test cases for different OS versions, below changes are made in this PR:
- Separate jobs for different scenarios as template under `template/job-template-per-os.yaml`
- Use the above template trigger test jobs for two OS versions - can be extended in the future:
   - **SLES12**
   - **RHEL7**
  The above two versions are currently supported by the Ansible [PR](https://github.com/Azure/sap-hana/pull/250).
- Adjust the cleanUp job so it can work with the new naming conversion for created resource group: `"${scenario}-${os}-$(Build.BuildId)"` eg. `v2-reuseNSG-RHEL7-1365`
- Adjust the `README.MD` accordingly.


templates/job-template-per-os.yaml
```diff
@@ -0,0 +1,107 @@
...
+  - script: |
+      az login --service-principal --user $(hana-pipeline-spn-id) --password  $(hana-pipeline-spn-pw) --tenant $(landscape-tenant) --output none
+      echo '##vso[task.setvariable variable=arm_id]$(az network nsg show --resource-group v2-allNew-${{parameters.osVersion}}-$(Build.BuildId) --name nsg-mgmt --query id --output tsv)' 
+  - template: update-place-holder.yaml
+    parameters:
+      scenario: $(scenario)
+      testCaseName: $(testcase)
+      placeHolder: var-rg-name
+      value: $(testcase)
+  - template: update-place-holder.yaml
```
persiaAziz: Can we rename the update-place-holder.yaml to something else in this PR. Specifically get rid of the "place-holder" and name it something like change resource change param or something?

rukawata: Sure! I have some options, let me know which causes less confusion :)

- update-template-json
- update-template-value
- replace-json-value
- ...

azure-pipelines-v2.yaml
```diff
@@ -150,26 +64,37 @@ stages:
 # Tag "Delete=True" will be added to mark all resource groups for   #
 # scheduled deletion done by azure-pipelines-cleanup.yaml           #
 #####################################################################
-- stage: DestroyingResources
+- stage: DestoryingResources
```
pabowers: typo: should be `DestroyingResources`

templates/job-template-per-os.yaml
rukawata: Change it to test-case-template-update.

azure-pipelines-v2.yaml
rukawata: Thanks! Adjusted.

PR #252, opened 2019-11-03
v2:Added resource block to create Azure file share

suhanselvan:
Changes in this PR:

-> Includes resource block to create Azure file share.

Current Terraform azure provider supports file shares under Standard storage account tier only. input file needs to be modified accordingly.


deploy/v2/input.json
```diff
@@ -179,6 +179,7 @@
       "account_replication_type": "",
       "account_kind": "",
       "container_name": "",
+      "share_name": "",
```
persiaAziz: rename to file_share_name

rukawata: Maybe let's change both: `blob_container_name` and `file_share_name`, so it's clearer one is for blob while the other is for azure file :) 
@suhanselvan please adjust accordingly, thanks!

suhanselvan: Changed.

PR #254, opened 2019-11-04
V2: Ansible - Bits download

saddi7087:
Mount the Azure File share and SAP Media download

PR #256, opened 2019-11-04
V2: Ansible  configure linux jump boxes

saddi7087:
Install HANA Studio on Linux Jumpboxes


deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+
+# Mount Azure File and Install Hana Studio on jumpbox
```
rukawata: This should be done together with other mounting in PR #254 (you had it for RTI and HANA servers there, the inventory for mounting should also include all jumpboxes).
Please remove it from PR #256 and add it to PR #254.

deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+
+- name: Mount storage for SAP bits on Jumpbox
+  mount:
+    path: "{{ azure_files_mount_path }}"
+    fstype: cifs
+    opts: "credentials=/etc/smbcredentials/{{ output.software.storage_account.name }}.cred,dir_mode=0755,file_mode=0755,serverino,defaults,nofail,vers=3.0"
+    src: "//{{ output.software.storage_account.name }}.file.core.windows.net/sapmnt"
+    state: mounted
+
+- name: Create hdbstudio install path
```
rukawata: Name shouldn't be hdbstudio install.
Idea is to install all components under component section in the output.json.

deploy/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,57 @@
...
+    - { role: os-disk-setup }
+    - { role: sap-media-download }
+    - { role: sap-hdb-xsa-install }
+
+- hosts: linux-jumpboxes
+  become: true
+  become_user: root
+  vars:
+    azure_files_mount_path: "/sapmnt"
+    hdbstudio_install_path: "/usr/sap/hanastudio"
```
rukawata: This should not be hard coded.
In the future we may consider the possibility to enable customer install under customized path, but for now, always install under default path.

deploy/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,57 @@
...
+    hdbstudio_software_loc: "{{ azure_files_mount_path }}/DB/DATA_UNITS/HDB_STUDIO_LINUX_X86_64"
+    hdbstudio_template_path: "/usr/sap"
```
rukawata: Why template need to be under /usr/sap?
Also we should remove the template after the installation is done.

deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
saddi7087: We cannot combine Jumpbox with RTI and DB servers for Azure file mounting Because ansible accesses to jumpbox is password and DB is ssh key. if we change the access part we can use same code for both.

deploy/ansible/sap_playbook.yml
saddi7087: If template is the install path, installer will fail throwing install path is not empty. I have keep the template file in some location other than installation location. Please suggest a location to keep template. yes, we can delete the template.

deploy/ansible/sap_playbook.yml
```diff
@@ -0,0 +1,57 @@
...
+        hana_database: "{{ item }}"
+      loop: "{{ output.databases|flatten(levels=1) }}"
+      when: item.platform == 'HANA'
+  roles:
+    - { role: os-preparation }
+    - { role: os-disk-setup }
+    - { role: sap-media-download }
+    - { role: sap-hdb-xsa-install }
+
+- hosts: linux-jumpboxes
```
rukawata: Accoding to Suhan's code, inventory for linux jumpboxes should be `jumpboxes_linux`

deploy/ansible/sap_playbook.yml
rukawata: It should work. Please check [this role](https://github.com/Azure/sap-hana/blob/master/deploy/vm/ansible/roles/saphana-install/tasks/main.yml):

```yaml
- name: deploy hdblcm install template
  template:
    src: "{{ 'hdbserver_hana2.j2' if use_hana2 else 'hdbserver_hana1.j2'}}"
    dest: /hana/shared/install/hdbserver_{{ sap_sid }}_install.cfg
```

saddi7087: This won't work because we are installing in different directory called  /hana/shared/install/SAP_HANA_DATABASE, so template is copied to install directory. 

- name: run hdblcm
  shell: "pwd=$(<../hdbserver_{{ sap_sid }}_passwords.xml); rm ../hdbserver_{{ sap_sid }}_passwords.xml; echo $pwd | ./hdblcm --batch --action=install --configfile='../hdbserver_{{ sap_sid }}_install.cfg' --read_password_from_stdin=xml"
  args:
    chdir: /hana/shared/install/SAP_HANA_DATABASE

deploy/ansible/sap_playbook.yml
saddi7087: This name taken from hosts file created by terraform not the output.json. If we have final version of output.json and hosts file, I will make changes accordingly.

deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
rukawata: We will not change the two options for authentication for all VMs. This is always the case here to have mixed usage of ssh key pair and password, especially for windows jumpboxes, we always use password. `ansible_password` needs to be set in the inventory file - @suhanselvan if this is not done, please add this.

deploy/ansible/sap_playbook.yml
rukawata: Currently Suhan already set it to jumpboxes_linux: check this [code on master](https://github.com/Azure/sap-hana/blob/master/deploy/v2/terraform/modules/output_files/ansible_inventory.tmpl).

saddi7087: I will modify the name accordingly.

deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
saddi7087: I will modify the code to mount azure file share on linux servers using the os-independent group.

deploy/ansible/sap_playbook.yml
saddi7087: modified the hosts

deploy/ansible/roles/hana-studio-linux-install/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+---
```
rukawata: All the paths are wrong. Should be under `deploy/v2/ansible`.

PR #261, opened 2019-11-12
Add option to execute Ansible playbook from Terraform

rukawata:
Introduced flag `ansible_execution`:

- If set to `true`, terraform apply will execute Ansible playbook from jumpbox
- If set to `false`, terraform will not execute Ansible playbook, only show the Ansible version on jumpbox

In the v2 pipeline testing, this flag is set to `false`, because we want to checkout the updated playbook from PR and execute that one instead.

### DIFF
--- a/deploy/v2/input.json
+++ b/deploy/v2/input.json
@@ -129,8 +129,8 @@
       "high_availability": "",
       "authentication": {
         "type": "",
-        "username": "",
-        "password": ""
+        "os_username": "",
+        "os_password": ""
       },
       "instance": {
         "sid": "",
--- a/deploy/v2/input.json
+++ b/deploy/v2/input.json
@@ -64,6 +64,7 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
+        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
@@ -83,6 +84,7 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
+        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
@@ -100,6 +102,7 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
+        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
--- a/deploy/v2/input.json
+++ b/deploy/v2/input.json
@@ -64,7 +64,6 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
-        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
@@ -84,7 +83,6 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
-        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
@@ -102,7 +100,6 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
-        "nic_name": "",
         "private_ip_address": "",
         "os": {
           "publisher": "",
@@ -132,8 +129,8 @@
       "high_availability": "",
       "authentication": {
         "type": "",
-        "os_username": "",
-        "os_password": ""
+        "username": "",
+        "password": ""
       },
       "instance": {
         "sid": "",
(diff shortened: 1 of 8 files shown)


deploy/v2/terraform/variables.tf
```diff
@@ -21,3 +21,8 @@ variable "sshkey" {
 variable "options" {
   description = "Configuration options"
 }
+
+variable "ssh-timeout" {
+  description = "Timeout for connection that used by provisioner"
```
lagalbra: nit: "Timeout for connection used by provisioner" or "Timeout for connection that **is** used by provisioner"

lagalbra: (and same rename suggestion for anywhere else this is used, such as deploy/v2/terraform/modules/jumpbox/variables.tf line 42)

deploy/v2/input.json
```diff
@@ -132,8 +129,8 @@
       "high_availability": "",
       "authentication": {
         "type": "",
-        "os_username": "",
-        "os_password": ""
+        "username": "",
+        "password": ""
```
lagalbra: Why are these fields and `nic_name` being changed with this particular PR? Do they affect the Ansible/Terraform connection?

deploy/v2/terraform/variables.tf
rukawata: Thanks. Addressed in both :)

deploy/v2/input.json
rukawata: Reversed.

lagalbra: What about `nic_name`?

deploy/v2/input.json
```diff
@@ -64,7 +64,6 @@
         "name": "",
         "destroy_after_deploy": "",
         "size": "",
-        "nic_name": "",
```
tniek: How is the `nic_name` related to this PR?

deploy/v2/terraform/main.tf
```diff
@@ -56,3 +57,24 @@ module "output_files" {
...
 }
+
+resource "null_resource" "ansible_playbook" {
+  depends_on = [module.hdb_node.dbnodes, module.jumpbox.prepare-rti]
+  connection {
+    type        = "ssh"
+    host        = module.jumpbox.rti-info.public_ip_address
+    user        = module.jumpbox.rti-info.authentication.username
+    private_key = module.jumpbox.rti-info.authentication.type == "key" ? file(var.sshkey.path_to_private_key) : null
+    password    = lookup(module.jumpbox.rti-info.authentication, "password", null)
```
tniek: So if `key` or `password` are not supplied, it is using a `null` value for location of SSH private key or password, respectively?
Would it not be better to fail right away?

rukawata: We do not have a way to verify and validate all inputs before we trigger the terraform deployment at the moment.
This may be possible when we integrate terraform with for example az extension or portal.

deploy/v2/input.json
rukawata: My bad, Laura pointed out another field that got into this PR.
I'm in parallel cleaning up the input json. Will add it back in this PR and clean them up in a separate PR.

deploy/v2/input.json
rukawata: Removed as well.

deploy/v2/terraform/main.tf
```diff
@@ -56,3 +57,24 @@ module "output_files" {
...
+    timeout     = var.ssh-timeout
+  }
+
+  # Run Ansible Playbook on jumpbox if ansible_execution set to true
+  provisioner "remote-exec" {
+    inline = [
+      "export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES",
+      "export ANSIBLE_HOST_KEY_CHECKING=False",
+      var.options.ansible_execution ? "ansible-playbook -i hosts ~/sap-hana/deploy/v2/ansible/sap_playbook.yml" : "ansible-playbook --version"
```
persiaAziz: Apologies if I am missing something. Where/how is the playbook getting copied again?

rukawata: The playbook was copied onto RTI in the jumpbox [module](https://github.com/Azure/sap-hana/blob/master/deploy/v2/terraform/modules/jumpbox/main.tf).
I trigger the ansible playbook in the main module instead of the jumpbox module is because it has dependency on both jumpbox and hana servers (hdbnodes).

PR #268, opened 2019-11-18
Track terraform deployment with IMDS

rukawata:
Since the Ansible playbook is not final, I only add tracking of deployemnt for Terraform execution.
Below filter can be used to check the statistic<sup>[1](#myfootnote1)</sup>:

```sql
| where TimeStamp > ago(3d)
| where UserAgent contains"SAP AutoDeploy/v2" and message contains"Terraform_finished"
```

<a name="myfootnote1">1</a>:There is normally a couple of hours delay until the data show up.


lagalbra:
whoooooaaa https://www.markdownguide.org/extended-syntax/#footnotes

PR #274, opened 2020-02-04
Reduce unnecessary jobs from pipeline

rukawata:
The total number of jobs in pipeline for each run is:

[number of resource deployment options<sup>[1](#myfootnote1)</sup>] * [number of supported OS versions<sup>[2](#myfootnote2)</sup>]
(eg. 4 testcases * 2 supported OS versions = 8 jobs)

However I do not see that it's necessary to have all combination and has below disadvantages:

- Increase the total time for each run
- Add more cost to the subscription

Therefore I reduce the number of jobs to:

[number of resource deployment options] + [number of supported OS version] - 1
(eg. 4 testcases + 2 supported OS versions - 1 = 5 jobs)

<a name="myfootnote1">1</a>: resource deployment options:
- createAllNew
- reuseRG
- reuseVnet
- reuseNSG

<a name="myfootnote2">2</a>: supported OS versions
- SLES12
- RHEL7


azure-pipelines-v2.yaml
```diff
@@ -35,16 +35,19 @@ stages:
         sudo -H pip install msrest==0.6.0
       displayName: 'Install python tools'
 #####################################################################
-# This stage create resources for below test cases:                 # 
-# - createAllNew: it creates all resources from scratch             # 
-# - reuseRG: it only reuses an existing resource group              #
-# - reuseVnet: it only reuses an existing vnet from createAllNew    #
-# - reuseNSG: it only reuses an existing NSG from createAllNew      #
+# This stage create resources for below test cases:                 #
+# - Below testcase are tested only with SLES12                      #
```
pabowers: typo: should be "test cases"

rukawata: Corrected :)

persiaAziz:
I am seeing this pipeline fails due the download of hdbserver needing a lot of retries and eventually timing out. I wonder what the issue is.

rukawata:
> I have been seeing since last week that this pipeline fails due the download of hdbserver needing a lot of retries and eventually timing out. I wonder what the issue is.

Me too. It's not related to v2 though. 
What we can do is to move the bits to a storage in a different zone - will take a look at that separately.

rukawata:
> I have been seeing since last week that this pipeline fails due the download of hdbserver needing a lot of retries and eventually timing out. I wonder what the issue is.

The deployment for v1 is in [eastus](https://github.com/Azure/sap-hana/blob/master/deploy/vm/test/terraform.tfvars).
The storage account we use is in westus2.

I would move all binaries to a sa that is in same region as deployment (eastus).
Let's see if that would fix the download failure.

rukawata:
> > I have been seeing since last week that this pipeline fails due the download of hdbserver needing a lot of retries and eventually timing out. I wonder what the issue is.
> 
> The deployment for v1 is in [eastus](https://github.com/Azure/sap-hana/blob/master/deploy/vm/test/terraform.tfvars).
> The storage account we use is in westus2.
> 
> I would move all binaries to a sa that is in same region as deployment (eastus).
> Let's see if that would fix the download failure.

@persiaAziz, after changing to use sa in eastus, the Azure.sap-hana pipeline (for v1) finished with 32m37s. Comparing to previous runs that had different and longer runtime (eg. 51m21s), this seems to improve the pipeline. Anyway, this is not related to the change in this PR, just make the pipeline that this PR has to go through more stable.

PR #280, opened 2020-02-08
Use Python 3 interpreter for ansible

rukawata:
Goal is to move away from python2 and use python3.
Ansible supports python3. To have that set as default, pip3 need be used for the installation.

Therefore upgrade component install to pip3.

### DIFF
--- a/deploy/v2/terraform/modules/jumpbox/main.tf
+++ b/deploy/v2/terraform/modules/jumpbox/main.tf
@@ -285,9 +285,9 @@ resource "null_resource" "prepare-rti" {
       # Install pip3
       "sudo apt -y install python3-pip",
       # Installs Ansible
-      "sudo apt install software-properties-common",
-      "sudo apt-add-repository --yes --update ppa:ansible/ansible-2.8",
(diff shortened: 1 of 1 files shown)


deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -285,9 +285,9 @@ resource "null_resource" "prepare-rti" {
       # Install pip3
       "sudo apt -y install python3-pip",
       # Installs Ansible
-      "sudo apt install software-properties-common",
-      "sudo apt-add-repository --yes --update ppa:ansible/ansible-2.8",
-      "sudo apt -y install ansible=2.8*",
+      "sudo -H pip3 install \"ansible>=2.8,<2.9\"",
```
tniek: See my comment in PR #281 - why ansible < 2.9?

rukawata:
> Why is this addressed again in PR #281 ?

Because PR #281 is address a different problem than python3.
However PR #281 can't be tested without the python3 addressed in PR #280.

Sorry for the confusion.

PR #281, opened 2020-02-08
Fix problem with winrm

rukawata:
There are some issues with the previous winrm configuration, that makes the ansible not able to run playbook on windows jumpboxes, especially with the problem below:

_Not able to logon and run playbook on windows jumpbox during the 1st terraform+ansible deployment - we see ansible hanging at [**Gathering facts**] phase. However the 2nd attempt of terraform + ansible would work._

This is the fix based on [the PR for python3](https://github.com/Azure/sap-hana/pull/280), and have tested with [sample playbook](https://github.com/Azure/sap-hana/tree/nancyc-winrm-fix-test) that it works from end to end for windows jumpbox.

### DIFF
--- a/deploy/v2/terraform/modules/jumpbox/main.tf
+++ b/deploy/v2/terraform/modules/jumpbox/main.tf
@@ -229,12 +229,21 @@ resource "azurerm_virtual_machine" "vm-windows" {
   os_profile_windows_config {
     provision_vm_agent = true
 
+    winrm {
+      protocol = "Http"
+    }
+
+    winrm {
+      protocol        = "Https"
+      certificate_url = azurerm_key_vault_certificate.key-vault-cert[count.index].secret_id
+    }
+
     # Auto-Login's required to configure WinRM
     additional_unattend_config {
       pass         = "oobeSystem"
       component    = "Microsoft-Windows-Shell-Setup"
       setting_name = "AutoLogon"
-      content      = "<AutoLogon><Password><Value>${var.jumpboxes.windows[count.index].authentication.password}</Value></Password><Enabled>true</Enabled><LogonCount>1</LogonCount><Username>${var.jumpboxes.windows[count.index].authentication.username}</Username></AutoLogon>"
+      content      = "<AutoLogon><Password><Value>${var.jumpboxes.windows[count.index].authentication.password}</Value></Password><Enabled>true</Enabled><LogonCount>2</LogonCount><Username>${var.jumpboxes.windows[count.index].authentication.username}</Username></AutoLogon>"
     }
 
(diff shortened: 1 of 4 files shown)


deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -285,9 +294,9 @@ resource "null_resource" "prepare-rti" {
       # Install pip3
       "sudo apt -y install python3-pip",
       # Installs Ansible
-      "sudo apt install software-properties-common",
-      "sudo apt-add-repository --yes --update ppa:ansible/ansible-2.8",
-      "sudo apt -y install ansible=2.8*",
+      "sudo -H pip3 install \"ansible>=2.8,<2.9\"",
```
tniek: Why ansible < 2.9?

deploy/v2/terraform/modules/jumpbox/outputs.tf
```diff
@@ -22,3 +22,7 @@ output "rti-info" {
 output "prepare-rti" {
   value = null_resource.prepare-rti
 }
+
+output "vm-windows" {
+  value = azurerm_virtual_machine.vm-windows
+}
```
tniek: Probably needs to be addressed in a separate PR, but not happy about the inconsistent naming (`prepare-rti` vs. `vm-windows`)

deploy/v2/terraform/modules/jumpbox/main.tf
rukawata: We want to fix using ansible larger release 2.8.
This was decided in https://github.com/Azure/sap-hana/pull/266.
If we would move to 2.9.* at certain point, we could change it by then.

rukawata: > Why ansible < 2.9?

We want to fix using ansible larger release 2.8.
This was decided in #266.
If we would move to 2.9.* at certain point, we could change it by then.

deploy/v2/terraform/modules/jumpbox/outputs.tf
rukawata: > Probably needs to be addressed in a separate PR, but not happy about the inconsistent naming (`prepare-rti` vs. `vm-windows`)

No. This is an workaround (same as `output "prepare-rti"`) and is required to make sure windows jumpboxes is ready before ansible-playbook runs.

rukawata: > Probably needs to be addressed in a separate PR, but not happy about the inconsistent naming (`prepare-rti` vs. `vm-windows`)

For the inconsistent naming, I will address that in a different PR since the names were already defined in the previous PRs.

PR #282, opened 2020-02-11
Add ansible code to install components on linux jumpboxes

rukawata:
Add roles to install HANA studio and HANA client on all Linux jumpboxes except RTI.

### DIFF
--- a/deploy/v2/ansible/group_vars/all.yml
+++ b/deploy/v2/ansible/group_vars/all.yml
@@ -3,12 +3,14 @@ disk_dict: {}
 start: 0
 azure_files_mount_path: "/sapmnt"
 components_list: []
+linux_comp_list: []
 hana_install_path: "/hana/shared/install"
 hana_software_loc: "{{ azure_files_mount_path }}/DB/DATA_UNITS"
 components: "{{ lookup('file', '../components.json') }}"
 hana_client_install_path: "/usr/sap/shared"
 output: "{{ hostvars['localhost'].output }}"
 hana_database: "{{ hostvars['localhost'].hana_database }}"
+linux_jumpboxes: "{{ hostvars['localhost'].jumpboxes.linux"
 sap_media_download_script_loc: "/home/{{ansible_user}}/sap-hana/deploy/v2"
 jumpbox_comp_dict: {}
 media_copy_loc: C:\Users\{{ ansible_user }}\install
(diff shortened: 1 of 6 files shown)


deploy/v2/ansible/group_vars/all.yml
```diff
@@ -3,12 +3,14 @@ disk_dict: {}
 start: 0
 azure_files_mount_path: "/sapmnt"
 components_list: []
+linux_comp_list: []
```
pabowers: what's the difference between `linux_comp_list` and `components_list`?

deploy/v2/ansible/roles/components-install/tasks/hana_client_linux.yml
```diff
@@ -1,13 +1,12 @@
...
 
-# Install HANA Client 
 - name: Create dictionary with installation variant for HANA Client 
   set_fact:
     client: "{{ item }}"
   loop: "{{ components }}"
   when: item.component == 'hana_client'
 
-- name: Install Hana Client
+- name: Install Hana Client on Linux
```
pabowers: I assume this is the linux jump host. Can the name please reflect that?

deploy/v2/ansible/roles/linux-jumpboxes-install/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+# Collect the component list from all linux jumpboxes except rti
```
pabowers: what is rti?

deploy/v2/ansible/group_vars/all.yml
rukawata: > what's the difference between `linux_comp_list` and `components_list`?

`components_list` is the list of components to be installed on HANA servers.
`linux_comp_list` is the list of components to be installed on Linux jumpboxes.

We could change components_list to hdb_comp_list, but I would prefer to do that in a separate PR.

deploy/v2/ansible/roles/linux-jumpboxes-install/tasks/main.yml
rukawata: > what is rti?

rti (runtime instance) is basically a bastion host that we use to run ansible-playbook.
We plan to have it destroyed (Eg. with terraform destroy) after the deployment finishes.

deploy/v2/ansible/roles/components-install/tasks/hana_client_linux.yml
rukawata: > I assume this is the linux jump host. Can the name please reflect that?

Sure. Done :)

deploy/v2/ansible/group_vars/all.yml
pabowers: sure, as long as it will be more clear in the future

PR #286, opened 2020-02-15
Add code to enable prometheus on HANA VMs

rukawata:
With this PR, we will be able to enable prometheus on HANA VMs if `"enable_prometheus": true` in `input.json`, in `options`. Related changes:
- `golang-github-prometheus-node_exporter` does not exist for `SLES12 SP2`, therefore we upgrade the pipeline testing to use a newer version `SLES12 SP4` (same code tested with `SLES12 SP3`, also works.)
- Pass information under `options` from Terraform to Ansible with `output.json`.
- Add Ansible code to enable prometheus for HANA servers.

### DIFF
--- a/deploy/v2/ansible/sap_playbook.yml
+++ b/deploy/v2/ansible/sap_playbook.yml
@@ -1,5 +1,9 @@
 ---
 
+# TODO: 
+# Remove YAML shorthand style for roles (Eg. - { role: os-preparation }) 
+# And replace with newer syntax on https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html
+
 - hosts: localhost
   become: true
   become_user: root
@@ -29,6 +33,10 @@
   roles:
     - { role: os-preparation }
     - { role: os-disk-setup }
+    - role: enable-prometheus
+      when:
+        - output.options.enable_prometheus == True
+        - ansible_facts['distribution_file_variety'] == 'SUSE'
 
 # Mount Azure File share on all linux servers
 - hosts: all_linux_servers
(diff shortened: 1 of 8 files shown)


persiaAziz:
What happens when the selected OS is SLES12 SP2 and `enable_prometheus` is true? Do we need any validation for that before the actual deployment starts?

rukawata:
> What happens when the selected OS is SLES12 SP2 and `enable_prometheus` is true? Do we need any validation for that before the actual deployment starts?

We can consider to add this to a validation check (maybe even through UI when we start working on that) of the input.JSON in the future.

rukawata:
> What happens when the selected OS is SLES12 SP2 and `enable_prometheus` is true? Do we need any validation for that before the actual deployment starts?

I just added a condition for enable_prometheus [here](https://github.com/Azure/sap-hana/blob/2cabb692d7769759f1604459f198e48113155b61/deploy/v2/ansible/sap_playbook.yml#L40).

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -29,6 +33,11 @@
   roles:
     - { role: os-preparation }
     - { role: os-disk-setup }
+    - role: enable-prometheus
+      when:
+        - output.options.enable_prometheus == True
+        - ansible_facts['distribution_file_variety'] == 'SUSE'
+        - ansible_facts['distribution_version'] is regex("(12.[3-9]|15.\d)")
```
tniek: Should be `regex("(12.[3-4]|15.\d)")`

rukawata: Addressed.

PR #288, opened 2020-02-20
Document and Simplify V2 Usage Process

ben-moss:
## Problem

There have been some subtle changes between V1 and V2 of the codebase, and currently no updated usage guide exists for the V2 codebase. There are also some areas of the process which can be challenging or off-putting for newcomers to the world of infrastructure managed by code - particularly when it brings together skillsets around SAP, Azure and Automated Provisioning and Configuration Management.

## Description

This PR provides a simplified run through of the codebase's lifecycle, including various prerequisite tasks required to run the deployment code. Some of the areas where gotchas often occur are simplified and made safer with the introduction of some basic bash scripting to assist the user and help them avoid common mistakes. These areas include:

1. Checking local workstation dependencies are installed at supported versions;
1. Checking the Azure CLI is logged in and configured to use the expected subscription;
1. Generating the service principal and configuring tool access via environment variables;
1. Editing config (JSON) files by hand;
1. Running Terraform from the correct directory and with the correct paths to config files.

The usage guide also contains notes on how long steps typically take to run and their cadence (how often they're expected to be run) in relation to the subscription, SAP system build, etc.

## Pre-Requisites

1. A fresh checkout of the codebase (using this PR's branch) to avoid local changes to config taking precedence
   - Check with `git status` and or `git diff`
1. A fresh terminal session and without the Azure/ARM environment variables set
   - Check with `env | sort | grep -E "ARM|AZURE"`
   - Clear existing values using `unset <variable name>`
1. `shellcheck` installed on the workstation you're testing from (for test 3)

## Test Instructions

- [x] 1. Review the [USAGE.md](https://github.com/Centiq/sap-hana/blob/decb3053aaf8c0af46487bc8b8f18f838be39124/deploy/v2/USAGE.md) guide for general readability and to ensure it is clear.
- [x] 2. Follow the usage guide to ensure the process works and the new utility scripts run successfully
- [x] 3. Shellcheck review the scripts by running `shellcheck util/*.sh` (no results should be shown and there should be a zero exit status from `echo $?` when run after shellcheck)
- [x] 4. Check some edge cases for reasonable error handling in the scripts, such as:
   - Not specifying command line arguments, when they're expected
   - Rerunning scripts a second time
   - Being logged out of the Azure CLI when a script is run that expects you to be logged in

### DIFF
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -107,7 +107,7 @@ Before running any of the following code/scripts, you should login to the Azure 
 
 ### Configuring Authorization with Azure
 
-In order for Terraform/Ansible to manage resources in Azure, a Service Principal is required.
+In order for Terraform/Ansible to manage resources in Azure, a _Service Principal_ is required.
 The following process creates a new service principal in Azure, and stores the details required in an authorization script on the local workstation.
 This script can then be used (_sourced_) to configure the required environment variables on the local workstation that allows Terraform/Ansible to run without prompting the user for further authentication information.
 
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -143,7 +143,7 @@ The SAP environments deployed by this codebase are configured by JSON input file
 These configuration files provide a high degree of customization for the user, but can be a little daunting if you are new to the codebase.
 Therefore example configuration files have been supplied with the code.
 
-The minimal amount of change required to an example configuration file, is to configure your SAP Launchpad credentials so that the code can automatically login and download the required SAP packages to install.
+The minimal amount of change required to an example configuration file is to configure your SAP Launchpad credentials so that the code can automatically login and download the required SAP packages to install.
 
 Configuring your SAP Launchpad credentials for the simplest example JSON input file, requires you to provide your SAP user and password to another utility script.
 
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -26,6 +26,8 @@ You also need some familiarity with using the following technologies/tools:
 
 Before continuing you should first obtain a copy of the code, so that you can use the utility scripts provided.
 
+**Note:** Currently, the utility scripts are only available for Linux/MacOS workstations. If you are interested in Windows support, then please upvote (:thumbsup:) the issue [Add utility scripts for Windows](https://github.com/Azure/sap-hana/issues/289).
+
 1. On the Linux command line, navigate to the directory you wish to clone the code within.
    This directory will be the parent directory of the directory containing the code. For example:
 
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -174,12 +174,12 @@ Configuring your SAP Launchpad credentials for the simplest example JSON input f
    **Note:** This process can take in the region of 90 minutes to complete.
    Particularly slow stages are:
 
-     - `Install XSA components` (~35 minutes)
-     - `Extract media archive` (~15 minutes)
-     - `Install SHINE` (~10 minutes)
-     - `Install HANA Database using hdblcm` (~10 minutes)
-     - `Download installation media` (~5 minutes)
      - `Installing OS package` (~5 minutes)
+     - `Download installation media` (~5 minutes)
+     - `Extract media archive` (~15 minutes)
+     - `Install HANA Database using hdblcm` (~10 minutes)
+     - `Install XSA components` (~35 minutes)
+     - `Install SHINE` (~10 minutes)
 
 1. To review/inspect the provisioned resources navigate to the `test_rg` resource group of your configured Azure subscription in Azure portal.
    By default, all the provisioned resources (excluding the service principal) are deployed into the same resource group.
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -1,0 +1,221 @@
+# Usage Guide for V2
+
+This document aims to describe the basic process to get started with V2 of the codebase.
+It does not go into detail about how to customize the code, but instead focusses on demonstrating the end-to-end lifecycle.
+
+## Prerequisites
+
+The code is designed to be used from a Unix-based OS such as MacOS, Linux, Cygwin, or Windows Subsystem for Linux.
+There are several options available:
+
+1. Use your local workstation's OS directly
+1. Use a Linux-based VM running on your local workstation (e.g. VirtualBox/Vagrant)
+1. Use a Linux-based VM running in Azure
+1. Use the Azure Cloud Shell
+
+Depending on the OS option you choose, you may need to install/upgrade certain tools.
+
+You also need some familiarity with using the following technologies/tools:
+
+1. Linux command line
+1. Git (ideally command line)
+1. Azure portal
+1. SAP Launchpad (login credentials required for SAP software downloads)
+
+### Obtaining the Code
+
+Before continuing you should first obtain a copy of the code, so that you can use the utiliy scripts provided.
+
+1. On the Linux command line, navigate to the directory you wish to clone the code within.
+   This directory will be the parent directory of the directory containing the code. For example:
+
+   ```text
+   cd ~/projects/
+   ```
+
+   **Note:** Ensure you choose a directory without any spaces in its absolute path, to avoid potential future issues with tooling that might not handle this setup.
+
+1. Clone this repository from GitHub. For example:
+
+   ```text
+   git clone https://github.com/Azure/sap-hana.git
+   ```
+
+   **Note:** See [Cloning a respository](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/cloning-a-repository) if you are not familiar with this process.
+
+1. Navigate into the project root directory. For example:
+
+   ```text
+   cd sap-hana
+   ```
+
+   **Note:** All of the following process steps should be run from the project root directory.
+
+### Checking Tool Dependencies
+
+Running the code requires the following tools with the minimal supported/tested versions:
+
+| Tool      | Minimum Version Supported / Tested |
+|-----------|------------------------------------|
+| Azure CLI | 2.0.63                             |
+| Terraform | 0.12.12                            |
+
+1. To easily check which tool versions you have installed, run the following utility script:
+
+   ```text
+   util/check_workstation.sh
+   ```
+
+   Example output:
+   ```text
+   azure-cli = 2.0.77
+   Terraform = 0.12.16
+   ansible = 2.8.1
+   ```
+
+   **Note:** Ansible is only a prerequisite of the workstation if you opt to split the Terraform and Ansible stages.
+
+### Configuring the Target Azure Subscription
+
+Before running any of the following code/scripts, you should login to the Azure CLI and always ensure you are configured to work with the correct Azure subscription.
+
+1. To login to the Azure CLI, run the following command and follow the guided login process:
+
+   ```text
+   az login
+   ```
+
+   **Note:** If you have access to multiple subscriptions you may need to use the following type of command to select the desired target subscription:
+
+   ```text
+   az account set --subscription <subscription name or id>
+   ```
+
+1. To easily check which Azure subscription is your current target, run the following utility script:
+
+   ```text
+   util/check_subscription.sh
+   ```
+
+   Example output:
+   ```text
+   Your current subscription is MyOrg Azure Subscription (ID=f1b1baa4-3eda-4940-b907-263813c5b967)
+   ```
+
+### Configuring Authorization with Azure
+
+In order for Terraform/Ansible to manage resources in Azure, a Service Principal is required.
+The following process creates a new service principal in Azure, and stores the details required in an authorization script on the local workstation.
+This script can then be used (_sourced_) to configure the required environment variables on the local workstation that allows Terraform/Ansible to run without prompting the user for further authentication information.
+
+1. To easily create the service principal and authorization script, run the following command providing the name you wish to give the service principal as the only command line argument (here the name `sp-eng-test` is used):
+
+   ```text
+   util/create_service_principal.sh sp-eng-test
+   ```
+
+   Example output:
+   ```text
+   Creating Azure Service Principal: sp-eng-test...
+   Changing "sp-eng-test" to a valid URI of "http://sp-eng-test", which is the required format used for service principal names
+   Creating a role assignment under the scope of "/subscriptions/0cbd16ac-c0b4-4c4a-9bca-4ca95477f0a9"
+     Retrying role assignment creation: 1/36
+     Retrying role assignment creation: 2/36
+   A service principal has been created in Azure > App registrations, with the name: sp-eng-test
+   Azure authorization details can be found within the script: set-sp.sh
+   You can enable this authorization by sourcing the script using the following command:
+   source set-sp.sh
+   ```
+
+1. To source the authorization script, setting the required environment variables, run the following:
+
+   ```text
+   source set-sp.sh
+   ```
+
+   **Note:** The authorization script contains secret information, which you should store and secure appropriately.
+
+   **Note:** This step will need to be repeated for every new shell session. For example, if you logout of your VM, or open a new terminal session.
+
+### Configuring Deployment Template
+
+The SAP environments deployed by this codebase are configured by JSON input files.
+These configuration files provide a high degree of customization for the user, but can be a little daunting if you are new to the codebase.
+Therefore example configuration files have been supplied with the code.
+
+The minimal amount of change required to an example configuration file, is to configure your SAP Launchpad credentials so that the code can automatically login and download the required SAP packages to install.
+
+Configuring your SAP Launchpad credentials for the simplest example JSON input file, requires you to provide your SAP user and password to another utility script.
+
+1. Run the following utility script to configure your SAP download credentials:
+
+   ```text
+   util/set_sap_download_credentials.sh <sap_user> <sap_password>
+   ```
+
+   **Note:** If your SAP Launchpad password has spaces in, you will need to enclose it in double quotes.
+
+## Build/Update/Destroy Lifecycle
+
+1. To easily initialize Terraform, run the following utility script:
+
+   ```text
+   util/terraform_v2.sh init
+   ```
+
+1. To easily deploy the system, run the following utility script:
+
+   ```text
+   util/terraform_v2.sh apply
+   ```
+
+   **Note:** This process can take in the region of 90 minutes to complete.
+   Particularly slow stages are:
+
+     - `Install XSA components` (~35 minutes)
+     - `Extract media archive` (~15 minutes)
+     - `Install SHINE` (~10 minutes)
+     - `Install HANA Database using hdblcm` (~10 minutes)
+     - `Download installation media` (~5 minutes)
+     - `Installing OS package` (~5 minutes)
+
+1. To review/inspect the provisioned resources navigate to the `test_rg` resource group of your configured Azure subscription in Azure portal.
+   By default, all the provisioned resources (excluding the service pricipal) are deployed into the same resource group.
+
+1. To easily delete the provisioned resources, run the following utility script:
+
+   ```text
+   util/terraform_v2.sh destroy
+   ```
+
+## Summary
+
+The following illustrates an example summary of the commands and processes required:
+
+```bash
+# Obtain the Code: Takes about 2 minutes and is performed once
+cd ~/projects/
+git clone https://github.com/Azure/sap-hana.git
+cd sap-hana
+
+# Check Tool Dependencies: Takes under a minute and is performed once
+util/check_workstation.sh
+
+# Configure Target Azure Subscription: Tales under a minute and is performed once per subscription
+az login
+util/check_subscription.sh
+
+# Configure Azure Authorization: Takes under a minute and is performed once per subscription
+util/create_service_principal.sh sp-eng-test
+source set-sp.sh
+
+# Configure Deployment Template: Takes under a minute and is performed once per SAP system build
+util/set_sap_download_credentials.sh S123456789 MySAPpass
+
+# Build/Update Lifecycle: Takes about 90 minutes and is performed once per SAP system build/update
+util/terraform_v2.sh init
+util/terraform_v2.sh apply
+
+# Destroy Lifecycle: Takes about 15 minutes and is performed once per SAP system build
+util/terraform_v2.sh destroy
+```
--- a/deploy/v2/USAGE.md
+++ b/deploy/v2/USAGE.md
@@ -61,6 +61,7 @@ Running the code requires the following tools with the minimal supported/tested 
 |-----------|------------------------------------|
 | Azure CLI | 2.0.63                             |
 | Terraform | 0.12.12                            |
+| Ansible   | 2.8.1 (see note below)             |
 
 1. To easily check which tool versions you have installed, run the following utility script:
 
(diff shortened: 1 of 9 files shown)


util/set_sap_download_credentials.sh
```diff
@@ -0,0 +1,72 @@
...
+set -o nounset
+
+# import common functions that are reused across scripts
+source util/common_utils.sh
+
+
+# location of the input JSON template
+readonly target_path="deploy/v2"
+# readonly target_code="${target_path}/terraform/"
+readonly target_json="${target_path}/template_samples/single_node_hana.json"
```
rukawata: There will be other template json for deployment of other scenarios.
Therefore `target_json` should be an optional input with default value set to `single_node_hana.json`.

util/terraform_v2.sh
```diff
@@ -0,0 +1,110 @@
...
+# exit immediately if an unset variable is used
+set -o nounset
+
+# import common functions that are reused across scripts
+source util/common_utils.sh
+
+
+readonly target_path="deploy/v2"
+readonly target_code="${target_path}/terraform/"
+readonly target_json="${target_path}/template_samples/single_node_hana.json"
```
rukawata: Same as the previous comment.

util/README.md
```diff
@@ -0,0 +1,9 @@
+# Utility Scripts
+
+This directory contains some simple bash scripts that aid usage of the main codebase.
+The scripts are intended to be run on the user's workstation, and not on the machines deployed in Azure.
+The scripts should be run as and when directed as part of [the main usage guide](../deploy/v2/USAGE.md).
+
+All scripts have been tested with `shellcheck` v0.7.0.
```
jeffaco: How about making this a link? Something like:

All scripts have been tested with [shellcheck](https://www.shellcheck.net/) v0.7.0.

util/check_subscription.sh
```diff
@@ -0,0 +1,47 @@
...
+	local subscription_id
+	subscription_name=$(echo "${subscription_info}" | grep '"name":' | cut -d'"' -f4)
+	subscription_id=$(echo "${subscription_info}" | grep '"id":' | cut -d'"' -f4)
+
+	echo -e "Your current subscription is ${subscription_name} (ID=${subscription_id})"
+}
+
+
+# Execute the main program flow with all arguments
+main "$@"
```
jeffaco: I like this flow. You insure that all functions are defined prior to executing anything, thereby avoiding the fact that the implementation must be given before calling it. Nice.

util/create_service_principal.sh
```diff
@@ -0,0 +1,120 @@
...
+	# Check there's just a single argument provided
+	case $# in
+		1)
+			:
+			;;
+		*)
+			continue_or_error_and_exit 1 "You must specify a single command line argument for the service principal name"
+			;;
+	esac
+}
```
jeffaco: It's faster and easier (much fewer lines, more clear of intent) to just check that the argument count is 1. But what does line 37 do, I've not seen that before?

util/create_service_principal.sh
```diff
@@ -0,0 +1,120 @@
...
+	# backup current value to restore to afterwards
+	local ifs_backup="${IFS}"
+	IFS=$'\n'
+
+	# create the service principal and capture the output
+	sp_details=$(az ad sp create-for-rbac --name "${service_principal_name}")
+	local sp_creation_status=$?
+
+	# check the SP was created successfully
+	continue_or_error_and_exit $sp_creation_status "There was a problem creating the service principal. If you are logged into Azure CLI, then it could relate to lack of admin/owner permissions."
```
jeffaco: The scripts are documented and written as if you don't expect a lot of prior user knowledge to use these scripts (which is great). But now you're giving an error message that's not clear (even to me) on what I can do. Is this error documented somewhere, along with steps to mitigate the problem?

Perhaps a separate FAQ, or ERRORS.md, or something like that?

This is important enough to me where I'm going to want you to respond to this prior to me signing off. But other than this (review still in progress), this looks pretty good so far.

util/create_service_principal.sh
```diff
@@ -0,0 +1,120 @@
...
+	client_id=$(echo "${sp_details}"  | grep appId | sed -e 's/.*appId.:.\(.*\),/\1/')
+	client_secret=$(echo "${sp_details}" | grep password | sed -e 's/.*password.:.\(.*\),/\1/')
+
+	# create new bash script for authorization
+	create_new_auth_script
+	append_line_to_auth_script "#!/usr/bin/env bash"
+	append_line_to_auth_script "export ARM_SUBSCRIPTION_ID=${subscription_id}"
+	append_line_to_auth_script "export ARM_TENANT_ID=${tenant_id}"
+	append_line_to_auth_script "export ARM_CLIENT_ID=${client_id}"
+	append_line_to_auth_script "export ARM_CLIENT_SECRET=${client_secret}"
```
jeffaco: Style nit. These seems overly wordy to me, when you could have had a comment at line 76, nix `create_new_auth_script` and `append_line_to_auth_script`, and then use something like `cat << EOD > output_file` followed by the lines you wanted. That would be more efficient too, but that's not my gripe here. It's just that you have multiple routines of indirection where doing inline with a `cat` command would give you what you wanted.

This isn't a show-stopper. Was there some reason you did it this way, though?

util/set_sap_download_credentials.sh
```diff
@@ -0,0 +1,72 @@
...
+{
+	# Check there're just two arguments provided
+	case $# in
+		2)
+			:
+			;;
+		*)
+			continue_or_error_and_exit 1 "You must specify 2 command line arguments for the SAP download credentials: a username and a password"
+			;;
+	esac
```
jeffaco: Again, this pattern is needlessly wordy. You can do something like `[ $# -ne 2 ] && continue_or_error_and_exit ...`. Or just have a multi-line `if` statement (three lines rather than 8.

Why are you choosing this pattern?

util/set_sap_download_credentials.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+
+function edit_json_template_for_sap_credentials()
+{
+	local sap_username="$1"
+	local sap_password="$2"
+
+	# use temp file method to avoid BSD sed issues on Mac/OSX
```
jeffaco: I don't understand "BSD sed issues on Mac/OSX". Can you give a reference to a WWW page that explains this? (You can put it in code if you wish, but I'm asking just for my knowledge - it's fine to just answer me here as well.)

util/terraform_v2.sh
```diff
@@ -0,0 +1,110 @@
...
+	print_line
+	print_line "Usage:"
+	print_line
+	print_tabbed_line "${script_path} <command>"
+	print_line
+	print_line "The commands are:"
+	print_tabbed_line "init"
+	print_tabbed_line "apply"
+	print_tabbed_line "destroy"
+	print_line
```
jeffaco: Hmm. Okay. But why not just substitute `print_line` with `echo`? And `print_tabbed_line "text"` with `echo "\ttext"`? Seems a lot shorter, and faster as well. Again, I'm less concerned about the speed. Just needless indirection and complexity when you can trivially do it in-line?

util/terraform_v2.sh
```diff
@@ -0,0 +1,110 @@
...
+{
+	local options="$1"
+
+	local command="terraform ${options}"
+
+	# describe the command that will be run (useful for debugging)
+	print_line "Running the following Terraform command:"
+	print_line
+	print_tabbed_line "${command}"
+	print_line
```
jeffaco: Ditto here, see above comment.

deploy/v2/USAGE.md
```diff
@@ -0,0 +1,221 @@
+# Usage Guide for V2
+
+This document aims to describe the basic process to get started with V2 of the codebase.
+It does not go into detail about how to customize the code, but instead focuses on demonstrating the end-to-end lifecycle.
+
+## Prerequisites
+
+The code is designed to be used from a Unix-based OS such as MacOS, Linux, Cygwin, or Windows Subsystem for Linux.
```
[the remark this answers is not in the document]
ben-moss: Thanks - I took the above list for OS support from the original doc.

For Windows, wouldn't it only work using RTI-based ansible, since ansible requires a Linux-like control machine?

If we want to claim this works on Windows, we might need to consider fixing the service principal section of the current README's instructions too (`export` env vars and `set-sh.sh would need Windows equivalents, right?).

deploy/v2/USAGE.md
```diff
@@ -0,0 +1,221 @@
...
+
+   ```text
+   util/check_workstation.sh
+   ```
+
+   Example output:
+   ```text
+   azure-cli = 2.0.77
+   Terraform = 0.12.16
+   ansible = 2.8.1
```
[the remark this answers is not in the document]
ben-moss: OK. I think I can handle that. The reason I left it out is because the doc is currently only describing the RTI version, where ansible is not required on the user's workstation. I'll update it.

deploy/v2/USAGE.md
```diff
@@ -0,0 +1,221 @@
...
+
+| Tool      | Minimum Version Supported / Tested |
+|-----------|------------------------------------|
+| Azure CLI | 2.0.63                             |
+| Terraform | 0.12.12                            |
+
+1. To easily check which tool versions you have installed, run the following utility script:
+
+   ```text
+   util/check_workstation.sh
```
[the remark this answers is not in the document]
ben-moss: I debated this point before deciding on the project root. I also considered the root `tools` dir too, as that already contained some shell scripts, but they seem to be for execution on the target systems, rather than the local workstation.

I feel there are some very good reasons for keeping at the root, which include:
1. No long paths in commands, not even as a `cd` at the start of the instructions.
1. If all the scripts run from the project root, and not specific subdirs, then there's never any doubt about where the user needs to be, irrespective of what code they're using or doc they're following, and how the relative paths need to be used.
1. Some (potentially all) of the scripts can be used in V1 codebase as well as V2, but this was just an initial introduction to the idea of scripting the setup, rather than the end design, as I wanted to test whether the idea gained any support/traction before investing too much effort.
1. It would be fairly trivial to have the scripts become an abstraction/indirection layer for using the repo. The V1 docs could be updated to use them, and then when you are ready to migrate to V2, the user experience (and much of the user docs) does not need to change, so existing users of either version, experience minimal change in how they engage with the code from a usage perspective.

Does that help? Happy to discuss in a design session too, if that's more appropriate.

deploy/v2/USAGE.md
```diff
@@ -0,0 +1,221 @@
...
+
+   **Note:** This process can take in the region of 90 minutes to complete.
+   Particularly slow stages are:
+
+     - `Install XSA components` (~35 minutes)
+     - `Extract media archive` (~15 minutes)
+     - `Install SHINE` (~10 minutes)
+     - `Install HANA Database using hdblcm` (~10 minutes)
+     - `Download installation media` (~5 minutes)
+     - `Installing OS package` (~5 minutes)
```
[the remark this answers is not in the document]
ben-moss: You're right, they're sorted by time. It comes from [an ansible plugin](https://docs.ansible.com/ansible/latest/plugins/callback/profile_tasks.html). I can reorder them to execution order - good suggestion.

util/set_sap_download_credentials.sh
ben-moss: Good idea - this was just an initial version - all the code evolves :)

As there was only one template in the codebase, I didn't go to the effort due to all the error handling, but I did make it a variable. However, I had the same idea for the Terraform script, but I think it's a feature to add when we introduce multiple V2 templates into the codebase.

Would you be happy for me to capture this as a new backlog item for the sake of keeping this PR's scope reasonable?

util/check_subscription.sh
ben-moss: Absolutely! Glad you appreciate it 👍 
We use this design pattern in most of our scripting. That way you can build up libraries of tried and tested reusable functions, and don't have to get tied up in ordering of `source` commands, and [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell)!

util/create_service_principal.sh
ben-moss: I believe `:` is [the best way of saying "do nothing"](https://stackoverflow.com/a/3224910) when bash forces you to put _something_ in a code block. It's useful if you're designing scripts and functions from a top down approach - as each function body can start with a skeleton implementation. For example, the following is already a valid bash script where I have planned out the approach without having to think about how each of the functions will be implemented:

```bash
#!/usr/bin/env bash

function main()
{
	 download_software_package
	 install_software_package
	 run_software_package
}

function download_software_package()
{
	:
}

function install_software_package()
{
	:
}

function run_software_package()
{
	:
}

main "$@"
```

Regarding the approach... it's a style that lends itself better to longer term maintenance, when there are multiple supported options. Note the function name isn't limited to handling a single argument, so if we extend the script to handle 3 arguments, where 2 are mandatory and the 3rd is optional, then the code updates are more obvious. Granted, this is an extreme case with the script currently only supporting 1 argument, but it's a scripting style we often use.

Happy to change, if you prefer, though?

util/create_service_principal.sh
ben-moss: Great point, and have done something similar before (like a runbook for troubleshooting common issues). I'd be happy to do that as and when we come across things.

Specifically for the SP creation, I don't think this is trivial, and I didn't want to get into the realms of masking/interpolating the errors that already come back from the Azure CLI when this fails (Note that the user will get errors from `az` as well as the error in the bash script).

In the past, we have put much more error wrapping around CLIs (including `az`), but you end up at the mercy of changes in the upstream CLI breaking your scripts.

In this case, would you be happy with:
1. Alert the user that there has been an error
2. Link to a doc that covers some common, symptoms and recommendations for remediation?

On a side note, I've looked for documentation on what permissions are actually required to create SPs and link them to a subscription, but cannot find any. The best I can determine from experimentation is that the user needs to either be a subscription Owner, or else be a subscription co-administrator (you seem to be able to be a Contributor and a Co-admin).

util/create_service_principal.sh
ben-moss: Great question. There are many reasons for indirection in the code style. This is based on many years of experience gained working with code in team environments - rather than code that only the original author (or a bash guru) understands. Here are some of them:

1. It's a form of self-documentation for those less familiar with bash when performance is not a concern - it's more human readable;
1. It allows refactoring the implementation without changing the calling function (encapsulation/separation of concerns);
1. It creates more testable code - we can write tests against the functions as code units;
1. It can aid portability as functions become reusable in the long term, and proven to work on different operating systems;
1. (And more specific to this example) it allows future evolution without having to pick out and change every call to `cat`, which we might want to also:
   - send to a log file
   - redact secrets from some output stream
   - check the target file exists
   - backup the target file first
   - etc.

That last one is a really powerful concept/pattern for long term maintainability of script across teams/personnel, but as always there are trade-offs.

util/set_sap_download_credentials.sh
ben-moss: See above answers, but happy to change.

util/set_sap_download_credentials.sh
ben-moss: Yeah, I should probably explain that better.

To get native (i.e. without forcing the client to update/install extras) Mac/OSX ships with BSD version of `sed`. This creates challenges using `sed` in a portable way - specifically (in our experience) [when doing inline file edits](https://stackoverflow.com/a/5694430).

Whilst there are claims that using a specific type of command option syntax avoids this problem, I believe I've seen the behaviour regress between OSX version upgrades. Therefore a safer approach is to avoid trying to solve the problem directly and use a workaround. The simplest, although not fool-proof, is to stream and filter the output to a temporary file, and then replace the original with the temporary file. There is another approach which is tidier from a filesystem perspective, but requires more code and indirection - detect the OS and adjust the sed commands based on OS.

util/terraform_v2.sh
ben-moss: See above: https://github.com/Azure/sap-hana/pull/288#discussion_r382557445
In particular point 5 for echo - as the code matures, we often end up sending STDOUT to logs as well as terminal to aid support with customers.

deploy/v2/USAGE.md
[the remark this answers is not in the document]
ben-moss: On reflection, I'm still confused about this. In V2, I believed that ansible was no longer required on the local workstation, irrespective of whether the code splits the Terraform and ansible stages.

Doesn't Terraform always provision Ansible on the RTI, and then all ansible is run from there:
https://github.com/Azure/sap-hana/blob/4f1795550f25a535c5078a25ce02a3527d1e8fd5/deploy/v2/terraform/modules/jumpbox/main.tf#L297

In a full deploy (`"ansible_execution": true`) this would be the case, but even when split (`"ansible_execution": false`), after Terraform is done, the user is given details of the RTI to login and then run ansible subsequently.

Perhaps there's a use case when ansible is run from the user's workstation that I'm unaware of? 😕

deploy/v2/USAGE.md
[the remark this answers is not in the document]
rukawata: Yes, we design RTI concept so that it works on Windows (after all, terraform runs on Windows/Linux/MacOS, ansible is the bottle neck :)).
And yes, you are correct we probably should adjust the SP section - will track that in issue #291

deploy/v2/USAGE.md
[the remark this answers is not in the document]
rukawata: @ben-moss your understanding is correct. Just if you run the script locally on Linux/MacOS, it's also possible to run the playbook from the user's workstation :) We do save a copy of the inventory file (`host`) and `output.json` on the user's workstation under `sap-hana/deploy/v2/ansible_config_files`

deploy/v2/USAGE.md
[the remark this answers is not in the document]
rukawata: The content under `tools` is currently not our scope - they park here temporarily.
Let's quickly discuss in the next meeting since this is not a blocker.

util/set_sap_download_credentials.sh
rukawata: That's fine. Please create an issue to track that. Thanks!

deploy/v2/USAGE.md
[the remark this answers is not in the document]
ben-moss: Thanks for the explanation. I'll add something in to try and capture this point.

deploy/v2/USAGE.md
```diff
@@ -0,0 +1,221 @@
...
+You also need some familiarity with using the following technologies/tools:
+
+1. Linux command line
+1. Git (ideally command line)
+1. Azure portal
+1. SAP Launchpad (login credentials required for SAP software downloads)
+
+### Obtaining the Code
+
+Before continuing you should first obtain a copy of the code, so that you can use the utility scripts provided.
```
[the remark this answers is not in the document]
ben-moss: Is this better, now?

util/set_sap_download_credentials.sh
ben-moss: Captured in #292

util/terraform_v2.sh
ben-moss: Now captured in #292

ben-moss:
@rukawata and @jeffaco - many thanks for all your great feedback. It's great to be challenged on these things as it really forces you to think about the what and why of all those decisions we make, and think about the code from another user's point of view :)

I believe I've attempted to address all the points: some with responses; some with code updates; and some with requests for more info or clarification on the actions I should take.

Would you possibly be able to re-review those points, and click resolve on any you're happy with, and respond on the others so I'm clear what actions need to happen to get your approvals.

Many thanks.

util/create_service_principal.sh
jeffaco: I think coding should be agile. Do what you need at the time, add as needed.

In this case, I think it's clearer and easier to understand to just have a much shorter `if` statement. If you get to the point where things change and you're taking multiple parameters, some optional, then you'll need to change your handling of the parameters anyway. You may do that via a `case` statement, or you may do it other ways, depending on specifics.

TL;DR I prefer a simple `if` statement in these cases, it's easier/faster to read & understand.

util/create_service_principal.sh
jeffaco: You don't need to tell me HOW to fix if it's complicated, but give a hint at least! If I got that error, I'd read it three or four times, sigh, and probably start WWW searches hoping for a hit. But even getting the search term "right" for this would be hard. A link to a doc would be great here, and highly appreciated by a user that happened to hit this.

util/create_service_principal.sh
jeffaco: Both work. I guess I'm not a fan of trying to make bash look like cobol or something. Bash is bash and, if you do significant work in it, you need to learn the language. Wrappers like this won't help. In the #5 case, that could be useful, but it could also be "bad" (you'd need to review all callers to make sure your change didn't conflict).

As I said, this was a style nit. I won't force you to conform. But it's odd, and that's not how we write our scripts here (for our primary product). More important, though, is consistency. If all the scripts that you write do this, then consistency is good. Checking into our code base where we NEVER do this, I'd balk. But for this, as long as you're consistent, then 👍 

Style nit only, feel free to resolve this.

util/set_sap_download_credentials.sh
jeffaco: Yeah, please do. This pattern makes the code harder to understand and interpret with no obvious gain.

util/set_sap_download_credentials.sh
jeffaco: Yeah, okay. I get that. Unfortunately, making Mac OS/X conform is probably going to be a "moving dart-board", particularly with Apple moving away from bash entirely (no longer the default shell in Catalina). I think Apple wants to reduce maintenance in areas of the system they don't directly use. And if WE use it, we suffer.

You might want to put a WWW link in a comment there, but not mandatory.

util/create_service_principal.sh
ben-moss: I'm struggling here @jeffaco. I'm just trying to understand what is acceptable for getting approval?

You asked for a hint, but I believe there are already two hints in the error message that cover the two reasons why this might fail (being logged out, or not having permissions).

Q1: Do you think these hints need to be more obvious, or am I missing something else?

You also mentioned linking to a doc, do you mean an Azure doc that exists already, or the one I've suggested I create in the repo? (If the former, then I previously stated that I haven't been able to find one. Maybe you know of something?).

Q2: Which type of doc link were you referring to?

deploy/v2/USAGE.md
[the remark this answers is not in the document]
rukawata: Track with issue #293

util/create_service_principal.sh
ben-moss: Now switched to heredoc style file creation

util/set_sap_download_credentials.sh
ben-moss: Switched to `if` style

util/create_service_principal.sh
ben-moss: Updated with link to external doc

util/create_service_principal.sh
ben-moss: Switched to `if` style

util/set_sap_download_credentials.sh
ben-moss: Link added

util/terraform_v2.sh
ben-moss: Reverted back to using `echo`

util/terraform_v2.sh
ben-moss: Ditto :)

util/create_service_principal.sh
```diff
@@ -0,0 +1,105 @@
...
+	# backup current value to restore to afterwards
+	local ifs_backup="${IFS}"
+	IFS=$'\n'
+
+	# create the service principal and capture the output
+	sp_details=$(az ad sp create-for-rbac --name "${service_principal_name}")
+	local sp_creation_status=$?
+
+	# check the SP was created successfully
+	continue_or_error_and_exit $sp_creation_status "There was a problem creating the service principal. If you are logged into Azure CLI, then it could relate to lack of admin/owner permissions. See ${sp_link} for further details."
```
jeffaco: This is perfect. My concern was giving enough of a hint to get the search term right. This takes all question out of the equation, the user knows where to go for help. Thanks for taking care of this!

PR #294, opened 2020-02-25
Document and Simplify V2 Usage Process (see #288)

ben-moss:
See #288 for full details.

### DIFF
--- a/util/common_utils.sh
+++ b/util/common_utils.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+###############################################################################
+#
+# Purpose:
+# This file allows bash functions to be used across different scripts without
+# redefining them in each script. It's designed to be "sourced" rather than run
+# directly.
+#
+###############################################################################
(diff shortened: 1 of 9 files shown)


util/common_utils.sh
```diff
@@ -0,0 +1,24 @@
...
+#   and an error message (string argument)
+# This function returns immediately if the status code is zero.
+# Otherwise it prints the error message to STDOUT and exits.
+# Note: The error is prefixed with "ERROR: " and is sent to STDOUT, not STDERR
+function continue_or_error_and_exit()
+{
+	local status_code=$1
+	local error_message="$2"
+
+	((status_code != 0)) && { echo "ERROR: ${error_message}"; exit 1; }
```
jeffaco: I missed this the first time. Technically, if you're printing an error, it should go to `stderr`, not `stdout`. It would be nice if you fixed this here, and in the other checkin for the other group. But not mandatory.

PR #296, opened 2020-02-25
Fix problem when depends_on is not defined for the comp

rukawata:
There is a bug when creating unique list for HANA database component, which will lead to below error when only components without dependency are defined (eg. hana_database, hana_linux_client, hana_linux_studio).

Error:

```
sap-hana/deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml" 40L, 990C written                
        "U",
        "d",
        "f",
        ",",
        " ",
        "'",
        "h",
        "_",
        "t",
        "]"
```

While correcting this wrong behavior, I also address the [comment](https://github.com/Azure/sap-hana/pull/282#discussion_r377801712) in 
PR #282 regarding variable naming so it's clearer.

### DIFF
--- a/deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
+++ b/deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
@@ -2,24 +2,26 @@
 
 - name: Create HANA components list from output JSON
   set_fact:
-    components_list: "{{ components_list + [ item.key ] }}"
+    hdb_comp_list: "{{ hdb_comp_list + [ item.key ] }}"
   loop: "{{ hana_database.components|dict2items }}"
 
 - name: Merge HANA Components list with its depends on components
   set_fact:
-    install_comp: "{{  [ item.depends_on ] + components_list }}"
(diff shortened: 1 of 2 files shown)


deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
```diff
@@ -2,24 +2,26 @@
...
 
 - name: Merge HANA Components list with its depends on components
   set_fact:
-    install_comp: "{{  [ item.depends_on ] + components_list }}"
+    hdb_comp_list: "{{  [ item.depends_on ] + hdb_comp_list }}"
   loop: "{{ components }}"
-  when: 'item.component in components_list'
+  when: 
+    - item.component in hdb_comp_list
+    - item.depends_on is defined
```
tniek: If I read your code correctly, the only change (apart from the naming) is checking for this condition, correct?

rukawata: Yes, the condition is the main problem.

PR #297, opened 2020-02-26
Utility script support for multiple input JSON templates

ben-moss:
## Problem

Changes introduced in #288 allow terraform to be easily run against a single JSON input template (single node HANA), but as more templates are added the scripts will need to allow for the user to make a choice on which to use (see #292).

The current scripts also write errors to STDOUT rather than STDERR (see #295).

## Description

This PR enables the V2 terraform script to be run using any input JSON template within the example templates directory, and provides the user with appropriate usage details and suggestions in the case when no template is given, or the given template does not exist. The USAGE guide is also updated to cover this interface change, and a new "minimal example" input JSON template is added `rti_only` that can be used to provision/destroy just the RTI instance. This can be used to test RTI-related provisioning, and runs in considerably less time than the SAP `single_node_hana` alternative.

The PR also fixes #295 by ensuring all error output created through the error output function is sent to STDERR.

**Note:** This PR does not cover the changes required to enable the `util/set_sap_download_credentials.sh` script to work with multiple input JSON templates. That feature will be added when more SAP input JSON templates are added.

## Pre-Requisites

1. Have followed the initial parts of [the USAGE guide up to section **Build/Update/Destroy Lifecycle**](https://github.com/Azure/sap-hana/blob/226cc37827333e5b96319f297c85313418f50144/deploy/v2/USAGE.md#buildupdatedestroy-lifecycle).
1. `shellcheck` installed on the workstation you're testing from (for test 3)

## Test Instructions

- [ ] 1. Review the [USAGE.md](https://github.com/Azure/sap-hana/pull/297/files#diff-7b2ac2a41f894bd0e6577ab6721d0d9d) guide updates to ensure they are clear.
- [ ] 2. Follow the usage guide to ensure the process works and the new utility scripts run successfully:
   - Potential alternative test cases:
      1. `util/terraform_v2.sh apply rti_only`
      1. `util/terraform_v2.sh destroy rti_only`
- [ ] 3. Shellcheck review the scripts by running `shellcheck util/*.sh` (no results should be shown and there should be a zero exit status from `echo $?` when run after shellcheck)
- [ ] 4. Check some edge cases for reasonable error handling in the `util/terraform_v2.sh` script, such as:
   - Not specifying command line arguments, when they're expected (e.g. `util/terraform_v2.sh`)
   - Specifying only partial command line arguments (e.g. `util/terraform_v2.sh apply`)
   - Specifying template files that don't exist  (e.g. `util/terraform_v2.sh apply test`)
[...]

Issue #292, opened 2020-02-21
Backlog: Ability for utility scripts to work with multiple input.json templates


ben-moss:
As [discussed here](https://github.com/Azure/sap-hana/pull/288/files#r382248486), the initial utility scripts introduced in #288 are only designed to work with the current (and sole) V2 template (singe node HANA). We should add the ability to specify an alternative template.

PR #298, opened 2020-02-26
Utility script support for multiple input JSON templates (#297)

### LINKED ISSUE #295: Backlog: Util scripts should send errors to SYSERR
See discussion: https://github.com/Azure/sap-hana/pull/294#discussion_r384008633

ben-moss:
## Problem

Changes introduced in #288 allow terraform to be easily run against a single JSON input template (single node HANA), but as more templates are added the scripts will need to allow for the user to make a choice on which to use (see #292).

The current scripts also write errors to STDOUT rather than STDERR (see #295).

## Description

This PR enables the V2 terraform script to be run using any input JSON template within the example templates directory, and provides the user with appropriate usage details and suggestions in the case when no template is given, or the given template does not exist. The USAGE guide is also updated to cover this interface change, and a new "minimal example" input JSON template is added `rti_only` that can be used to provision/destroy just the RTI instance. This can be used to test RTI-related provisioning, and runs in considerably less time than the SAP `single_node_hana` alternative.

The PR also fixes #295 by ensuring all error output created through the error output function is sent to STDERR.

**Note:** This PR does not cover the changes required to enable the `util/set_sap_download_credentials.sh` script to work with multiple input JSON templates. That feature will be added when more SAP input JSON templates are added.

## Pre-Requisites

1. Have followed the initial parts of [the USAGE guide up to section **Build/Update/Destroy Lifecycle**](https://github.com/Azure/sap-hana/blob/226cc37827333e5b96319f297c85313418f50144/deploy/v2/USAGE.md#buildupdatedestroy-lifecycle).
1. `shellcheck` installed on the workstation you're testing from (for test 3)

## Test Instructions

- [x] 1. Review the [USAGE.md](https://github.com/Azure/sap-hana/pull/297/files#diff-7b2ac2a41f894bd0e6577ab6721d0d9d) guide updates to ensure they are clear.
- [x] 2. Follow the usage guide to ensure the process works and the new utility scripts run successfully:
   - Potential alternative test cases:
      1. `util/terraform_v2.sh apply rti_only`
      1. `util/terraform_v2.sh destroy rti_only`
- [x] 3. Shellcheck review the scripts by running `shellcheck util/*.sh` (no results should be shown and there should be a zero exit status from `echo $?` when run after shellcheck)
- [x] 4. Check some edge cases for reasonable error handling in the `util/terraform_v2.sh` script, such as:
   - Not specifying command line arguments, when they're expected (e.g. `util/terraform_v2.sh`)
   - Specifying only partial command line arguments (e.g. `util/terraform_v2.sh apply`)
   - Specifying template files that don't exist  (e.g. `util/terraform_v2.sh apply test`)
- [x] 5. Check that errors are now going to STDOUT by running through the following:
  - [x] Check the following does not display the error "ERROR: 'test' is not a valid <JSON template name> to use with the 'destroy' option" on the terminal:
    ```bash
    util/terraform_v2.sh destroy test 2>error.log
    ```
  - [x] Check the above error is in the `error.log` file:
    ```bash
    cat error.log
    ```

## References

1. See #297 for original commit history

### DIFF
--- a/deploy/v2/template_samples/rti_only.json
+++ b/deploy/v2/template_samples/rti_only.json
@@ -59,7 +59,7 @@
         "os": {
           "publisher": "Canonical",
           "offer": "UbuntuServer",
-          "sku": "16.04-LTS"
+          "sku": "18.04-LTS"
         },
         "authentication": {
           "type": "key",
--- a/deploy/v2/template_samples/rti_only.json
+++ b/deploy/v2/template_samples/rti_only.json
@@ -1,0 +1,92 @@
+{
+  "infrastructure": {
+    "region": "eastus",
+    "resource_group": {
+      "is_existing": "false",
+      "name": "test-rg"
+    },
+    "vnets": {
+      "management": {
+        "is_existing": "false",
+        "name": "vnet-mgmt",
+        "address_space": "10.0.0.0/16",
+        "subnet_mgmt": {
+          "is_existing": "false",
+          "name": "subnet-mgmt",
+          "prefix": "10.0.1.0/24",
+          "nsg": {
+            "is_existing": "false",
+            "name": "nsg-mgmt",
+            "allowed_ips": [
+              "0.0.0.0/0"
+            ]
+          }
+        }
+      },
+      "sap": {
+        "is_existing": "false",
+        "name": "vnet-sap",
+        "address_space": "10.1.0.0/16",
+        "subnet_admin": {
+          "is_existing": "false",
+          "name": "subnet-admin",
+          "prefix": "10.1.1.0/24",
+          "nsg": {
+            "is_existing": "false",
+            "name": "nsg-admin"
+          }
+        },
+        "subnet_db": {
+          "is_existing": "false",
+          "name": "subnet-db",
+          "prefix": "10.1.2.0/24",
+          "nsg": {
+            "is_existing": "false",
+            "name": "nsg-db"
+          }
+        }
+      }
+    }
+  },
+  "jumpboxes": {
+    "windows": [],
+    "linux": [
+      {
+        "name": "rti",
+        "destroy_after_deploy": "true",
+        "size": "Standard_D2s_v3",
+        "disk_type": "StandardSSD_LRS",
+        "os": {
+          "publisher": "Canonical",
+          "offer": "UbuntuServer",
+          "sku": "16.04-LTS"
+        },
+        "authentication": {
+          "type": "key",
+          "username": "azureadm"
+        },
+        "components": [
+          "ansible"
+        ]
+      }
+    ]
+  },
+  "databases": [],
+  "software": {
+    "storage_account_sapbits": {
+      "is_existing": false,
+      "account_tier": "Premium",
+      "account_replication_type": "LRS",
+      "account_kind": "FileStorage",
+      "file_share_name": "bits"
+    },
+    "downloader": {}
+  },
+  "sshkey": {
+    "path_to_public_key": "~/.ssh/id_rsa.pub",
+    "path_to_private_key": "~/.ssh/id_rsa"
+  },
+  "options": {
+    "ansible_execution": false
+  }
+}
--- a/util/terraform_v2.sh
+++ b/util/terraform_v2.sh
@@ -21,16 +21,22 @@ set -o nounset
 source util/common_utils.sh
 
 
+readonly input_file_term='<JSON template name>'
 readonly target_path="deploy/v2"
 readonly target_code="${target_path}/terraform/"
-readonly target_json="${target_path}/template_samples/single_node_hana.json"
+readonly target_template_dir="${target_path}/template_samples"
 
 
 function main()
 {
-	# handle no command being supplied
+
+	# default to empty string when 0 args supplied
 	local terraform_action=''
 	[ $# -eq 0 ] || terraform_action="$1"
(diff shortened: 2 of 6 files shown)


deploy/v2/template_samples/rti_only.json
```diff
@@ -0,0 +1,92 @@
...
+    "linux": [
+      {
+        "name": "rti",
+        "destroy_after_deploy": "true",
+        "size": "Standard_D2s_v3",
+        "disk_type": "StandardSSD_LRS",
+        "os": {
+          "publisher": "Canonical",
+          "offer": "UbuntuServer",
+          "sku": "16.04-LTS"
```
rukawata: My bad, we actually switch to support Ubuntu 18.04 instead. However it was not reflected in the `single_node_hana.json` under `deploy/v2/templates_samples`. Can you update that and test the same? Thanks!

rukawata: My bad, we actually switch to support Ubuntu 18.04 instead. However it was not reflected in the `single_node_hana.json` under `deploy/v2/templates_samples`. Can you update that and test the same? Thanks!

util/terraform_v2.sh
```diff
@@ -21,35 +21,59 @@ set -o nounset
...
 		'destroy')
-			terraform_destroy
+			check_command_line_arguments_for_template "$@"
+			terraform_destroy "${template_name}"
 			;;
 		*)
-			print_usage_info
+			print_usage_info_and_exit
 			;;
 	esac
```
rukawata: The only command requires one input is `init`, why don't we:

1. first check if the `terraform_action` is `init`
2. other wise, check the input counts is 2
3. then trigger `apply`/`destroy` depends on `terraform_action`

util/terraform_v2.sh
```diff
@@ -106,5 +152,45 @@ function run_terraform_command()
 	${command}
 }
 
+
+# Given a template file name (without file extension)
```
rukawata: I don't think it's bad to keep the file extension when using the shell script wrapper.
It's clearer to the user that we are using JSON format template, especially when the template got adjusted and does not work, they could first check if there is something wrong with the JSON itself.

util/terraform_v2.sh
```diff
@@ -106,5 +152,45 @@ function run_terraform_command()
...
+
+
+# This function pretty prints all the currently available template file names
+function print_allowed_json_template_names()
+{
+	# list JSON files in the templates dir
+	# filter the output of 'find' to extract just the filenames without extensions
+	# prefix the results with indents and hyphen bullets
+	find ${target_template_dir} -name '*.json' | sed -e 's/.*\/\(.*\)\.json/  - \1/'
+}
```
rukawata: I understand this is for more user friendly experience when using the script :)
The three functions can be just part of the main function since I do not see how we would add more to those functions, especially:

- one liner `print_allowed_json_template_names`.

- `get_json_template_path` is basically string concatenation.

We can move it into main function with the comments you have, which is sufficient to for readability :)

util/terraform_v2.sh
ben-moss: As a user, I am lazy :)
The `.json` is the same for all inputs - a whole ten extra characters to type everytime I run a provision and destroy cycle ;)
Aside from readability, this also ensures the user doesn't have to type the whole file path to select the file.
The idea is to build the CLI as a near approximation of what the user might want to see/understand in say a web portal - selecting from a list of pre-canned scenarios.

util/terraform_v2.sh
ben-moss: Flattening out the functions into the calling code means the code is duplicated to every point it's called (as would any potential bugs in the code!). This makes maintenance a nightmare.

I appreciate there's some hesitancy over adopting this style of writing bash, but it comes from many years of experience and is described in many good practice guides for bash. Even if the function only contains one line (e.g. string concatenation), it still means the calling function doesn't have to work out/remember how it was done previously when it needs to be done again.

deploy/v2/template_samples/rti_only.json
ben-moss: Should everything be switched to 18.04 as the pipelines are also running 16.04?:
https://github.com/Azure/sap-hana/search?q=%2216.04%22&unscoped_q=%2216.04%22
Maybe it's better to change that on a more targeted PR against `master` that I can rebase from?

ben-moss: Now captured on the backlog in #299

util/terraform_v2.sh
ben-moss: I think it's a trade-off either way.
More nesting isn't really a nice thing IMO, and in your proposal, you need to consider 0 args error handling too. Having written a lot of CLI scripts that wrap complex commands with a simpler interface, I've always found `case` without nesting to be the most readable and maintainable.

Perhaps, the pragmatic way to manage counter-proposals is to push a PR with counter-proposal code onto the PR branch that's under review. And then the counter-proposal code can be reviewed and tested by the original PR author?

util/terraform_v2.sh
ben-moss: We could discuss the merits of either approach on #300

util/terraform_v2.sh
rukawata: As discussed, will make this change as backlog.

PR #304, opened 2020-02-28
Disable ansible execution and minimize testing scope

rukawata:
Currently, each pipeline test will takes 1.5 - 2 hrs to finish.
This has become bottle neck for PR to get updated and merged.
Therefore we introduce the below changes, which will reduce the pipeline test to ~15 minutes.

- Disable ansible execution for [milestone 1](https://github.com/Azure/sap-hana/milestone/1) since the this one is terraform only. Will enable ansible execution afterwards.
- Simplify template.json to minimal set (to just deploy single HANA instance)

For the complete scenario, there will be separate PR (#305) to schedule daily job for more coverage, which includes:

- Single HANA instance installation
- XSA installation
- SHINE installation
- HANA Studio/Client installation on Linux/Windows jumpboxes

### DIFF
--- a/templates/ansible-deployment-steps.yaml
+++ b/templates/ansible-deployment-steps.yaml
@@ -19,8 +19,6 @@ steps:
       git checkout $(System.PullRequest.SourceBranch)      
       EOF
       echo "Ansible Deployment disabled"
-# Disable ansible playbook execution since current milestone is TF only
-#      ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" 'OBJC_DISABLE_INITIALIZE_FORK_SAFETY="YES" ANSIBLE_HOST_KEY_CHECKING="False" ansible-playbook -i hosts sap-hana/deploy/v2/ansible/sap_playbook.yml'      
     displayName: 'Ansible Deployment'
     env:
       TF_VAR_azure_service_principal_id: $(hana-pipeline-spn-id)
--- a/templates/ansible-deployment-steps.yaml
+++ b/templates/ansible-deployment-steps.yaml
@@ -18,7 +18,9 @@ steps:
       cd sap-hana
       git checkout $(System.PullRequest.SourceBranch)      
       EOF
-      ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" 'OBJC_DISABLE_INITIALIZE_FORK_SAFETY="YES" ANSIBLE_HOST_KEY_CHECKING="False" ansible-playbook -i hosts sap-hana/deploy/v2/ansible/sap_playbook.yml'
+      echo "Ansible Deployment disabled"
+# Disable ansible playbook execution since current milestone is TF only
+#      ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" 'OBJC_DISABLE_INITIALIZE_FORK_SAFETY="YES" ANSIBLE_HOST_KEY_CHECKING="False" ansible-playbook -i hosts sap-hana/deploy/v2/ansible/sap_playbook.yml'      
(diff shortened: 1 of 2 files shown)


templates/tempGen/template.json
```diff
@@ -50,47 +50,8 @@
...
-        "os": {
-          "publisher": "MicrosoftWindowsServer",
-          "offer": "WindowsServer",
-          "sku": "2016-Datacenter"
-        },
-        "authentication": {
-          "type": "password",
-          "username": "azureadm",
-          "password": "Sap@hana2019!"
-        },
```
tniek: It seems impractical to remove all of these just for the sake of executing the tests more quickly. What if at a later point we decide that a Windows jumpbox should be included in the tests?

That said, is it possible to leave all of these (highlighted) parameters and just adjust the `components` in line 68?

templates/tempGen/template.json
```diff
@@ -50,47 +50,8 @@
...
-        "disk_type": "StandardSSD_LRS",
-        "os": {
-          "publisher": "Canonical",
-          "offer": "UbuntuServer",
-          "sku": "18.04-LTS"
-        },
-        "authentication": {
-          "type": "key",
-          "username": "azureadm"
-        },
```
tniek: Same as above, can we keep these and just remove the entries in `components` (line 89)?

templates/tempGen/template.json
```diff
@@ -136,17 +97,7 @@
...
-        "hana_client_linux": [],
-        "xs": [
-          "xsac_alm_pi_ui",
-          "xsac_portal_serv",
-          "xsac_services",
-          "xsac_ui5_fesv5",
-          "xsac_ui5_sb",
-          "xsac_xsa_cockpit"
-        ],
-        "shine": []
```
tniek: I guess it's not possible to keep these, as it means that these components would get installed?

templates/ansible-deployment-steps.yaml
```diff
@@ -18,7 +18,9 @@ steps:
       cd sap-hana
       git checkout $(System.PullRequest.SourceBranch)      
       EOF
-      ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" 'OBJC_DISABLE_INITIALIZE_FORK_SAFETY="YES" ANSIBLE_HOST_KEY_CHECKING="False" ansible-playbook -i hosts sap-hana/deploy/v2/ansible/sap_playbook.yml'
+      echo "Ansible Deployment disabled"
+# Disable ansible playbook execution since current milestone is TF only
+#      ssh -i /tmp/sshkey -o StrictHostKeyChecking=no "${rti_user}"@"${rti_public_ip}" 'OBJC_DISABLE_INITIALIZE_FORK_SAFETY="YES" ANSIBLE_HOST_KEY_CHECKING="False" ansible-playbook -i hosts sap-hana/deploy/v2/ansible/sap_playbook.yml'
```
tniek: Rather than commenting this out, this should be parametrized

templates/tempGen/template.json
rukawata: The current project has nothing to do with windows jumpbox.
Plus we will have a daily test (see PR #305) where we do include windows jumpbox as well as all components.

templates/tempGen/template.json
rukawata: See above [comment](https://github.com/Azure/sap-hana/pull/304#discussion_r386075792).

templates/tempGen/template.json
rukawata: No. This increase the time for each test to significant amount and now becomes a blocker.
However, this is still be tested just on the daily basis (see [comment](https://github.com/Azure/sap-hana/pull/304#discussion_r386075792)).

templates/ansible-deployment-steps.yaml
rukawata: Ack. Removed. Thanks!

PR #312, opened 2020-03-05
Addition of LB to single node HANA

ben-moss:
Sharing current work in progress...

PR contains:
- A new TF module to aid with Azure resource naming conventions
  - See: [README for module](https://github.com/Azure/sap-hana/blob/a51c8f7f4e995c9b3b22ec6cfed6bfffc3d6b802/deploy/v2/terraform/modules/azure_sap_naming_conventions/README.md)
- A new TF module for HANA load balancers
- A usage example of how the module for HANA LBs can be invoked to provision and configure a LB with a single node HANA instance

Build this PR demo via:
- `util/terraform_v2.sh init`
- `util/terraform_v2.sh apply single_node_hana`

Notes:
- The health probe and LB rules can take a while to appear in the Azure portal
- There may be some problems with the ansible when testing, but that's unrelated to the TF changes in this PR (we believe)

Example Naming Convention Screenshot:
image


deploy/v2/terraform/modules/hdb_load_balancer/variables.tf
```diff
@@ -0,0 +1,53 @@
...
+    ]
+
+    "2" = [
+      "3${var.instance_number}13",
+      "3${var.instance_number}15",
+      "3${var.instance_number}40",
+      "3${var.instance_number}41",
+      "3${var.instance_number}42",
+      "5${var.instance_number}13",
+      "5${var.instance_number}14",
```
rukawata: What are these 2 rules for? I don't see them in the [doc](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability).

ben-moss: These were added by mistake in working on a generic ASCS/HANA LB module.
Agreed action to only include the 3 x rules from the current reference architecture:
> Load-balancing rules: 30313 TCP, 30315 TCP, 30317 TCP

util/terraform_v2.sh
```diff
@@ -54,6 +54,7 @@ function main()
 		'destroy')
 			check_command_line_arguments_for_template "$@"
 			terraform_destroy "${template_name}"
+			;;
```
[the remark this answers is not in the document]
ben-moss: Rebasing from master should now resolve this.

deploy/v2/terraform/modules/azure_sap_naming_conventions/main.tf
```diff
@@ -0,0 +1,15 @@
...
+
+module "resource_group_name" {
+  source   = "gsoft-inc/naming/azurerm//modules/networking/load_balancer"
+  name     = local.net
+  prefixes = [local.env, local.loc]
+  suffixes = ["INFRASTRUCTURE"]
+}
+
+module "load_balancer_name" {
+  source   = "gsoft-inc/naming/azurerm//modules/networking/load_balancer"
```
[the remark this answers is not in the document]
ben-moss: The 3rd-party module is already coded to work with all the existing constraints in Azure resource naming, but after discussion we agreed to drop this idea, and re-code all this login in the project codebase.

deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -73,6 +73,60 @@ resource "azurerm_network_interface" "nics-dbnodes-db" {
...
+  for_each            = local.loadbalancers
+  resource_group_name = var.resource-group[0].name
+  loadbalancer_id     = azurerm_lb.hana-lb[0].id
+  name                = "hana-${each.value.sid}-lb-hp"
+  port                = "625${each.value.instance_number}"
+  protocol            = "Tcp"
+  interval_in_seconds = 5
+  number_of_probes    = 2
+}
+
```
rukawata: Please add below comments into the code.

```suggestion
# TODO:
# Current behavior, it will try to add all VMs in the cluster into the backend pool, which would not work since we do not have availability sets created yet.
# In a scale-out scenario, we need to rewrite this code according to the scale-out + HA reference architecture.
```

peterchoo:
Squashing and Merging to create PR to master.

PR #315, opened 2020-03-07
[DRAFT] Use symlink to share common variables across modules

rukawata:
There are several variables shared across different modules.
This PR will change to use symlink to share the `terraform/variable.tf` among all modules.

As a follow up item, there will be a separate PR to extend the content of the shared `terraform/variable.tf` with better descriptions as well as default values if not assigned.


rukawata:
I will close this PR since symlink won't work on Windows.

Issue #220, opened 2019-09-05
ha_pair deployment failed with "Failed to provide Package python-pyparsing-2.1.10-1.14.noarch"


avistavkin:
ha_pair deployment failed with the 

next error:

"Failed to provide Package python-pyparsing-2.1.10-1.14.noarch (SLE12-SP3-SAP-Pool). Do you want to retry retrieval?"

More detials about error and terraform.tfvars files are attached:
[ha_pair_error.txt](https://github.com/Azure/sap-hana/files/3579806/ha_pair_error.txt)
[terraform.tfvars.txt](https://github.com/Azure/sap-hana/files/3579809/terraform.tfvars.txt)

rukawata:
Thanks for the logs. We will look into this.

Thanks,
Nancy

rukawata:
Open bug with Suse: https://bugzilla.suse.com/show_bug.cgi?id=1150395

PR #316, opened 2020-03-09
[HOTFIX] Fix broken saplandscaperunner

rukawata:
**Background:**
saplandscaperunner keeps failing for two major reasons:
- Install package intermittent fails for the first package being installed after VM created
- Resource creation/deletion fails when ansible is used to create/delete azure resources

**Fix includes:**
- Remove workaround for issue #220. 
  - By updating all supported image to SLES12 SP5, [SUSE bug](https://bugzilla.suse.com/show_bug.cgi?id=1150395) is away 
(fixed with `cloud-regionsrv-client 9.0.3` and now we always get `cloud-regionsrv-client 9.0.8`)
  - Workround for issue #220 has side effect that sometimes make the repo not available
- Add retry logic for the first (not all) package install
  - There is no easy way to add retry logic to all package installs
  - We only need to retry for the fist package since afterwards the first success installing package, the repo should be avaible for the following package installation
- Add logic to only remove `linux_bastion` VM if it is installed
  - There is a compatibility issue between `azure-cli` and `azure-cli-core` when installing them on the same machine ([link](https://github.com/ansible/ansible/pull/61238#issuecomment-527998744))
  - There is a bug in the current logic where we try to delete a linux bastion although it is not created - it worked with `azure-cli-core 2.0.35` which is questionable. However by adding azure-cli to the same docker image, `azure-cli-core` will be on a higher version (currently pin to `2.0.71` to resolve most of the compatibility issue), and hits the compatibility issue.
- Rename  [azure_rm_publicipaddress_facts](https://docs.ansible.com/ansible/latest/modules/azure_rm_publicipaddress_facts_module.html) to azure_rm_publicipaddress_info since azure_rm_publicipaddress_facts is outdated.

### DIFF
--- a/deploy/vm/modules/playbook-execution/variables.tf
+++ b/deploy/vm/modules/playbook-execution/variables.tf
@@ -178,7 +178,7 @@ variable "url_retries_cnt" {
 }
 
 variable "url_retries_delay" {
-  description = "The time between each attempt to download the installation bits from the URLs"
+  description = "The time (in seconds) between each attempt to download the installation bits from the URLs"
   default     = 10
 }
 
@@ -188,7 +188,7 @@ variable "package_retries_cnt" {
 }
 
 variable "package_retries_delay" {
-  description = "The time between each attempt to install packages"
+  description = "The time (in seconds) between each attempt to install packages"
   default     = 10
 }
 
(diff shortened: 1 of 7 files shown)


deploy/vm/modules/playbook-execution/variables.tf
```diff
@@ -182,6 +182,16 @@ variable "url_retries_delay" {
   default     = 10
 }
 
+variable "package_retries_cnt" {
+  description = "The number of attempts to install packages"
+  default     = 10
+}
+
+variable "package_retries_delay" {
+  description = "The time between each attempt to install packages"
```
tniek: I assume this is in seconds?

rukawata: Yes. Also add the unit to url_retries_delay for completion :)

PR #318, opened 2020-03-10
Addition of LB to single node HANA

peterchoo:
PR to Master following fork PR #312 

See issue #302

PR contains:
- An update to the HANA DB TF module to add a load balancer

Prerequisites:
- Follow the [Usage Guide](https://github.com/Azure/sap-hana/blob/master/deploy/v2/USAGE.md) to ensure code configuration and authorization to Azure

Build this PR demo via:
- `util/terraform_v2.sh init`
- `util/terraform_v2.sh apply single_node_hana`

Notes:
- The health probe and LB rules can take a while to appear in the Azure portal
- There may be some problems with the ansible when testing, but that's unrelated to the TF changes in this PR (we believe)

PR #323, opened 2020-03-11
Staging PR: Adding Availability Set to HDB Node module

peterchoo:
This PR adds an availability set in, and associates any VMs deployed within the module with it.

Some observations I have about this code:

1. The source data used is the same `local.loadbalancers` hash created - should this be renamed? If so, to what?
1. As this is using the same dataset as the Load Balancer, it will have the same issue about future work with multiple nodes/databases.
1. The [comment added to the Load Balancer PR](https://github.com/Azure/sap-hana/pull/318/files#diff-07f5118af64e7689be637b6e2fa43b29R110-R112) suggests that there may be nodes added by this module which are not part of the backend pool/availability set

### DIFF
--- a/deploy/v2/terraform/modules/hdb_node/main.tf
+++ b/deploy/v2/terraform/modules/hdb_node/main.tf
@@ -73,6 +73,63 @@ resource "azurerm_network_interface" "nics-dbnodes-db" {
   }
 }
 
+# LOAD BALANCER ===================================================================================================
+
+resource "azurerm_lb" "hana-lb" {
+  for_each            = local.loadbalancers
+  name                = "hana-${each.value.sid}-lb"
+  resource_group_name = var.resource-group[0].name
+  location            = var.resource-group[0].location
+
+  frontend_ip_configuration {
+    name                          = "hana-${each.value.sid}-lb-feip"
+    subnet_id                     = var.subnet-sap-db[0].id
+    private_ip_address_allocation = "Static"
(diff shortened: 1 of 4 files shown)


deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -130,6 +130,16 @@ resource "azurerm_lb_rule" "hana-lb-rules" {
...
 
+# AVAILABILITY SET ================================================================================================
+
+resource "azurerm_availability_set" "hana-as" {
+  for_each            = local.loadbalancers
+  name                = "${each.value.sid}-as"
+  location            = var.resource-group[0].location
+  resource_group_name = var.resource-group[0].name
+  managed             = true
+}
```
rukawata: `platform_fault_domain_count` needs to be defined.
Probably you would also need to set `platform_update_domain_count` 
Otherwise, the deploy failed with error:

```
Error: compute.AvailabilitySetsClient#CreateOrUpdate: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidParameter" Message="The specified fault domain count 3 must fall in the range 1 to 2." Target="platformFaultDomainCount"

  on deploy/v2/terraform/modules/hdb_node/main.tf line 135, in resource "azurerm_availability_set" "hana-as":
 135: resource "azurerm_availability_set" "hana-as" {
```

peterchoo:
> See comment. Thanks!

Looking now.

> Also azure pipeline will fail if you create PR from Centiq repo to Azure repo.

I've pointed it back to a staging branch.

deploy/v2/terraform/modules/hdb_node/main.tf
peterchoo: Interesting! I didn't have this failure when I tested this. Surprising that the default for the `platform_fault_domain_count` is outside the acceptable range.

Values have been added.

PR #333, opened 2020-03-17
[HOTFIX] Revert vm implementation for hdb node to fix #330

### LINKED ISSUE #330: Failing to connect to hdb host 10.1.2.4 via ssh
Failing to connect to hdb host via 10.1.2.4. I can see two NICs (10.1.1.4 and 10.1.2.4) are attached to the VM hdb1 from Portal, however i am having a difficulty to connect from linux jumpbox to hdb1 via ssh. I wonder if the second NIC was correctly configured / enabled from the OS side so that the OS acknowledge all the attached NICs. BTW I didn't encounter this issue yesterday. 
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/multiple-nics

rukawata:
Using azurerm_linux_virtual_machine creates problem to SSH into HANA servers from jumpboxes (issue #330).
Therefore revert that part of change until the issue is resovled.

### DIFF
--- a/deploy/v2/terraform/modules/hdb_node/main.tf
+++ b/deploy/v2/terraform/modules/hdb_node/main.tf
@@ -167,47 +167,62 @@ resource "azurerm_managed_disk" "data-disk" {
 }
 
 # Manages Linux Virtual Machine for HANA DB servers
-resource "azurerm_linux_virtual_machine" "vm-dbnode" {
-  for_each            = local.dbnodes
-  name                = each.value.name
-  computer_name       = each.value.name
-  location            = var.resource-group[0].location
-  resource_group_name = var.resource-group[0].name
-  availability_set_id = azurerm_availability_set.hana-as[0].id
-  network_interface_ids = [
-    azurerm_network_interface.nics-dbnodes-admin[each.key].id,
-    azurerm_network_interface.nics-dbnodes-db[each.key].id
-  ]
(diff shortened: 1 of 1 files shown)


deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -167,55 +167,70 @@ resource "azurerm_managed_disk" "data-disk" {
...
+# TODO:
+# Currently there is an issue with SSH logon to HANA database VM if it is created with azurerm_linux_virtual_machine
+# Will migrate this code to use azurerm_linux_virtual_machine when the above issue fixed
+
+# Creates HANA database VM using azurerm_virtual_machine
+resource "azurerm_virtual_machine" "vm-dbnode" {
+  for_each                      = local.dbnodes
+  name                          = each.value.name
+  location                      = var.resource-group[0].location
+  resource_group_name           = var.resource-group[0].name
```
peterchoo: Availability set association is missing in this revert.

```suggestion
  resource_group_name           = var.resource-group[0].name
  availability_set_id           = azurerm_availability_set.hana-as[0].id
```

ben-moss: LGTM 👍

rukawata: > Availability set association is missing in this revert.

Sorry, forgot it again. Now added. Thanks!

PR #334, opened 2020-03-17
Reuse existing storage account in v2 pipeline

rukawata:
With this PR:
1. We will be able to skip download SAP bits if it exists and is extracted.
2. We will be use existing storage account with prepared SAP bits for v2 pipeline ([link](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build?definitionId=6&_a=summary))

We will still test the E2E process including download SAP bits in v2 daily pipeline ([link](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build?definitionId=32&_a=summary)).

### DIFF
--- a/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
+++ b/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
@@ -21,7 +21,7 @@
 
 - name: Download installation media
   shell: "python3 {{ sap_media_download_script_loc }}/python/downloader/basket.py --config /home/{{ ansible_user}}/output.json --dir {{ azure_files_mount_path }}"
-  when: extract_path_status == false
+  when: extract_path_status.stat.exists == false
 
 - name: Identify media archive
   shell: "ls {{ azure_files_mount_path }}/DB|grep ZIP"
--- a/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
+++ b/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
@@ -15,8 +15,13 @@
     pip3 install beautifulsoup4
   become: true
 
+- name: Check whether the installation media already exists and extracted
+  stat: path={{ azure_files_mount_path }}/DB/DATA_UNITS
+  register: extract_path_status
+
 - name: Download installation media
   shell: "python3 {{ sap_media_download_script_loc }}/python/downloader/basket.py --config /home/{{ ansible_user}}/output.json --dir {{ azure_files_mount_path }}"
+  when: extract_path_status == false
 
 - name: Identify media archive
   shell: "ls {{ azure_files_mount_path }}/DB|grep ZIP"
@@ -27,3 +32,4 @@
     src: "{{ azure_files_mount_path }}/DB/{{bits_file.stdout}}"
     dest: "{{ azure_files_mount_path }}/DB"
     remote_src: yes
+    creates: "{{ azure_files_mount_path }}/DB/DATA_UNITS"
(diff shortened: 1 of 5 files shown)


deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
```diff
@@ -15,8 +15,13 @@
     pip3 install beautifulsoup4
   become: true
 
+- name: Check whether the installation media already exists and extracted
```
tniek: nit: "and extract it"

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
```diff
@@ -15,8 +15,13 @@
...
+  stat: path={{ azure_files_mount_path }}/DB/DATA_UNITS
```
tniek: Not a fan of the hard-coded path (what if it changes).
Can you instead check if there are files/folders inside the DB directory that you create?

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
```diff
@@ -27,3 +32,4 @@
     src: "{{ azure_files_mount_path }}/DB/{{bits_file.stdout}}"
     dest: "{{ azure_files_mount_path }}/DB"
     remote_src: yes
+    creates: "{{ azure_files_mount_path }}/DB/DATA_UNITS"
```
tniek: See above

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
rukawata: In a storage account that belongs to customer, they can create any directory they want.

If we check any folders are created, then which folder, this is unknown.
If we check if any files exists, we could only sample one of the files/folders SAP creates in their deliverables, then we don't have control over that either.

This whole procedure needs to follow certain rule (eg. fix the name of the folder), so that we at least know what to expect.

tniek: Agree that a customer could create/modify/delete anything they want. That’s not the point.

Rather than checking if a specific folder inside the DB folder (which I understand is created by our scripts) exists, I’m asking if we can check if anything has been placed into the DB folder (file or new folder). That way, we don’t make ourselves dependent on a specific folder if SAP at some point decides to change the contents of their archive contents.

rukawata: I think we can't get around with SAP defined folder structures.
Maybe we can get away with defining our own extract target, however the sub folder in that extracted folder is still controlled by SAP.

I agree to have this hardcoded path here is not good. I will implement that as a global variable for ansible.

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
rukawata: "been extracted", meaning extract of ZIP file was done before.

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
rukawata: Updated with global variable.

deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
```diff
@@ -27,3 +32,4 @@
     src: "{{ azure_files_mount_path }}/DB/{{bits_file.stdout}}"
```
tniek: 1. Spacing after {{ and before }}
2. Why is `stdout` needed here?

rukawata: This is to find the ZIP file without hardcode the ZIP file name :)
I fixed the spacing. Thanks for catching that.

PR #328, opened 2020-03-12
Migrate to azurerm 2.0

rukawata:
Migrate to use azure_rm 2.0 in Terraform.

PR #335, opened 2020-03-18
Fix issue using azurerm_linux_virtual_machine

rukawata:
In PR #328 the primary nic is assigned with the wrong nic.
With `azurerm_linux_virtual_machine`, the first nic in `network_interface_ids` is the primary nic.
This PR will fix that.

### DIFF
--- a/deploy/v2/terraform/modules/hdb_node/main.tf
+++ b/deploy/v2/terraform/modules/hdb_node/main.tf
@@ -175,8 +175,8 @@ resource "azurerm_linux_virtual_machine" "vm-dbnode" {
   resource_group_name = var.resource-group[0].name
   availability_set_id = azurerm_availability_set.hana-as[0].id
   network_interface_ids = [
-    azurerm_network_interface.nics-dbnodes-admin[each.key].id,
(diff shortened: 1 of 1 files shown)


tniek:
> In PR #328 the primary nic is assigned to the wrong nic.
> 
> This PR will fix that.

Don’t you mean “the primary nic is assigned to the wrong subnet?”

rukawata:
> > In PR #328 the primary nic is assigned to the wrong nic.
> > This PR will fix that.
> 
> Don’t you mean “the primary nic is assigned to the wrong subnet?”

I meant primary nic is assigned **with** the wrong nic.
With `azurerm_linux_virtual_machine`, the first nic in `network_interface_ids` is the primary nic.
I just added that to the PR description.

We do have two follow up work items which requires additional clarification on:

1. Currently in db subnet is used for communication between jumpboxes and HANA servers. I personal think it's not correct but pending on getting further conformation.
2. Currently in the NSG applied on db subnet, it only allows traffic from jumpboxes to HANA servers, without denying other traffic. It needs to be added.
I will act on the above two items in a separate PR.

PR #336, opened 2020-03-18
Add script to enable kdump

prtyag:
This script allows customer to enable kdump on their bare metal system. The steps used for kump is being recommended by suse. please see [this ](https://github.com/SUSE-Enceladus/azure-li-services/issues/201)issue to get more information on this.
This script has been tested on LI and it worked fine.


tools/enable-kdump.sh
```diff
@@ -0,0 +1,50 @@
...
+        exit 1
+    fi
+}
+
+# check if the kexec-tool is enabled
+rpm_kexec_tools_pkg=$(rpm -qa | grep kexec-tools)
+ExitIfFailed $? "kxec-tools required to enable kdump, please install"
+
+# get low value reported by kdumptool calibrate
+low=$(kdumptool calibrate | grep "^Low:" | tr -dc '0-9')
```
jaiawasthi: what are the low / high values here generally ?

tools/enable-kdump.sh
```diff
@@ -0,0 +1,50 @@
...
+ExitIfFailed $? "Enable to start kdump service"
+
+systemctl enable kdump
+ExitIfFailed $? "Failed to enable kdump service"
+
+# set kernel.sysrq to 184(remomended value)
+sysctl kernel.sysrq=184
+
+# load the new kernel.sysrq
+sysctl -p
```
jaiawasthi: exitif failed just incase we werent able to laod the new values ?

tools/enable-kdump.sh
```diff
@@ -0,0 +1,50 @@
...
+#high memory to use for kdump
+high_to_use=$(( $high*$mem))
+
+# replace high and low value in /boot/grub2/grub.cfg
+sed -i "s/crashkernel=[0-9]*M,high/crashkernel=$high_to_use\M,high/gI" /boot/grub2/grub.cfg
+ExitIfFailed $? "Enable to change kernal crash high value in /boot/grub2/grub.cfg"
+
+sed -i "s/crashkernel=[0-9]*M,low/crashkernel=$low\M,low/gI" /boot/grub2/grub.cfg
+ExitIfFailed $? "Enable to change kernal crash low value in /boot/grub2/grub.cfg"
+
```
jaiawasthi: i had a doc as well for kdump.
It mentions below steps as well, are we missing those ?
 Configure a location where the core dumps will be saved in /etc/sysconfig/kdump configuration file, make sure KDUMP_DUMPLEVEL is set to non-zero (recommended to use 31). Also decide the number of OLD dumps to be saved.

tools/enable-kdump.sh
prtyag: please see [this ](https://doc.opensuse.org/documentation/leap/archive/42.3/tuning/html/book.sle.tuning/cha.tuning.kexec.html#sec.tuning.kexec.crashkernel)

tools/enable-kdump.sh
prtyag: yes but i'm not sure about those parameter as of now so we can keep those parameter to be default but if everyone aggrees on that i can do that. BTW you are talking about the doc that CRE gives customer to configure kdump? if yes  i have access to that.

tools/enable-kdump.sh
prtyag: sure will change it

tools/enable-kdump.sh
jaiawasthi: yes, the same doc, just ensure the defaults are good enough for us

hrushib:
Looks good to me.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,50 @@
+ExitIfFailed()
+{
+    if [ $1 != 0 ]; then
+        echo "$2 ! Exiting !!!!"
+        exit 1
+    fi
+}
+
```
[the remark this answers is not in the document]
hrushib: We should add header and mention details with example.

tools/enable-kdump.sh
prtyag: added support to make sure default value for SAVEDIR and KDUMP_DUMPLEVEL are set.

tools/enable-kdump.sh
prtyag: done

tools/enable-kdump.sh
```diff
@@ -0,0 +1,59 @@
...
+sed -i "s/crashkernel=[0-9]*M,high/crashkernel=$high_to_use\M,high/gI" /boot/grub2/grub.cfg
+ExitIfFailed $? "Enable to change kernal crash high value in /boot/grub2/grub.cfg"
+
+sed -i "s/crashkernel=[0-9]*M,low/crashkernel=$low\M,low/gI" /boot/grub2/grub.cfg
+ExitIfFailed $? "Enable to change kernal crash low value in /boot/grub2/grub.cfg"
+
+# set KDUMP_SAVEDIR in /etc/sysconfig/kdump
+sed -i "s/^KDUMP_SAVEDIR=\".*\"/KDUMP_SAVEDIR=\"\/var\/crash\"/gI" /etc/sysconfig/kdump
+
+# set KDUMP_DUMPLEVEL to 31(recommender)
```
sasarava: 'recommender' - is this a typo? It should be 'recommended' I guess

tools/enable-kdump.sh
```diff
@@ -0,0 +1,59 @@
...
+sed -i "s/^KDUMP_DUMPLEVEL=[0-9]*/KDUMP_DUMPLEVEL=31/gI" /etc/sysconfig/kdump
+
+# start and enable kdump service
+systemctl start kdump
+ExitIfFailed $? "Enable to start kdump service"
+
+systemctl enable kdump
+ExitIfFailed $? "Failed to enable kdump service"
+
+# set kernel.sysrq to 184(remomended value)
```
sasarava: Typo in recommended.

tools/README.md
```diff
@@ -8,3 +8,5 @@ dumpsystem.sh: Produces dumps of the bare metal (LI or VLI) system to
 
 sar-iostat.sh: Collect statistics from sar and iostat, control overall
   disk space utilized.
+
+enable-kdump.sh: Enables kdump on a bare metal machine.
```
sasarava: Do we need to be more specific here? 'bare metal machine' is not specific. In the script, its mentioned only LI.

schaefi:
I was asked to review this pull request. I'm happy to share my thoughts. Before I go into details you should notice several complains from the shellcheck utility. I'll attach them here:

```
shellcheck enable-kdump.sh 

In enable-kdump.sh line 1:
# This script enables kdump on LI systems
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang.


In enable-kdump.sh line 5:
    if [ $1 != 0 ]; then
         ^-- SC2086: Double quote to prevent globbing and word splitting.


In enable-kdump.sh line 12:
rpm_kexec_tools_pkg=$(rpm -q kexec-tools)
^-----------------^ SC2034: rpm_kexec_tools_pkg appears unused. Verify use (or export if used externally).


In enable-kdump.sh line 28:
high_to_use=$(( $high*$mem))
                ^---^ SC2004: $/${} is unnecessary on arithmetic variables.
                      ^--^ SC2004: $/${} is unnecessary on arithmetic variables.

For more information:
  https://www.shellcheck.net/wiki/SC2148 -- Tips depend on target shell and y...
  https://www.shellcheck.net/wiki/SC2034 -- rpm_kexec_tools_pkg appears unuse...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

Those should be taken into account.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,59 @@
...
+rpm_kexec_tools_pkg=$(rpm -q kexec-tools)
+ExitIfFailed $? "kxec-tools required to enable kdump, please install"
+
+# get low value reported by kdumptool calibrate
+low=$(kdumptool calibrate | grep "^Low:" | tr -dc '0-9')
+ExitIfFailed $? "Failed to get Low value using kdumptool calibrate"
+
+# get high value reported by kdumptool calibrate
+high=$(kdumptool calibrate | grep "^High:" | tr -dc '0-9')
+ExitIfFailed $? "Failed to get High value using kdumptool calibrate"
```
schaefi: kdumptool reports key value pairs that can be imported in the shell environment. Instead of calling it twice and grepping the desired value you could consider something like this:

```bash
eval $(kdumptool calibrate | sed -e s"@: @=@")
if [ -z "${High}" ];then
    echo "Failed to get High value from kdumptool"
fi
if [ -z "${Low}"];then
    echo "Failed to get Low value from kdumptool"
fi
```

tools/enable-kdump.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# get system memory in tb
+mem=$(free --tera | awk 'FNR == 2 {print $2}')
+ExitIfFailed $? "Failed to get memory using free command"
+
+#high memory to use for kdump
+high_to_use=$(( $high*$mem))
```
schaefi: The math should be done only if mem is greater than 1TB

tools/enable-kdump.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# set KDUMP_DUMPLEVEL to 31(recommender)
+sed -i "s/^KDUMP_DUMPLEVEL=[0-9]*/KDUMP_DUMPLEVEL=31/gI" /etc/sysconfig/kdump
+
+# start and enable kdump service
+systemctl start kdump
+ExitIfFailed $? "Enable to start kdump service"
+
+systemctl enable kdump
+ExitIfFailed $? "Failed to enable kdump service"
```
schaefi: Your script changes the kernel memory region used for the crash dump. Those changes requires a reboot to become effective. You can enable the kdump service here but starting it is imho pointless without a reboot. On reboot an enabled service gets also started. So in my opinion your script needs to end with a reboot of the machine. I also recommend that it's not the script that does it but provides a message to the user that the crash dump memory region has changed and a reboot is required to make it effective

tools/enable-kdump.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# set kernel.sysrq to 184(remomended value)
+sysctl kernel.sysrq=184
```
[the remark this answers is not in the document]
schaefi: yes it's worth a comment what that number does. To be honest I have no idea.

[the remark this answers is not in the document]
prtyag: I also don't have any idea what this value is. I saw this value in the document that CRE is using for enabling a kdump so i used here. and it looks like default value is also 184.

tools/enable-kdump.sh
prtyag: updated.

tools/enable-kdump.sh
prtyag: updated.

tools/enable-kdump.sh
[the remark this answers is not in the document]
hrushib: https://www.suse.com/support/kb/doc/?id=000016171

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+# in the form of key value pair so these can be
+# imported in shell varible
+eval $(cat /etc/os-release | sed -e s"@: @=@")
+
+# check if the os and version is supported by this script
+supported="false"
+for i in "${supported_os[@]}"; do
+    if [[ "$NAME" == "$i" ]]; then
+        for j in "${supported_version[@]}"; do
+            if [[ "$VERSION" == "$j" ]]; then
```
lalitb: where is $VERSION defined ? I don't think this is default bash variable.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+# get OS name and OS version
+# /etc/os-release file has this information
+# in the form of key value pair so these can be
+# imported in shell varible
+eval $(cat /etc/os-release | sed -e s"@: @=@")
+
+# check if the os and version is supported by this script
+supported="false"
+for i in "${supported_os[@]}"; do
+    if [[ "$NAME" == "$i" ]]; then
```
lalitb: how is $NAME getting initialized ?

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+}
+
+# operating system supported by this script
+supported_os=(
+    "SLES"
+)
+
+# operating system versions supported by this script
+supported_version=( "12-SP3"
```
lalitb: why do we need two different variables. Just define:
supported_os= (
"SLES 12 -SP4",
--
)
This will also avoid nested loop added to validate the OS

tools/enable-kdump.sh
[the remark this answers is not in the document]
prtyag: added supported os and version detail.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+            if [[ "$VERSION" == "$j" ]]; then
+                supported="true"
+                break
+            fi
+        done
+        break
+    fi
+done
+if [[ "$supported" == "false" ]]; then
+    echo "This script does not support current OS $NAME, VERSION $VERSION. Please raise request to support this OS and Version"
```
sasarava: ` Please raise request to support this OS and Version`. This may not be very helpful. Do we need to be explicit on how to raise the request?

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+rpm -q kexec-tools
+ExitIfFailed $? "kxec-tools required to enable kdump, please install"
+
+# get low and high value reported by kdumptool calibrate
+# kdumptool calibrate reports key value pair
+# so these can be imported in shell environment
+eval $(kdumptool calibrate | sed -e s"@: @=@")
+ExitIfFailed $? "Failed to run kdumptool calibrate command"
+
+# get system memory in tb
```
sasarava: `tb` to `TB`

tools/enable-kdump.sh
prtyag: that I'll mention in the public document

tools/enable-kdump.sh
prtyag: as discuss offline, this will be loaded as a env variable.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# set kernel.sysrq to 184(recommended)
+sysctl kernel.sysrq=184
+ExitIfFailed $? "Failed to set kernel.sysrq value to 184"
+
+# load the new kernel.sysrq
+sysctl -p
+ExitIfFailed $? "Failed to load new kernel.sysrq value"
+
+echo "KDUMP is successfully enabled, please reboot system to apply change"
```
sasarava: `KDUMP is successfully enabled, please reboot system to apply change` - Please check the grammar of this sentence. My suggestion would be "KDUMP is successfully enabled, please reboot the system to apply the change"

tools/enable-kdump.sh
prtyag: etc/os-release file has name and version so for implementation purpose did it.

tools/enable-kdump.sh
prtyag: see line 29.

tools/enable-kdump.sh
```diff
@@ -0,0 +1,100 @@
...
+    "12-SP5"
+    "15-SP1"
+    "15-SP2"
+)
+
+# get OS name and OS version
+# /etc/os-release file has this information
+# in the form of key value pair so these can be
+# imported in shell varible
+eval $(cat /etc/os-release | sed -e s"@: @=@")
```
sasarava: Will this file present in all the OSs? If it is not present, how will it be handled? https://unix.stackexchange.com/questions/351557/on-what-linux-distributions-can-i-rely-on-the-presence-of-etc-os-release

prtyag: yes in the RHEL and SUSE this will always be present i'll mention this in the public document.

PR #341, opened 2020-03-19
Adding instructions to re-license SAP HANA

dwj300:


lagalbra:
Out of curiosity, where did these steps come from? If we made them, could they instead be a markdown file so we could add details over time more easily?

dwj300:
@Laura, they came from Ralf, and I will work to move them to a MD (although its hard with images) and eventually post them on docs.microsoft.com

PR #352, opened 2020-03-24
Fix a bug in resume download

rukawata:
There is a bug in the current downloader where SMP returns error 416 Range Not Satisfiable.


deploy/v2/python/downloader/SAP_DLM.py
```diff
@@ -152,14 +152,14 @@ def download(self):
 
         # Second request to download the file from new or resume
         while True:
-            resume_header = ({'Range': 'bytes={self.last_pos}-'})
+            resume_header = ({'Range': 'bytes=%d-' % self.last_pos})
             resp  = DLM.sess.get(DLM.url_token, params=payload, timeout=resp_timeout_sec, stream=True, headers=resume_header)
             if self.last_pos > 0:
                 print("Resume at last_pos %s" % self.last_pos)
-            if resp.status_code == 200:
+            if resp.status_code in [200, 206]:
```
tniek:
```suggestion
            if resp.status_code >= 200 and resp.status_code < 300:
```
(to allow other successful HTTP responses)

deploy/v2/python/downloader/SAP_DLM.py
```diff
@@ -152,14 +152,14 @@ def download(self):
...
                 break
             print("Received status code %d -> retrying..." % resp.status_code)
-        assert(resp.status_code == 200), \
+        assert(resp.status_code in [200, 206]), \
```
tniek:
```suggestion
        assert(resp.status_code >= 200 and resp.status_code < 300), \
```

Issue #287, opened 2020-02-19
Bug: Addition of Prometheus changes break existing deployments


ben-moss:
### Summary

The latest changes made to the `master` repo in #286 seem to have broken deployments that were working prior to that PR's merge.

### Steps to Reproduce

1. Checkout and use v2 code prior to #286 (e.g. `git checkout 94fac2a53677ccf3e0f2680bf2e2311535e77a72`)
1. Do all the basic setup, like SP, Terraform Init, etc. (if not done already)
1. Ensure you have no local changes/config in files, other than SAP media download credentials set in example template
1. From the project root, run the deploy: `terraform apply -auto-approve -var-file=deploy/v2/template_samples/single_node_hana.json deploy/v2/terraform/`

### Expected Behaviour

Deployment completes successfully without error.

### Actual Behaviour

Deployment fails at the Ansible stage, on the following task:

```text
TASK [enable-prometheus : Set OS version] ****************************************************************************************
fatal: [10.1.2.4]: FAILED! => {"msg": "The conditional check 'output.options.enable_prometheus == True' failed. The error was: error while evaluating conditional (output.options.enable_prometheus == True): 'dict object' has no attribute 'options'\n\nThe error appears to be in '/home/azureadm/sap-hana/deploy/v2/ansible/roles/enable-prometheus/tasks/main.yml': line 5, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set OS version\n  ^ here\n"}
```

### Relevant Resources (URLs, screenshots, logs, etc.)

The following is expecting `output.options` to exist, but it doesn't if the customer is using a codebase version prior to #286 (at the time of writing this is the very latest, so unlikely that the customer might have updated):

https://github.com/Azure/sap-hana/blob/4f1795550f25a535c5078a25ce02a3527d1e8fd5/deploy/v2/ansible/sap_playbook.yml#L38

### Possible fixes

1. A trivial fix might be to use something along these lines (⚠️code not tested ⚠️):
   ```
   when:
     - output.options is defined
     - (output.options.enable_prometheus is defined) and output.options.enable_prometheus
   ```
1. A more robust best practice for avoiding similar issues would be to provide defaults for all role variables (i.e. don't expect the calling code to be smart enough to have defined everything - especially in the `output.json`)

ben-moss:
[...]

PR #353, opened 2020-03-24
Update README.MD with prometheus.

rukawata:


secustor:
Could you add the `prometheus_enabled` option to the examples in this PR too? 

We have run in this issue today.

rukawata:
> Could you add the `prometheus_enabled` option to the examples in this PR too?
> 
> We have run in this issue today.

Thanks for your feedback! We added that to all the samples we provided under `deploy/v2/template_samples`.

ben-moss:
I found this back in #287. It really should be defaulted in ansible to prevent future templates having the same problem. FWIW @robfallows also ran into the same problem today.

rukawata:
> I found this back in #287. It really should be defaulted in ansible to prevent future templates having the same problem. FWIW @robfallows also ran into the same problem today.

Yeah. The challenge here is also how and where to maintain all the variables in the input JSON (terraform and ansible share the same variable file aka input JSON).
Since terraform always do a replace instead of merge if the variable is map or object, we can't rely on terraform to take consolidate the customized value with default value here.

We have this [template](https://github.com/Azure/sap-hana/blob/master/templates/tempGen/template-daily.json) used in our daily test, and this [template](https://github.com/Azure/sap-hana/blob/master/templates/tempGen/template.json) used in our PR. Those templates are guaranteed to run fine. So my thinking is I will point to those templates in the README.MD for people who are new to the script to try it out.

Hope this helps.

Thanks,
Nancy

deploy/v2/README.MD
```diff
@@ -241,6 +242,8 @@ And we only include OS configuration for:
 - SLES 12
 - RHEL 7.
 
+<sup>[2](#myfootnote2)</sup>: With [Prometheus](https://prometheus.io/) enabled, you will have more possiblity when you monitor your HANA instance using Azure Monitor for SAP Solutions (preview).
```
persiaAziz: nit: enabling promethus will enrich the set of telemetry even more when you monitor your Hana Instance using Azure Monitor for SAP Solutions

rukawata: Thanks! Addressed this comment.

PR #354, opened 2020-03-24
Fix edge case with rti_only template sample

rukawata:
This PR fix the terraform code when there is nothing under dbnodes (when using [rti_only](https://github.com/Azure/sap-hana/blob/master/deploy/v2/template_samples/rti_only.json) template).


deploy/v2/terraform/modules/hdb_node/main.tf
```diff
@@ -129,7 +129,7 @@ resource "azurerm_network_interface_backend_address_pool_association" "hana-lb-n
 }
 
 resource "azurerm_lb_rule" "hana-lb-rules" {
-  count                          = length(local.loadbalancers[0].ports)
```
persiaAziz: Curious: 
Why is the loadbalancer an array and why we do we always use the [0] index in line 135?
Which scenario has more than one load balancers?

persiaAziz: Please add comment/reference to the load balancer architecture if possible

rukawata: > Curious:
> Why is the loadbalancer an array and why we do we always use the [0] index in line 135?
> Which scenario has more than one load balancers?

This is because we can multiple database instances in one deployment.
And we will have one load balancer per database instance - it can't be shared across instances.

rukawata: > Please add comment/reference to the load balancer architecture if possible

The architecture with loadbalancer is here:
https://github.com/Azure/sap-hana/blob/master/deploy/v2/terraform/README.md

persiaAziz: thanks

PR #357, opened 2020-03-25
Create script to save log files if an event occurred

jeffaco:
This script is required for a customer to capture TCP/IP packets
when a specific error appears in the system log file. When an
NFS timeout occurs, capture packets around the time that the
timeout occurred.


tools/README.md
```diff
@@ -3,6 +3,10 @@ Placeholder for bare metal system tools
 checksystem.sh: Ensures that the bare metal (LI or VLI) system is free
   of common deployment errors.
 
+collect-logs.sh: Save log files as long as an event occurred within a
```
rukawata: The script save tcpdumps if events occur (by comparing count from grep in /var/log/messages), is that correct?
If yes, this is not very clear in the README.md. It would be good if we are more specific which log files (tcpdumps?) are saved, and what event (nfs: server not responding?). Thanks!

jeffaco: The idea, from the functions and from the variables up top, is that this is reasonably easy to change for other purposes. I have no problems making README.md more specific, though. Fixed.

ben-moss:
It's minor @jeffaco , but shellcheck v0.7.0 highlights the following:

```text
$ shellcheck tools/*.sh

In tools/collect-logs.sh line 80:
  rm -rf ${DUMP_DIR}/*
         ^-----------^ SC2115: Use "${var:?}" to ensure this never expands to /* .
```

tools/collect-logs.sh
```diff
@@ -0,0 +1,121 @@
...
+SEARCH_FILTER="nfs: server .* not responding, still trying"
+
+TCPDUMP_OUTPUT="${DUMP_DIR}/tcpdump.out"
+
+# Function definitions
+
+function start_jobs() {
+  # Start collecting our data
+
+  /usr/sbin/tcpdump -i any -w "${TCPDUMP_OUTPUT}" 2>/dev/null &
```
persiaAziz: Why any interface? Can't we just limit to the interface that talks with nfs server? You will get a whole lot more unnecessary logs if there are multiple interfaces.

jeffaco: Changed to take this by parameter. Needs to be edited by the customer before use, but it is "any" by default so it can at least run.

persiaAziz: Can we then pass this as a parameter to the shell file like : TCPDUMP_INTERFACE = $1? Instead of having the user change it in the file?

persiaAziz: Talked offline. I am still not a fan of using any interface as default. But going to sign off anyway

PR #368, opened 2020-03-27
Fixes #348 by allowing RG names to be set via env var for util scripts

### LINKED ISSUE #348: Automate Unique Resource Group Names with Utility Scripts
## Problem Statement

The `util` scripts provide a simplified interface into using and testing the codebase without over-customization for the user (typically an Engineer). This should allows multiple users to run the same deployments and get the same (or expectedly similar) results. Currently there is an issue when two or more users attempt to deploy into the same subscription, because the templated input JSON files contain hard-coded details - in particular the resource group name, which needs to be unique within an Azure subscription.

## Enhancement

The `util` scripts should be enhanced to support editing the resource group name through a script rather than manual editing (in a similar manner to `util/set_sap_download_credentials.sh`). This could then be either run by an engineer, or embedded as part of the `terraform_v2.sh` wrapper script. Through this, deployments can auto-generate a more-unique resource group name from, for example, properties of the local environment (i.e. environment variables). This follows the existing behaviour of the `util/terraform_v2.sh` script, such that:

When the user runs `util/set_resource_group.sh <resource group name> <target template>`
Then the script:
- [x] Sets the given resource group name in the given template
- [x] Fails with a suitable error when either no resource group name or no target template are provided or the given template does not exist
- [x] When a non-valid set of inputs is provided, the script should provide a list of valid <target template> files based on those available in the templates dir. For example:
  ```text
  Where <JSON template name> is one of the following:
  - clustered_hana
  - rti_only
  - single_node_hana
  ```

## Notes

1. The scripts only have to work for the V2 templates in the standard directory (e.g. single_node_hana, clustered_hana, etc.)
1. To aid future refactoring the code should follow the patterns and conventions already laid out for similar functionality in `terraform_v2.sh`.
1. Some valid sets of inputs are:
   - `util/set_resource_group.sh "rg-dev-hana" single_node_hana`
   - `util/set_resource_group.sh "rg-prd-hana" clustered_hana`
1. Some invalid sets of inputs are:
   - `util/set_resource_group.sh "" single_node_hana`
   - `util/set_resource_group.sh "rg"`
1. There's some existing code within the build pipeline setup that solves the same kind of problem to allow multiple pipelines to deploy in parallel (these solution could be consolidated)

## Dependencies

None, but relates to #340 Capability to specify target input JSON for set_sap_download_credentials.sh

## Future Work

1. Generalize the ability to programatically edit the JSON templates for other use cases, such as:
   1. Switching between SLES and RHEL
   1. Setting HANA version
   1. Setting the SID/Instance Number
   1. Switching the ansible execution on/off
1. Integrate the use of these scripts within the pipeline to create more readable pipeline test scenarios

## Checklist

- [x] Usage documentation updated as necessary
- [x] Architecture documentation updated as necessary

## References

None

ben-moss:
## Problem

See #348

## Description

This PR allows the util scripts to automatically set the input JSON's resource group name to the value set in the environment variable `SAP_HANA_RESOURCE_GROUP`.

## Pre-Requisites

`jq` installed

## Test Instructions

1. Start with a clean PR branch checkout, and test with `git status` showing no changes.
1. Unset any potential previous env var `unset SAP_HANA_RESOURCE_GROUP`
1. Run `util/terraform_v2.sh plan rti_only`
1. Check the RTI template hasn't changed with `git diff deploy/v2/template_samples/rti_only.json` (AC1)
1. Run `SAP_HANA_RESOURCE_GROUP="rg-rti-only" util/terraform_v2.sh plan rti_only`
1. Check the RTI template has changed as expected with `git diff deploy/v2/template_samples/rti_only.json` (AC2)

## Acceptance Criteria

Note: The following acceptance testing has been completed in preparation of this PR. See https://github.com/Centiq/sap-hana/pull/17

- [x] AC1: Resourse group name not set in the template if the env var is not set
- [x] AC2: Resource group name is set in the template if the env var is set

### DIFF
--- a/util/terraform_v2.sh
+++ b/util/terraform_v2.sh
@@ -34,26 +34,13 @@ function main()
 	local terraform_action=''
 	[ $# -eq 0 ] || terraform_action="$1"
 
-	# default to empty string when 1 or less args supplied
-	local template_name=''
-	[ $# -le 1 ] || template_name="$2"
-
 	# dispatch appropriate command
 	case "${terraform_action}" in
 		'init')
 			terraform_init
 			;;
-		'plan')
-			check_command_line_arguments_for_template "$@"
-			terraform_plan "${template_name}"
-			;;
-		'apply')
-			check_command_line_arguments_for_template "$@"
-			terraform_apply "${template_name}"
-			;;
-		'destroy')
-			check_command_line_arguments_for_template "$@"
-			terraform_destroy "${template_name}"
+		'plan'|'apply'|'destroy')
+			dispatch_terraform_template_action "$@"
 			;;
 		'clean')
 			terraform_clean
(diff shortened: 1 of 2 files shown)


util/terraform_v2.sh
```diff
@@ -34,37 +34,67 @@ function main()
...
+# If the environment variable is empty/undefined, then it leaves the resource group name unchanged
+function configure_resource_group_for_template()
+{
+	local template_name="$1"
+
+	# default environment variable to empty string if not set
+	local rg_name_from_env="${SAP_HANA_RESOURCE_GROUP-}"
+
+	if [[ "${rg_name_from_env}" != "" ]]; then
+		util/set_resource_group.sh "${rg_name_from_env}" "${template_name}"
```
rukawata: I understand we are not changing `set_resource_group.sh` in this PR.
Just by using that script in this context, I feel it would be good if we can add an echo in `set_resource_group.sh`, like: `"Resource group set to xxx"`, so it is clearer that this has been done.

ben-moss: Great suggestion - now done here: https://github.com/Azure/sap-hana/pull/368/files#diff-363563c695e0f566f91d1c0b026496f7R95-R98

PR #393, opened 2020-04-08
Change hardcode path to dynamic path

YunziZhangAzure:
1. Update path of python/downloader to make Ansible scripts portable
2. Append {sid} to path of hana client install and hana_install_path
3. change output.json path from hardcode way to dynamic way


deploy/v2/ansible/group_vars/all.yml
```diff
@@ -5,14 +5,14 @@ azure_files_mount_path: "/sapmnt"
 hdb_comp_list: []
 linux_comp_list: []
 windows_comp_list: []
-hana_install_path: "/hana/shared/install"
+hana_install_path: "/hana/shared/{{hana_database.instance.sid}}/install"
 hana_software_loc: "{{ azure_files_mount_path }}/DB/DATA_UNITS"
 components: "{{ lookup('file', '../components.json') }}"
-hana_client_install_path: "/usr/sap/shared"
+hana_client_install_path: "/usr/sap/{{ hana_database.instance.sid }}"
```
rukawata, line 11: HANA client normally are shared for several SID.
Therefore make not too much sense to put the installer under SID.
Any reasons for doing that?

deploy/v2/ansible/group_vars/all.yml
```diff
@@ -5,14 +5,14 @@ azure_files_mount_path: "/sapmnt"
...
 output: "{{ hostvars['localhost'].output }}"
 hana_database: "{{ hostvars['localhost'].hana_database }}"
 linux_jumpboxes: "{{ hostvars['localhost'].output.jumpboxes.linux }}"
 windows_jumpboxes: "{{ hostvars['localhost'].output.jumpboxes.windows }}"
-sap_media_download_script_loc: "/home/{{ansible_user}}/sap-hana/deploy/v2"
+#sap_media_download_script_loc: "/home/{{ansible_user}}/sap-hana/deploy/v2"
```
rukawata, line 16:
```suggestion
```
Do not need to keep the commented line :) Unless there is a reason to keep it as comment.

deploy/v2/ansible/group_vars/all.yml
YunziZhangAzure: updated in the new commit.

deploy/v2/ansible/group_vars/all.yml
YunziZhangAzure: updated in the new commit.

PR #415, opened 2020-04-15
Add TODO's to logically seperate code.

mkdeegan:
Added TODO Comments to logically separate code into logical units for RTI.
Added vm-rti.tf as placeholder to receive the identified code.


deploy/v2/terraform/modules/jumpbox/main.tf
```diff
@@ -79,6 +79,13 @@ resource "azurerm_network_interface" "nic-windows" {
   }
 }
 
+/*-----------------------------------------------------------------------------8
+TODO 20200414-MKD: Move Resource block
+
+  INTO: vm-rti.tf
+
+  resource "azurerm_public_ip" "public-ip-linux"
```
rukawata, line 87: resource block for pip, nic, nsg, and vm can't be removed since they are also used to create other jumpboxes that are not rti.
The way this can be done is:
- make rti not one of the jumpboxes
- copy code that's shared between jumpboxes and rti
- remove code that is only used by rti to rti.tf

mkdeegan:
Yes, that is the appropriate methodology to accomplish the task. 
Also the object ID for the resources should altered to a more efficient naming convention. If your creating a public ID , that is in the resource name… so it is redundant to have it in the object ID… so the object ID should just be “rti"

Issue #339, opened 2020-03-18
Capability to manage Azure Fencing Agent


ben-moss:
## Problem Statement

Clustering requires an Azure fencing agent to be created to manage the VMs in a cluster should one node need to notify Azure that a STONITH activity needs to take place. Azure fencing agents are implemented with a service prinicpal in Azure AD, and therefore the process to provision/destory a cluster, in turn, requires that a corresponding service principal is provisioned/destoryed in Azure AD. This might require a different (often higher) set of privileges to those typically required in provisioning/destroying Azure resources, and so bundling this process with the main Azure resource provisioning and configuration for clusters is not always appropriate.

## Enhancement

Ensure that the codebase can support activities to provision/destroy an appropriate Azure fencing agent for HANA cluster management.
This will be demonstrated by introducing appropriate utility scripts such that:

Given the user has the appropriate permissions
When the user runs `util/create_fencing_agent.sh <SID>`
Then the script:
- [x] Creates the fencing agent service principal for the given SAP HANA SID with the appropriate permissions
- [x] Assigns the appropriate custom role to the service principal
- [x] Stores the service principal details in a local file `fencing-agent-<sid>.sh` (similar to #288 ) so they can be used when configuring the clustering software when the VMs are configured (Terraform will copy the file to the RTI for ansible to use in configuring the cluster)
- [x] Fails with a suitable error when the user does not have the correct permissions to create service principals
- [x] Fails with a suitable error when no SID or an existing SID (matching local file) is given

## Notes

1. At the current time, no reference architecture is available for HA clustering of SLES on Azure using the Azure fencing agent STONITH method
1. See the [**Important** note in Create SAP HANA cluster resources](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability#create-sap-hana-cluster-resources)
1. **Assumption:** The entity relationship between the clustering service principal and a clustered SAP system/input JSON is 1:1, rather than reusing the same fencing agent across all SAP systems in a subscription. However, it's technically possible to have a single fencing agent responsible for all SAP systems - even across subscriptions.

## Dependencies

- [x] #337 Capability to provision Azure resources for clustered HANA systems on SLES

## Future Work

1. [Manage the clustering service principal through Terraform](https://www.terraform.io/docs/providers/azuread/r/service_principal.html)
1. Use a Key Vault to store the clustering SP credentials

## Checklist

- [x] Usage documentation updated as necessary
[...]

Issue #19, opened 2018-08-15
sap_instancenum not considered at provisioning


tniek:
**Description:** During hdblcm install via Ansible, the specified instance number is not considered.

**How to reproduce:** In terraform.tfvars, set sap_instancenum = "01"; after deployment, there will be install with instance number 00 instead.

PR #421, opened 2020-04-15
Azure fencing agent management

centiqmurphy:
# Problem

As part of #339 Clustering requires an Azure fencing agent to be created to manage the VMs in a cluster.

## Description

This PR:

When the user runs `util/create_fencing_agent.sh <SID>`
Then the script:

- [x] Creates the fencing agent service principal for the given SAP HANA SID with the appropriate permissions.
- [x] Assigns the appropriate custom role to the service principal.
- [x] Stores the service principal details in a local file `fencing-agent-<sid>.sh` (similar to #288 ) so they can be used when configuring the clustering software when the VMs are configured (Terraform will copy the file to the RTI for ansible to use in configuring the cluster).
- [x] Fails with a suitable error when the user does not have the correct permissions to create service principals.
- [x] Fails with a suitable error when no SID or an existing SID (matching local file) is given.
- [x] Runs on both Linux and Mac OS (as per other util scripts).

The Terraform has been updated to copy the generated script to the RTI VM. If clustering is not required, an empty file will be generated on the RTI.

## Pre-Requisites

Workstation with `az cli` and access to the Centiq Microsoft Azure Sponsorship 

## Commitment to Test

- Testing takes 15 minutes
- Reviewing 2 files 

## Test Instructions and Acceptance Criteria
✋ The following testing was performed in [PR #19 on the Centiq Fork](https://github.com/Centiq/sap-hana/pull/19)
### Test case 1: Fails when value for SID is not provided

Run `util/create_fencing_agent.sh`

- [ ] `ERROR: You must specify a single command line argument for the SAP SID. For example: ./util/create_fencing_agent.sh HN1`

### Test case 2: Completes without error when SID is provided

Run `util/create_fencing_agent.sh B4Z`

- [ ] Script completes with `A service principal has been created in Azure > App registrations, with the name: fencing-agent-B4Z`

 Example output:

   ```text
   Creating Azure Service Principal: fencing-agent-B4Z...
   Changing "fencing-agent-B4Z" to a valid URI of "http://fencing-agent-B4Z", which is the required format used for service principal names
   Found an existing application instance of "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx". We will patch it
   Creating a role assignment under the scope of "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
      Role assignment already exits.

   Role definition already exists
   Role assignment already exists
   A service principal has been created in Azure > App registrations, with the name: fencing-agent-B4Z
   Azure authorization details can be found within the script: set-clustering-auth-B4Z.sh
   The Azure authorization details are copied to the RTI during Terraform provisioning for usage by Ansible.
   ```
  
✋ If you have try to create a SP already created by another user the following error will display:

- [ ] `'http://fencing-agent-<SID>' already exists.
ERROR: There was a problem creating the service principal. If you are logged into Azure CLI, then it could relate to lack of admin/owner permissions. See https://docs.microsoft.com/en-us/cli/azure/create-an-azure-service-principal-azure-cli for further details.`

### Test case 3: SP already exists in Azure AD

Run `util/create_fencing_agent.sh B4Z`

- [ ] ERROR: Authorization file already exists: set-clustering-auth.sh. Please reuse, move, or remove it. is displayed.

### Tidy up

Run `az ad sp list --show-mine`

- SP `fencing-agent-B4Z` with `"adminConsentDisplayName": "Access fencing-agent-B4Z"` is listed.

- Delete SP `az ad sp delete --id < objectId >'

- Delete fencing agent auth script `rm set-clustering-auth-B4Z.sh`


util/create_fencing_agent.sh
```diff
@@ -0,0 +1,131 @@
...
+#######################################################################################################################################################
+
+# exit immediately if an unset variable is used
+set -o nounset
+
+# import common functions that are reused across scripts
+source util/common_utils.sh
+
+# name of the script where the auth info will be saved
+readonly auth_script="set-clustering-auth-${1}.sh"
```
rukawata, line 19: Maybe consider giving default value here eg:
```suggestion
readonly auth_script="set-clustering-auth-${1-HN1}.sh"
```
Other wise, just run util/create_fencing_agent.sh given an unclear error.
```
:sap-hana nancyc$ util/create_fencing_agent.sh
util/create_fencing_agent.sh: line 19: 1: unbound variable
```

util/create_fencing_agent.sh
```diff
@@ -0,0 +1,131 @@
...
+{
+	check_command_line_arguments "$@"
+
+	local SID="${1-}"
+
+	create_service_principal_script "${SID}"
+}
+
+
+function check_command_line_arguments()
```
rukawata, line 34: When we have default value defined [here](https://github.com/Azure/sap-hana/blob/956d19823822684ecc9072ef35d9064dd87adeef/util/create_fencing_agent.sh#L19), it should proceed with the default value.

util/create_fencing_agent.sh
```diff
@@ -0,0 +1,131 @@
...
+
+	# replace original JSON template file with temporary filtered one
+	mv "${temp_template_json}" "${template_file}"
+
+	# ensure role definition exsits
+	local fencing_template='Linux Fence Agent Role'
+	local sp_prefix='http://'
+	role_list="$(az role definition list --name 'Linux Fence Agent Role')"
+	if [[ "${role_list}" == "[]" ]]; then
+		az role definition create --role-definition "${template_file}"
```
rukawata, line 102: 1. How about error handling when role creation fail?
2. Can we move the role creation before the SP creation? If the role creation fail, it does not make sense to continue with the SP creation.

util/create_fencing_agent.sh
rukawata, line 19: Can we change the location of this sh to some tmp folder? Or even a tmp folder under util. 
So it's not mixed with other github code.

util/create_fencing_agent.sh
```diff
@@ -0,0 +1,131 @@
...
+	assignment_list="$(az role assignment list --assignee "${sp_prefix}${service_principal_name}" --role "${fencing_template}")"
+	if [[ "${assignment_list}" == "[]" ]]; then
+		az role assignment create --assignee "${sp_prefix}${service_principal_name}" --role "${fencing_template}"
+	else echo "Role assignment already exists"
+	fi
+
+	IFS="${ifs_backup}"
+
+	echo "A service principal has been created in Azure > App registrations, with the name: ${service_principal_name}"
+	echo "Azure authorization details can be found within the script: ${auth_script}"
```
rukawata, line 115: can we print the abs path of the shell script? So it is clearer.

deploy/v2/USAGE.md
```diff
@@ -132,7 +132,37 @@ This script can then be used (_sourced_) to configure the required environment v
    The Azure authorization details are automatically used by the utility scripts if present.
    ```
 
-   **Note:** The generated authorization script contains secret information, which you should store and secure appropriately.
+   **Note:** If you are provisioning a clustered system, then you also need to create a fencing agent service principal
+
+2. To easily create the service principal and authorization script, run the following command providing the HANA SID you wish to be included in the service principal name as the only command line argument (here the SID `T0D` is used):
+
+   ```text
+   util/create_fencing_agent.sh T0D
```
rukawata, line 140: In the [code](https://github.com/Azure/sap-hana/blob/956d19823822684ecc9072ef35d9064dd87adeef/util/create_fencing_agent.sh#L40), it gives example with HN1. Please make sure it's consistent here.

util/create_fencing_agent.sh
ben-moss: The default should be "" (empty string) to avoid the error, but stay consistent with other util script behaviour. Otherwise it should read the SID from the template (although there is no active/specified template in this context).

I gave this feedback previously, but it got missed/regressed.

util/create_fencing_agent.sh
ben-moss: That would be inconsistent with the other util script behaviour and the specification in #339. I think it should be a backlog item if it's really required, but I suggest this is better managed some other way in the future (e.g. separate defaults config file/input.json).

util/create_fencing_agent.sh
ben-moss: Let's not wrap too much what the `az` will already do, but agree that we could manage non-zero exit status and not continue at that point. Perhaps better done with `set -o errexit`, though

util/create_fencing_agent.sh
ben-moss: Original implementation put it in `util` but I asked for it to go alongside `set-sp.sh` for consistency. The script is not really temporary, it contains similar info that might eventually go into a key vault, so not keen to put in any kind of `tmp` directory.

util/create_fencing_agent.sh
ben-moss: Currently this is consistent with the other util scripts. If really required this should be a backlog item that covers all the scripts.

deploy/v2/USAGE.md
ben-moss: Good catch. I think the doc really needs some changes to put this section in with the system build, as that's where it belongs. To be consistent with the rest of the doc it should probably document usage with a placeholder (e.g. `<sid>`) rather than a specific example.

util/create_fencing_agent.sh
rukawata: It's fine if all files are under `sap-hana/util`, however the `set-clustering-auth-<SID>.sh` is under `sap-hana`.
Can you please change that?

ben-moss: This means retesting a bunch of script scenarios again, and some of those go beyond the scope of this PR. Please can it go on the backlog rather than hold this up any longer? :)

util/create_fencing_agent.sh
ben-moss: Lots more error handling now added @rukawata . Thanks for highlighting this 👍

util/create_fencing_agent.sh
ben-moss: Now captured on backlog in #427

util/create_fencing_agent.sh
ben-moss: Now captured on backlog in #427

deploy/v2/terraform/main.tf
```diff
@@ -92,6 +92,7 @@ resource "null_resource" "ansible_playbook" {
       "curl -i -H \"Metadata: \"true\"\" -H \"user-agent: SAP AutoDeploy/${var.auto-deploy-version}; scenario=${var.scenario}; deploy-status=Terraform_finished\" http://169.254.169.254/metadata/instance?api-version=${var.api-version}",
       "export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES",
       "export ANSIBLE_HOST_KEY_CHECKING=False",
+      "source ~/export-clustering-sp-details.sh",
```
rukawata, line 95: source should be done in `deploy/v2/terraform/modules/jumpbox/main.tf` -> `"null_resource" "prepare-rti"`.
Documented in backlog item #428.

util/create_fencing_agent.sh
```diff
@@ -0,0 +1,158 @@
...
+	local sp_uri="http://${service_principal_name}"
+
+	local role_list=''
+	role_list=$(az role definition list --name "${fencing_template}")
+	local role_list_status=$?
+	continue_or_error_and_exit $role_list_status "There was a problem determining if the Fencing Agent role already exists. If you are logged into Azure CLI, then it could relate to lack of network connectivity."
+
+	# check if fencing agent role already exists and create it if not
+	if [[ "${role_list}" == "[]" ]]; then
+		az role definition create --role-definition "${template_file}"
```
rukawata, line 77: The `template_file` definition is now in line 126 and script gave error:
`util/create_fencing_agent.sh: line 77: template_file: unbound variable`.
I create bug for that.

ben-moss: Sorry about that (late night!) I also noticed a few other things to tidy up this morning.

PR #422, opened 2020-04-15
add pre-check of os-disk-setup

YunziZhangAzure:
1. restructure os-disk-setup role
2. add precheck to os-disk-setup role, if installation paths all exist, volume group, LVG, mount disk will be skipped.
3. If the scenario is bare metal, creating SWAP is skipped.
4. When creating Logical volume group, every time LVG will occupy the whole volume group.

### DIFF
--- a/deploy/v2/ansible/roles/os-disk-setup/tasks/installation_path_prechecks.yml
+++ b/deploy/v2/ansible/roles/os-disk-setup/tasks/installation_path_prechecks.yml
@@ -1,0 +1,15 @@
+---
+
+# Check if installation paths are correct. 
+- name: Pre-check installation path
+  stat:
+    path: "{{ item.value.mount_point }}"
+  loop: "{{ disk_dict|dict2items }}"
+  register: path_status
+
+# If any one of paths is incorrect, path_status_flag will be defined.
+- name: Get intallation path status
+  set_fact:
+     path_status_flag: "{{ item.stat.exists }}"
+  loop: "{{ path_status.results }}"
+  when: item.stat.exists == False
--- a/deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
+++ b/deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
@@ -13,47 +13,8 @@
   loop: "{{ hdb_sizes[hana_database.size].storage }}"
   when: item.name != 'os'
 
-- name: Create physical volume group
-  lvg: 
-    vg: "vg_{{ item.key }}"
-    pvs: "{{ item.value.devices }}"
-  with_dict: "{{ disk_dict }}"
-  register: physical_vg
-  failed_when: physical_vg is failed
-  ignore_errors: true
+# Check if installation paths are ready. If paths are unready, path_status_flag will be set. Then subsequent tasks will be executed.
+- import_tasks: installation_path_prechecks.yml
 
-- name: Create logical volume group
-  lvol:
-    state: present
-    vg: "vg_{{ item.key }}"
-    lv: "lv_hana_{{ item.key }}"
-    size: 100%FREE
-  with_dict: "{{ disk_dict }}"
-  register: logical_vg
-  failed_when: logical_vg is failed
-  ignore_errors: true 
- 
-- name: Create mount points
-  file:
-    path: "{{ item.value.mount_point }}"
-    state: directory
-  with_dict: "{{ disk_dict }}"
-
-- name: Create file systems
-  filesystem:
-     fstype: "{{ hana_database.filesystem | default('xfs') }}"
-     dev: "/dev/vg_{{ item.key }}/lv_hana_{{ item.key }}"
-  with_dict: "{{ disk_dict }}"
-  register: file_system
-  failed_when: file_system is failed
-
-- name: Mount volumes
-  mount:
-    path: "{{ item.value.mount_point }}"
-    fstype: "{{ hana_database.filesystem | default('xfs') }}"
-    opts: "defaults,nofail"
-    src: "/dev/vg_{{ item.key }}/lv_hana_{{ item.key }}"
-    state: mounted
-  with_dict: "{{ disk_dict }}"
-  register: vol_mount
-  failed_when: vol_mount is failed
+- include_tasks: lvm_mount.yml
+  when: path_status_flag is defined
--- a/deploy/v2/ansible/roles/os-preparation/tasks/common/os_config.yml
+++ b/deploy/v2/ansible/roles/os-preparation/tasks/common/os_config.yml
@@ -19,17 +19,22 @@
 - name: Task Disable Transparent Hugepages & Configure Processor C-States 3
   command: grub2-mkconfig -o /boot/grub2/grub.cfg
 
-- name: Create SWAP space
-  lineinfile:
-    dest: /etc/waagent.conf
-    regexp: "^{{ item.prop }}="
-    line: "{{ item.prop }}={{ item.value }}"
-  with_items:
-    - { prop: "ResourceDisk.Format", value: "y" }
-    - { prop: "ResourceDisk.EnableSwap", value: "y" }
-    - { prop: "ResourceDisk.SwapSizeMB", value: "{{ (hdb_sizes[hana_database.size].compute.swap_size_gb) * 1024|int }}" }
+# If current host is LargeInstance, creating swap will be skipped. 
+- name: Prepare SWAP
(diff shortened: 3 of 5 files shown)


deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
```diff
@@ -13,6 +13,14 @@
   loop: "{{ hdb_sizes[hana_database.size].storage }}"
   when: item.name != 'os'
 
+# Check if filesystem and mounting point are correct. If correct, subsequent tasks will be skipped
+- name: Pre-check current filesystem status
```
rukawata, line 17: How about using `stat` module in Ansible?

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
```diff
@@ -13,6 +13,14 @@
...
+  command: df -h | grep 'vg_{{ item.key }}-lv_hana_{{ item.key }}' | awk '{print $6}' | grep '{{ item.value.mount_point}}' 
+  with_dict: "{{ disk_dict }}"
```
rukawata, line 19: Use loop so it's up-to-date.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
```diff
@@ -13,6 +13,14 @@
...
+  register: filesystem_mount_status
+  failed_when: filesystem_mount_status is failed
```
rukawata, line 21: When you loop through dict, the content are in the results, eg. `filesystem_mount_status.results`.
Also another thing to consider is, if only part of the mount points are configured but not all, how should we handle?
IMO, in that situation, to continue will also fail since it's partially done/not done.
Therefore, the check should be, we only proceed with the rest tasks if none mount points in the dict is configured.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
```diff
@@ -21,23 +29,29 @@
...
+  when:
+    filesystem_mount_status is failed
 
 - name: Create logical volume group
   lvol:
     state: present
     vg: "vg_{{ item.key }}"
     lv: "lv_hana_{{ item.key }}"
-    size: 100%FREE
+    size: 100%VG
```
rukawata, line 40: Any reason to change to `VG`?

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
```diff
@@ -21,23 +29,29 @@
...
   with_dict: "{{ disk_dict }}"
   register: logical_vg
   failed_when: logical_vg is failed
-  ignore_errors: true 
+  ignore_errors: true
+  when:
+    filesystem_mount_status is failed
```
rukawata, line 46: Instead of having the failed condition several times, you can use `block` to have all the disk configuration tasks together.
Then you only have to add condition once :)

deploy/v2/ansible/roles/os-preparation/tasks/common/os_config.yml
```diff
@@ -28,8 +35,12 @@
     - { prop: "ResourceDisk.Format", value: "y" }
     - { prop: "ResourceDisk.EnableSwap", value: "y" }
     - { prop: "ResourceDisk.SwapSizeMB", value: "{{ (hdb_sizes[hana_database.size].compute.swap_size_gb) * 1024|int }}" }
+  when:
+    swap_status is failed
```
rukawata, line 39: See above comment about using `block`.

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -1,12 +1,13 @@
...
 
 - hosts: localhost
+  connection: local
   become: true
   become_user: root
   pre_tasks:
     - name: Include output JSON
       include_vars:
-        file: "/home/{{ ansible_user }}/output.json"
+        file: "{{ inventory_dir }}/output.json"
```
rukawata, line 10: this is from a different PR correct?

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
YunziZhangAzure: Going to use stat.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
YunziZhangAzure: got it.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
YunziZhangAzure: Will use registered result.

The logic will be "If any one of these paths doesn't exist, subsequent tasks will be executed". To guarantee this "idempotency", the parameter "size" of logical volume will be changed to 100%VG, meaning the logical volume should occupy the whole space of volume group every time the task will be executed. This solution can apply to the scenario that some paths don't exist.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
YunziZhangAzure: The logic will be "If any one of these paths doesn't exist, subsequent tasks will be executed". To guarantee this "idempotency", the parameter "size" of logical volume will be changed to 100%VG, meaning the logical volume should occupy the whole space of volume group every time the task will be executed. This solution can apply to the scenario that some paths don't exist.

deploy/v2/ansible/roles/os-disk-setup/tasks/main.yml
YunziZhangAzure: Good idea, thanks!

deploy/v2/ansible/sap_playbook.yml
YunziZhangAzure: Forgot to include this change in last pr.

YunziZhangAzure: Will leave this change to future

deploy/v2/ansible/roles/os-disk-setup/tasks/installation_path_prechecks.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+# Check if installation paths are correct. 
+- name: Pre-check installation path
+  stat:
+    path: "{{ item.value.mount_point }}"
+  loop: "{{ disk_dict|dict2items }}"
+  register: path_not_exist
```
rukawata, line 8: How about this?

```suggestion
  register: path_status
```

deploy/v2/ansible/roles/os-disk-setup/tasks/installation_path_prechecks.yml
```diff
@@ -0,0 +1,15 @@
...
+
+# If any one of paths is incorrect, path_unready_status_flag will be defined.
+- name: Get intallation path status
+  set_fact:
+     path_unready_status_flag: "{{ item.stat.exists}}"
```
rukawata, line 13: Also nit suggestion:
```suggestion
     path_status_flag: "{{ item.stat.exists}}"
```

deploy/v2/ansible/roles/os-preparation/tasks/common/os_config.yml
```diff
@@ -19,17 +19,22 @@
...
-    dest: /etc/waagent.conf
-    regexp: "^{{ item.prop }}="
-    line: "{{ item.prop }}={{ item.value }}"
-  with_items:
-    - { prop: "ResourceDisk.Format", value: "y" }
-    - { prop: "ResourceDisk.EnableSwap", value: "y" }
-    - { prop: "ResourceDisk.SwapSizeMB", value: "{{ (hdb_sizes[hana_database.size].compute.swap_size_gb) * 1024|int }}" }
+# If current host is bare metal, creating swap will be skipped. 
+- name: Prepare SWAP
+  when: hana_database.size != "bare_metal"
```
rukawata, line 24: LargeInstance is probably more properate? And in the current sizing, we don't use symbol. So let's maintain it that way :)
```suggestion
  when: hana_database.size != "LargeInstance"
```

deploy/v2/ansible/roles/os-disk-setup/tasks/installation_path_prechecks.yml
YunziZhangAzure: Good catch thanks!

PR #438, opened 2020-04-19
Update mount azure file cases to apply to LI scenario

YunziZhangAzure:
1. when host = hanadbnodes and scenario = LI, mounting azure files will be skipped
2. Update TF ansible_inventory.tmpl so that localhost will use "local" connection method by default instead of ssh.
3. Add {sid} to hana_install_path


ben-moss:
@YunziZhangAzure - what does LI stand for?

YunziZhangAzure:
Large Instance. :)

PR #455, opened 2020-04-27
Add check of installation paths in Large Instance

YunziZhangAzure:
1. unzip is preinstalled on most images. No need to use the lastest version.
   And for Large Instance, there is no internet available, set state =
   latest may require downloading through internet.
2. Add check of paths status in Large Instance, if not all paths are
   ready, playbook fails instantly. Because there is nothing we can do
   if they are not ready.

### DIFF
--- a/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
+++ b/deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
@@ -6,7 +6,7 @@
 - name: Install unzip
   package: 
     name: unzip
-    state: latest
+    state: present
 
 # Explicitly install bs4 with python3
 - name: Install beautifulsoup4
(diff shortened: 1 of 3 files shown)


deploy/v2/ansible/roles/sap-media-download/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
 - name: Install unzip
   package: 
     name: unzip
-    state: latest
+    state: present
```
rukawata, line 9: What is the problem with latest?

YunziZhangAzure: When state = latest, it will detect if there is a latest version of unzip. If the current version is not the latest, it will try to pull a latest one from repository. Since large instance doesn't have internet access, it will prompts an error here.

PR #457, opened 2020-04-28
Ignore .DS_Store from MacOS

rukawata:


.gitignore
```diff
@@ -25,3 +25,4 @@ set-sp.sh
 
 # ignore any clustering auth scripts
 set-clustering-auth*.sh
+.DS_Store
```
tniek, line 28: Not needed?

rukawata: No. I will just do that on my local machine.

PR #478, opened 2020-05-08
Add ansible to configure SBD fencing

YunziZhangAzure:
1. Add a role iscsi-target-install to set up iSCSI target servers and
   create iSCSI device on iSCSI target server. Since this iSCSI can be
   used in multiple clusters(Database, ASCS, nfs etc.) through
   configuration, this role is designed to be generic.
2. Add SBD setup yml files to set up sbd device on hanadbnodes when linux distribution is Suse.
3. Azure fencing agent will be bypassed when it is large instance.
4. Add azure scheduled events to allow time for the application to prepare
   for events like VM shutdown, VM redeployment, etc.

### DIFF
--- a/deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
+++ b/deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
@@ -53,7 +53,7 @@
     dest: /etc/corosync/corosync.conf
     mode: 0600
 
-- name: Ensure the Corosync service is restarted
+- name: Ensure the Corosync service is restarted on primary node.
   when: inventory_hostname == hana_database.nodes[0].ip_admin_nic
   systemd:
     name: corosync
--- a/deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
+++ b/deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
@@ -6,6 +6,11 @@
 - name: Ensure a list of package version is available for checking the cloud-netconfig-azure version
   package_facts:
 
+# If iSCSI devices are provisioned and the os is Suse, SBD will be set up
+- name: Ensure SBD devices are set up on hanadbnodes when provided
+  when: groups['iscsi'] | length > 0
+  include_tasks: sbd_device_setup.yml
+
 # eth1 is the "db" NIC
 - name: Ensure clustering can manage Virtual IPs on the Database Interface
   when: ansible_facts.packages["cloud-netconfig-azure"] and (ansible_facts.packages["cloud-netconfig-azure"][0].version | float) < 1.3
@@ -17,7 +22,7 @@
 
 # https://rpm.pbone.net/index.php3/stat/45/idpl/27916721/numer/8/nazwa/ha-cluster-init
 - name: Ensure Primary node initiates the Cluster
-  when: ansible_hostname == hana_database.nodes[0].dbname
+  when: inventory_hostname == hana_database.nodes[0].ip_admin_nic
   block:
     - name: Ensure csync2 is configured
       shell: crm cluster init -y csync2 --interface eth1
@@ -29,7 +34,7 @@
       shell: "crm cluster init -y cluster --name 'hdb_{{ hana_database.instance.sid | upper }}' --interface eth1"
 
 - name: Ensure Secondary node joins the Cluster
-  when: ansible_hostname == hana_database.nodes[1].dbname
+  when: inventory_hostname == hana_database.nodes[1].ip_admin_nic
   block:
     - name: Ensure the configuration files are synchronised
       shell: "crm cluster join -y -c {{ hana_database.nodes[0].ip_db_nic }} csync2 --interface eth1"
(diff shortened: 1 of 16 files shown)


deploy/v2/ansible/sap_playbook.yml
```diff
@@ -71,6 +71,18 @@
       tasks_from: media_hanadbnode_copy.yml
     when: hana_database.size == "LargeInstance"
 
+# Red Hat doesn't support SBD on cloud platforms.
```
rukawata, line 74: Since the condition says 'Suse', it's more straight forward to say, Only SLES support SBD fencing.

deploy/v2/ansible/roles/iscsi-target-install/tasks/pre_checks.yml
```diff
@@ -0,0 +1,15 @@
+---
+
```
rukawata, line 2: All those pre-checks are conditions to invoke role `iscsi-target-install`, therefore I assume they always holds true?
Maybe reconsider this, whether we want the check in the playbook? or in the task.

deploy/v2/ansible/sap_playbook.yml
```diff
@@ -71,6 +71,18 @@
...
+- hosts: iscsi
+  become: true
+  become_user: root
+  roles:
+   - role: iscsi-target-install
+     when: 
+       - hana_database.high_availability
+       - groups['iscsi'] | length > 0
+       - hostvars[hana_database.nodes[0].ip_admin_nic].ansible_os_family is defined
+       - hostvars[hana_database.nodes[0].ip_admin_nic].ansible_os_family == 'Suse'
```
rukawata, line 84: if ansible_os_family is Suse, it's probably defined :) Is there any condition I missed here?

deploy/v2/ansible/roles/iscsi-target-install/tasks/pre_checks.yml
YunziZhangAzure: After reconsideration, I want to refactor this situation in this way:
For precheck within iscsi-target-install role, I want to check two cases: (1) iscsi device is provisoned. (2) the os distro of iscsi vm is Suse. 
For sap-playbook.yml, I want to check (1) high_availability is enabled (2) the os of hanadbnodes is Suse.

My thought is: the iscsi-target-install role is like a producer. The workflow in sap_playbook.yml is like a consumer. They are independent. 

The precheck of iscsi-target-install should only check the prerequisites of this role. That being said, if these precheck don't pass, the requirements are not met, then the role cannot continue executing. "(1) iscsi device is provisoned. (2) the os distro of iscsi vm is Suse." These two are the minimum requirements as far as I can see for now. We can extend it to more checks if it's needed in the future. 

The sap_playbook.yml as our main workflow, we only want when these two conditions -- "(1) high_availability is enabled (2) The os of hanadbnodes is Suse." are met, it invokes the iscsi-target-install role to install iscsi targets. 

Any comments are welcomed!

deploy/v2/ansible/sap_playbook.yml
YunziZhangAzure: Merge to the last comment.

deploy/v2/template_samples/clustered_hana_with_sbd.json
```diff
@@ -95,7 +95,7 @@
           "sku": "16.04-LTS"
         },
         "authentication": {
-          "type": "password",
+          "type": "key",
           "username": "azureadm",
           "password": "Sap@hana2019!"
```
rukawata, line 100: If we use authentication type key, password is not required and can be removed.

YunziZhangAzure: Ok will do!

deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
```diff
@@ -79,14 +84,50 @@
...
         passwd="{{ sap_hana_fencing_agent_client_password }}"
+    
+    - name: Ensure the STONITH SBD is created
+      when: groups['iscsi'] | length > 0
+      block:
+
+        - name: Check if SBD resource exists
+          shell: crm resource list | grep 'sbd' | awk '{ print $1; }'
+          register: sbd_device
+          failed_when: false
```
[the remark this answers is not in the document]
YunziZhangAzure: Probably not, because this check needs crm already installed, which is executed in cluster-Suse.yml and crm only can be used in Suse (not Redhat)

deploy/v2/ansible/roles/iscsi-target-install/tasks/iscsi_device_create.yml
```diff
@@ -0,0 +1,17 @@
+---
+
+- name: Ensure SBD device for clusters is created
+  vars:
+    cluster_name: "{{ item.key }}"
+    storage_object: sbd{{ cluster_name }}
+    target: "{{ iscsi_object }}.{{ cluster_name }}.local:{{ cluster_name }}"
+    initiator1: "{{ iscsi_object }}.{{ item.value[0] | default(hana_database.nodes[0].dbname) }}.local:{{ item.value[0] | default(hana_database.nodes[0].dbname) }}"
```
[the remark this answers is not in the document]
YunziZhangAzure: Done!

deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
```diff
@@ -49,12 +54,72 @@
     mode: 0600
 
 - name: Ensure the Corosync service is restarted
```
rukawata, line 56:
```suggestion
- name: Ensure the Corosync service is restarted on primary node.
```

deploy/v2/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
```diff
@@ -49,12 +54,72 @@
...
+  when: inventory_hostname == hana_database.nodes[0].ip_admin_nic
+  systemd:
+    name: corosync
+    state: restarted
+
+- pause:
+    seconds: 20
+
+- name: Ensure the Corosync service is restarted
```
rukawata, line 65:
```suggestion
- name: Ensure the Corosync service is restarted on secondary node
```

PR #482, opened 2020-05-13
Add support to append missing commandline parameters. 

prtyag:
In some of the old SLES12-SP2 image commandline parameters is missing by default because of which kdump service is not working as exptected. So this PR checks the essential commandline parameters and append if not found.

### DIFF
--- a/tools/enable-kdump.sh
+++ b/tools/enable-kdump.sh
@@ -52,6 +52,14 @@ fi
 rpm -q kexec-tools
 ExitIfFailed $? "kxec-tools required to enable kdump, please install"
 
+# check if the GRUB_CMDLINE_LINUX_DEFAULT parameter exist in /etc/default/grub file
+# missing command line options will be appended to GRUB_CMDLINE_LINUX_DEFAULT
+egrep "^GRUB_CMDLINE_LINUX_DEFAULT" /etc/default/grub
+if [[ "$?" == "1" ]]; then # in this case append the parameter to the file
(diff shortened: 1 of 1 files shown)


tools/enable-kdump.sh
```diff
@@ -52,6 +52,14 @@ fi
...
 
+# check if the GRUB_CMDLINE_LINUX_DEFAULT parameter exist in /etc/default/grub file
+# missing command line options will be appended to GRUB_CMDLINE_LINUX_DEFAULT
+egrep "^GRUB_CMDLINE_LINUX_DEFAULT" /etc/default/grub
+if [[ "$?" == "1" ]]; then # in this case append the parameter to the file
+    echo "GRUB_CMDLINE_LINUX_DEFAULT=\"\"" >> /etc/default/grub
+    ExitIfFailed $? "Enable to add GRUB_CMDLINE_LINUX_DEFAULT parameter in /etc/default/grub"
+fi
+
 ReplaceLowHighInGrubFile()
```
lalitb, line 63: I think these steps can be performed as part of new steps added to check for missing parameters. We can remove these parameters before running new steps, which will ensure it will get added with new values ?

prtyag: This can be done but then we need to test this change on the older blade/images that we have already said this script works. since some of the blades i have already given to ops that's why i though of not touching the previous code and add a new functionality which is completely independent of the current code.

PR #484, opened 2020-05-14
Add inventory validation to playbook

rukawata:
## Problem
We observed several bugs with mainly SLES images in the past. Eg:
1. hostname change back and forth.
1. zypper repo not configured.

By adding validation for the HANA VMs, the goal is to:
1. work around those bugs before we start playbook.
1. otherwise fail the playbook early with proper fail message.

## Description
With this PR:
1. Add inventory validation for hostname and repo configurations.
1. Address the problem that we can't run playbook twice if HANA installation is complete.
1. Address some nit issue with HANA version regex.

## Tests:
1. Run playbook against HANA VM where hostname is set incorrect (eg. linux).
> The hostname was corrected.
1. Run playbook against HANA VM where zypper repo is not configured and can't be fixed.
> The playbook failed at that point.
1. Run playbook to install a new HANA database. Then run the same playbook again.
> The HANA database installation task was skipped.


deploy/v2/ansible/roles/components-install/tasks/hana_database.yml
```diff
@@ -24,7 +24,7 @@
   template:
     src: hdbserver_hana2.j2 
     dest: "{{ hana_install_path }}/hdbserver_{{ hana_database.instance.sid }}_install.cfg"
-  when: hana_database.db_version is regex("2.\d{2}.0([4-9]\d|[0-9]\d{2,})") 
+  when: hana_database.db_version is regex("2.\d{2}.0([0-9]\d|[0-9]\d{2,})")
```
YunziZhangAzure, line 27: Maybe '.' should be changed to \\. to skip sequence?

deploy/v2/ansible/roles/inventory-validation/tasks/repo_validation.yml
```diff
@@ -0,0 +1,30 @@
+---
+
+- name: Ensure zypper repo is configured on HANA VMs for SLES
+  when: ansible_os_family == 'Suse'
+  block:
+    - name: Check zypper repo on HANA VMs
+      command: zypper lr
```
YunziZhangAzure, line 7: May I ask what is purpose of this task?
If it aims to resolving the missing package issue, 'zypper lr' may not be effective. Because 'zyppper lr' can be successful in most cases even if there is some packages missing. 
Or you are trying to accomplish other things?

deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
```diff
@@ -2,3 +2,4 @@
 
 - import_tasks: hdb_pre_checks.yml
 - import_tasks: hana_server_installation.yml
+  when: sid_existence_check_result.rc != 0
```
YunziZhangAzure, line 5: Thanks for this work to make installation of hana database idempotent which was always a pain for me when testing. 

but the logic here seems to become 'If hana database exists, then all installation on db nodes will be skipped', but what if the customer installs database at first, then he/she wants to install extra studio or xs at second time. In the logic here, the installation of extra studio/xs will be skipped.

deploy/v2/ansible/roles/inventory-validation/tasks/repo_validation.yml
rukawata: No, this is only to attempt to save the play when no repo is configured at all.
As for certain package/repo missing, I have not yet found a way to include as pre-check.

deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
rukawata: It would be nice to have also installed component checked.
I could parse the installed component using hdblcm.
However the component naming is not consistent with component.JSON.
Therefore I will have to:
1. either introduce a dictionary to map the component name from hdblcm to our JSON
1. or modify the component.JSON

I see the 2nd way is a cleaner solution. What is your opinion @YunziZhangAzure ?

YunziZhangAzure: Ya, the second solution is good.

Another thinking is can these pre checks go into the 'components-install' role? Before installing each component, do a precheck.

Also, I don't know if it is the case that some information in component.json is same for all customers(like the naming you mentioned), maybe we should let these values go underneath the default folder of the role to become a default value? In this way, we don't do mapping or even don't need to modify the component.JSON. I'm not sure about it. just a guess :)

Another thing is I realized I hard coded "studio/client Installation path on hana dbnode" before, I will update it to use value from hdb_sizes.json.

deploy/v2/ansible/roles/inventory-validation/tasks/repo_validation.yml
YunziZhangAzure: Got it! Thanks!

deploy/v2/ansible/roles/hdb-server-install/tasks/main.yml
rukawata: Actually I just took a look at how we can get the installed component list, which is an XML based string. It does not seem to work as simple as I thought and may introduce a big architecture change.

Based on that, IMO it's preferred to just do a SID check at the moment. If they would like to add additional component, it can be manually done.

rukawata: Thanks for the advice!
As discussed, I implement the pre-check into component install role :)
As for HANA Studio/Client installation, I will implement some kind of pre-check as well, but will be in a separate PR.

PR #508, opened 2020-05-27
Adjust IP space for iSCSI

rukawata:
## Problem
iSCSI device need to have it's own subnet in sap vnet, since it is not tight to any SID.

## Description
- iSCSI will deploy it's own subnet, use .4 - 
- Move LB to still use .4 - .9
- Move DB VMs to use .10 -

## Test
local run terraform plan to confirm the IP spaces.

### DIFF
--- a/deploy/template_samples/clustered_hana_with_sbd.json
+++ b/deploy/template_samples/clustered_hana_with_sbd.json
@@ -48,7 +48,7 @@
         "subnet_iscsi": {
           "is_existing": "false",
           "name": "subnet-iscsi",
-          "prefix": "10.1.3.0/24",
+          "prefix": "10.1.4.0/24",
           "nsg": {
             "is_existing": "false",
             "name": "nsg-iscsi"
--- a/deploy/template_samples/clustered_hana_with_sbd.json
+++ b/deploy/template_samples/clustered_hana_with_sbd.json
@@ -44,11 +44,20 @@
             "is_existing": "false",
             "name": "nsg-db"
           }
+        },
+        "subnet_iscsi": {
+          "is_existing": "false",
+          "name": "subnet-iscsi",
+          "prefix": "10.1.3.0/24",
+          "nsg": {
+            "is_existing": "false",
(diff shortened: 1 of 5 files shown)


deploy/terraform/modules/common_infrastructure/iscsi.tf
```diff
@@ -8,6 +8,60 @@
...
+# Imports data of existing SAP iSCSI subnet
+data "azurerm_subnet" "iscsi" {
+  count                = local.iscsi_count == 0 ? 0 : (local.subnet_iscsi.is_existing ? 1 : 0)
+  name                 = split("/", local.subnet_iscsi.arm_id)[10]
+  resource_group_name  = split("/", local.subnet_iscsi.arm_id)[4]
+  virtual_network_name = split("/", local.subnet_iscsi.arm_id)[8]
+}
+
+# Creates SAP iSCSI subnet nsg
+resource "azurerm_network_security_group" "iscsi" {
```
YunziZhangAzure, line 32: The VNET, Subnet and NSG are using the same local name "iscsi".  Is this intentional? Should we give each of them a different name to identify each of them and maybe for future use?

rukawata: Yes, this is intentional. We should migrating the new code in the same way.

deploy/template_samples/clustered_hana_with_sbd.json
```diff
@@ -44,11 +44,20 @@
             "is_existing": "false",
             "name": "nsg-db"
           }
+        },
+        "subnet_iscsi": {
+          "is_existing": "false",
+          "name": "subnet-iscsi",
+          "prefix": "10.1.3.0/24",
+          "nsg": {
```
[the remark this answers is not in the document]
rukawata: Sorry, it should be 10.1.4.0 instead. I forgot to change it here.
This is to give space for app subnet, so my PR does not collide with Pete's.

PR #519, opened 2020-06-03
[HOTFIX] Skip prometheus if error

rukawata:
## Problem
There is a change in cert of Prometheus repo.
This caused add repo task to fail in pipeline.

## Description
The rest of the playbook has no dependency on prometheus.
Therefore we skip prometheus if add repo fails, but continue with the rest of playbook

### DIFF
--- a/deploy/ansible/roles/enable-prometheus/tasks/main.yml
+++ b/deploy/ansible/roles/enable-prometheus/tasks/main.yml
@@ -20,8 +20,7 @@
   ignore_errors: yes
 
 - name: Ensure Prometheus is installed and configured
-  when: (add_repo_status.failed is defined and add_repo_status.failed == false) or
-        (add_repo_status.skipped is defined and add_repo_status.skipped == True)
+  when: not (add_repo_status.failed is defined and add_repo_status.failed == true)
   block:
   - name: Install Prometheus with Node and HA Cluster exporters
     zypper:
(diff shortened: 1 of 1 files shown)


deploy/ansible/roles/enable-prometheus/tasks/main.yml
```diff
@@ -16,28 +16,34 @@
...
-  service:
-    name: prometheus-ha_cluster_exporter
-    state: started
-  when: hana_database.high_availability == True
+  register: add_repo_status
+  ignore_errors: yes
+
+- name: Ensure Prometheus is installed and configured
+  when: (add_repo_status.failed is defined and add_repo_status.failed == false) or
+        (add_repo_status.skipped is defined and add_repo_status.skipped == True)
```
YunziZhangAzure, line 24: I guess the logic here is when: not (add_repo_status.failed is defined and add_repo_status.failed == true) ?

rukawata: The task can only continue if the repo is configured successfully **or** was configured before.
And the add_repo_status looks like below for both condition:
```
ok: [10.1.1.5] => {
    "msg": {
        "changed": true,
        "failed": false,
        "repodata": {
            "alias": null,
            "autorefresh": "1",
            "enabled": "1",
            "gpgcheck": "1",
            "name": null,
            "priority": null,
            "url": "https://provo-mirror.opensuse.org/repositories/server:/monitoring/SLE_12_SP5/server:monitoring.repo"
        },
        "state": "present"
    }
}

ok: [10.1.1.5] => {
    "msg": {
        "changed": false,
        "skip_reason": "Conditional result was False",
        "skipped": true
    }
}
```

rukawata: Ah, you are correct. With the way you define it would cover both. I will update that. Thanks!

PR #525, opened 2020-06-04
Change NSG rule between management subnet and SAP VNET- Admin subnet

YunziZhangAzure:
1. Change the NSG between management subnet and Admin subnet
2. In test pipeline, change the number of iSCSI devices to be provisioned from 3 to 1 to save test
   time. Will add another test case to test 3 iSCSI devices in daily test pipeline.
3. Update test pipeline, before executing ansible, update the NSG of management, subnet-db, subnet-admin.

### DIFF
--- a/deploy/terraform/modules/hdb_node/hdb-nsg.tf
+++ b/deploy/terraform/modules/hdb_node/hdb-nsg.tf
@@ -4,9 +4,25 @@
 |                                                                              |
 +--------------------------------------4--------------------------------------*/
 
+# Creates network security rule to allow internal traffic for SAP db subnet
+resource "azurerm_network_security_rule" "nsr-internal-db" {
+  count                       = var.infrastructure.vnets.sap.subnet_db.nsg.is_existing ? 0 : 1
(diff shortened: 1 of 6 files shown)


deploy/terraform/modules/hdb_node/hdb-nsg.tf
```diff
@@ -33,5 +33,5 @@ resource "azurerm_network_security_rule" "nsr-admin" {
   source_port_range           = "*"
   destination_port_range      = "*"
   source_address_prefix       = var.infrastructure.vnets.management.subnet_mgmt.prefix
-  destination_address_prefix  = var.infrastructure.vnets.sap.subnet_db.prefix
+  destination_address_prefix  = var.infrastructure.vnets.sap.subnet_admin.prefix
```
rukawata, line 36: [Here](https://github.com/Azure/sap-hana/blob/f2ec3de58df92303da5b28fe9668accb731b5fbc/deploy/terraform/modules/hdb_node/hdb-nsg.tf#L20) needs to be swapped as well to deny traffic to `subnet_db`.

PR #531, opened 2020-06-05
Move subnets and NSG to appropriate modules

peterchoo:
## Problem

The end goal is to have the resources specific to the HDB or Application layers deployed only when that particular layer is deployed.  At present the subnets and NSGs for the HDB module and the application tier are both deployed in the `common_infrastructure` module.

See #518 

## Description

This PR moves the Application subnet and NSG to the `app_tier` module, and the `admin` and `db` subnets/NSGs to the HDB module.

Closes #518

## Pre-Requisites

Knowledge of Terraform

## Commitment to Test

- Testing takes 30 minutes
- Reviewing 12 files

### Deployment Testing

This has been listed first as the deployment can take upwards of 10 minutes to complete. While waiting to check the eventual result, the Code Review notes below can be looked at.

1. Ensure the Terraform modules have been initialised:\
   `util/terraform_v2.sh init`
1. Ensure the SAP download credentials have been set:\
   `util/set_sap_download_credentials.sh "<username>" "<password>" "single_node_hana"`
1. Plan the deployment:\
   `util/terraform_v2.sh plan "single_node_hana"`
   - [ ] Plan completes successfully without error
1. Apply the deployment plan:\
   `util/terraform_v2.sh apply "single_node_hana"`
   - [ ] Deployment completes successfully.
     :hand: Recent deployments have failed due to ARM issues in Azure, check the [Azure Status page](https://status.azure.com/en-gb/status) if deployment fails.

### Code Review 

1. The Common Infrastructure module:
   - [ ] [No longer creates the Admin, DB, or App NSGs](https://github.com/Azure/sap-hana/pull/531/files#diff-1cd04a5f01a866c3b657a2a4533b1688L17-L40)
   - [ ] [No longer imports the Admin, DB, or App NSG data if it already exists](https://github.com/Azure/sap-hana/pull/531/files#diff-1cd04a5f01a866c3b657a2a4533b1688L47-L67)
   - [ ] [No longer creates the Admin, DB, or App Subnets](https://github.com/Azure/sap-hana/pull/531/files#diff-ef75e58f59b9bbfd1b258a282065191bL65-L91)
   - [ ] [No longer imports the Admin, DB, or App Subnet data if it already exists](https://github.com/Azure/sap-hana/pull/531/files#diff-ef75e58f59b9bbfd1b258a282065191bL100-L123)
[...]

PR #540, opened 2020-06-06
Enhance to deploy without databases

rukawata:
## Description
Make it possible to deploy without database(s).

## Test
Deploy successful with rti_only:
```
$ util/terraform_v2.sh apply rti_only
...
Apply complete! Resources: 33 added, 0 changed, 0 destroyed.

Outputs:

ansible-jumpbox-public-ip-address = 52.188.203.46
ansible-jumpbox-username = azureadm
```


rukawata:
> Is it better to remove subnet-admin and subnet-db in rti.json as well?

That would come in a separate PR after PR #531 is completed. 
Otherwise this will create too many conflicts.

PR #544, opened 2020-06-09
Enhance the performance of Ansible stage

YunziZhangAzure:
1. Reduce the pausing time of waiting corosync to restart
2. Add the condition that when extracted folder exists, the following steps about downloading and integrity checking will be skipped.

### DIFF
--- a/deploy/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
+++ b/deploy/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
@@ -60,7 +60,7 @@
     state: restarted
 
 - pause:
-    seconds: 20
+    seconds: 15
 
 - name: Ensure the Corosync service is restarted on secondary node
   when: inventory_hostname == hana_database.nodes[1].ip_admin_nic
@@ -69,7 +69,7 @@
     state: restarted
 
 - pause:
-    seconds: 20
+    seconds: 15
 
 - name: Ensure the STONITH SBD is created
   when: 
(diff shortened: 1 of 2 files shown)


deploy/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
```diff
@@ -232,14 +232,5 @@
     - name: Ensure any required cluster resources are cleaned up
       shell: "crm resource cleanup rsc_SAPHana_{{ hana_database.instance.sid | upper }}_HDB{{ hana_database.instance.instance_number }}"
 
-    - name: Check current SBD status after clean resources
```
rukawata, line 235: Why this is removed?

deploy/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
```diff
@@ -69,7 +69,7 @@
     state: restarted
 
 - pause:
-    seconds: 20
+    seconds: 15
```
rukawata, line 72: Have to done enough testing to make sure 15s is sufficient?

YunziZhangAzure: So far, all tests look good and stable.

deploy/ansible/roles/hana-os-clustering/tasks/cluster-Suse.yml
YunziZhangAzure: This step is just to print the result. I used this step to determine when the SBD transit to a second node. Since the transit doesn't happen any more, this step can just be removed for the sake of simiplicity.

Issue #545, opened 2020-06-09
Deploying to an existing resource group [BUG]


KimForss:
When deploying to an existing resource group: 

if the parameter json file defines a location (in the infrastructure node) that is different from the location of the existing resource group the networking deployment will fail 

**Expected behavior**
Successful deployment of all resources into the existing resource group.

**Additional context**
It seems some of the location settings for the resources in the modules are using the location from infrastructure node and some are using the resource group location

PR #568, opened 2020-06-17
Fix the wrong rg location assignment when existing rg is used

rukawata:
## Problem
Fix issue #545 

## Solution
Always take the location of the rg regardless new rg or existing rg is used.

## Test
1. Create rg in westus2.
1. Deploy with rg location eastus in JSON.
1. All resources are correctly deployed in westus2.


rukawata:
> Thanks for your pr! Do you mind changing the following place as well?
> https://github.com/Azure/sap-hana/blob/985ea352f2cafcbc588cfc9341ed1bbbca3f1ba6/deploy/terraform/modules/common_infrastructure/iscsi.tf#L35

Good catch! Missed that one! Now it's updated as well.

rukawata:
> Lol, you are fast! sorry, I was about to say there are several places in the this file which needs to be changed. Could you kindly change them to use the location of resource group?
> https://github.com/Azure/sap-hana/blob/985ea352f2cafcbc588cfc9341ed1bbbca3f1ba6/deploy/terraform/modules/common_infrastructure/iscsi.tf
> 
> Line 69 and line 91 for example.

Np! Done :)

PR #570, opened 2020-06-19
Set secondary site with readaccess

rukawata:
## Problem
If we run the HSR configuration for the second time, it failed because secondary is already configured with HSR and do not have read access.

```
TASK [hana-system-replication : Ensure failure with meaningful message if no HN1 tenant db] ***
task path: /home/azureadm/sap-hana/deploy/ansible/roles/hana-system-replication/tasks/pre_checks.yml:50
Read vars_file 'vars/ha-packages.yml'
skipping: [10.1.1.10] => {
    "changed": false,
    "skip_reason": "Conditional result was False"
}
fatal: [10.1.1.11]: FAILED! => {
    "changed": false,
    "msg": "Automated HANA System Replication configuration is only supported for systems with a SYSTEMDB\nand multi-container systems where the tenant database is named as HN1.\nIf your requirements differ, please raise a GitHub issue: https://github.com/Azure/sap-hana/issues/new/choose\n"
}
```

## Solution
One option would be to change the `logmode` to `logreplay_readaccess`, however the HSR won't be able to repeat for other `logmode`.
Therefore the below change has been made:
- make sure the single/multiple container check only performs on primary node is HSR has been configured - meaning secondary node may not have read access.
- make sure `hana_has_tenant_db` is validated before use in condition - when HSR has been configured, `hana_has_tenant_db` will be undefined for secondary node hence fail the play.

## Test
Run playbook on an HSR configured cluster and no error.

### DIFF
--- a/deploy/ansible/roles/hana-system-replication/tasks/pre_checks.yml
+++ b/deploy/ansible/roles/hana-system-replication/tasks/pre_checks.yml
@@ -28,34 +28,37 @@
   failed_when: hana_process_list_result.rc != 3
   changed_when: false
 
-- name: Check whether HANA has been installed as a single container (no SYSTEMDB)
-  become_user: "{{ sid_admin_user }}"
-  shell: >
-    source ~/.bashrc ;
-    {{ hdbsql_systemdb_command }} "SELECT DATABASE_NAME FROM SYS.M_DATABASES"
-  register: hana_container_status
-  changed_when: false
-  failed_when: false
+- name: Check on the primary node whether installed HANA is single container or multi-container
+  when: ansible_hostname == hana_database.nodes[0].dbname
+  block:
+  - name: Check whether HANA has been installed as a single container (no SYSTEMDB)
+    become_user: "{{ sid_admin_user }}"
+    shell: >
+      source ~/.bashrc ;
+      {{ hdbsql_systemdb_command }} "SELECT DATABASE_NAME FROM SYS.M_DATABASES"
+    register: hana_container_status
+    changed_when: false
+    failed_when: false
 
-- name: Ensure systemdb fact is set
-  set_fact:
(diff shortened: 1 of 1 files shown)


deploy/ansible/roles/hana-system-replication/tasks/pre_checks.yml
```diff
@@ -82,7 +86,7 @@
 - name: Check that HANA DB SYSTEM user can access the tenant database {{ hana_tenant_database_name }}
   # If HSR is already enabled, we cannot read from the secondary node
   when:
-    - hana_has_tenant_db
+    - hana_has_tenant_db is defined and hana_has_tenant_db
     - (ansible_hostname == hana_database.nodes[0].dbname) or not hana_system_replication_enabled
```
YunziZhangAzure, line 90: It seems to me that one of these condition is adequate. Eg: when "(ansible_hostname == hana_database.nodes[0].dbname) or not hana_system_replication_enabled" ia true, 'hana_has_tenant_db is defined and hana_has_tenant_db' will be true. Otherwise, it will fail at line 75?

rukawata: Which condition you are referring to?
Iine 75 shows:     `when: not hana_has_system_db`.

There is one condition where it's a multi-container but with no tenant.
Under that condition, line 75 will not fail, but we should check the tenant existence here.

YunziZhangAzure: You are right! I'm wrong. I thought it was hana_has_system_db here.

rukawata: np :) Thanks for reviewing the PR!

PR #579, opened 2020-06-22
Enable deployment of App tier in test pipeline

YunziZhangAzure:
1. Enable deployment of SCS/ERS, Web dispatcher, including HA and Single mode in regular test pipeline.
2. Set timeout of testcases with only Terraform stage to 60 mins. Set timeout of testcases with Terraform and Ansible to 180mins.
3. All reusable testcases only have Terraform stage.

### DIFF
--- a/templates/job-template-allReuseCases.yaml
+++ b/templates/job-template-allReuseCases.yaml
@@ -3,6 +3,38 @@ parameters:
   osImage: ''
 
 jobs:
+- job: createAllNew_APP_SN_${{parameters.osVersion}}
+  timeoutInMinutes: 60
+  variables:
+    scenario: sap-allNew-APP-SN
+    testcase: $(scenario)-${{parameters.osVersion}}-$(Build.BuildId)
+  steps:
+  - template: test-case-template-update.yaml
+    parameters:
+      scenario: $(scenario)
+      testCaseName: $(testcase)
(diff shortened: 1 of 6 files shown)


templates/job-template-allReuseCases.yaml
```diff
@@ -3,6 +3,38 @@ parameters:
   osImage: ''
 
 jobs:
+- job: createAllNew_APP_SN_${{parameters.osVersion}}
+  timeoutInMinutes: 180
```
rukawata, line 7: Maybe a smaller timeout since it's only terraform?

templates/job-template-allReuseCases.yaml
```diff
@@ -3,6 +3,38 @@ parameters:
...
+      scenario: $(scenario)
+      testCaseName: $(testcase)
+      placeHolder: var-rg-name
+      value: $(testcase)
+  - template: terraform-deployment-steps.yaml
+    parameters:
+      testCaseName: $(testcase)
+      osImage: ${{parameters.osImage}}
+- job: createAllNew_APP_HA_${{parameters.osVersion}}
+  timeoutInMinutes: 180
```
rukawata, line 23: Same as above :)

YunziZhangAzure: Ok, I change it to 60 mins.

PR #583, opened 2020-06-23
Add VM sizing json file for applications

### LINKED ISSUE #517: Terraform: Application Tier VM and Disk Sizing
# Problem statement

There is no specification for the Application Tier VMs.  #515 has assumed a `Standard_D4a_v4` will be sufficient.  This is currently ailing pipelines tests due to quota restrictions.

## Required Information

For both the SCS/ERS, and the Application VMs

1. VM Size(s) - will there only be one, or will there be a t-shirt size list of sizes to pick?
1. OS disk type and size(s) - t-shirt size list?
1. Additional Data disks?
   1. If so, sizings?

peterchoo:
# Problem

See #517, addresses [feedback left here](https://github.com/Azure/sap-hana/issues/517#issuecomment-647668685).

## Description

This PR adds an `app_sizes.json` file with a default for each of the `app`, `scs`, and `web` groups.

Closes #517

This also adds the JSON Adjusting Template doc line for the [web dispatcher server count](https://github.com/Azure/sap-hana/pull/583/files#diff-d346278c57da72080e62e06c2648c2d9R142)


peterchoo:
> LGTM! The only thing misses is the ability to define and create multiple data disk from sizing json.
> Meaning, the amount of data disk should be configurable if needed.

Apologies, juggling a couple of PRs today, raised as #586

deploy/terraform/modules/app_tier/variables_local.tf
```diff
@@ -40,24 +47,44 @@ locals {
...
   # Default VM config should be merged with any the user passes in
-  app_sku_map = merge(
-    {
-      app = "Standard_D4s_v3,false"
-      scs = "Standard_D4s_v3,false"
-      web = "Standard_D4s_v3,false"
+  app_sizing = lookup(local.sizes.app, local.vm_sizing, {
+    "compute": {
+      "vm_size": "Standard_D4s_v3",
+      "accelerated_networking": false
```
YunziZhangAzure, line 53: Not sure whether the default value here is needed, because vm_sizing is already given a default value as "Default", and the default value is already defined in app_sizes.json. This point won't affect the functionality, so I can approve this pr right now. Maybe discuss it later.

peterchoo: As the `vm_sizing` value can be overridden by the user, if they provide something not in the JSON (at this point in time, anything other than `Default`) then the deployment will fail.  I added this as a back up so that something can be be deployed.

YunziZhangAzure: IMO, if the customers provide wrong input, the script should fail, instead of giving customers an illusion that it is executed correctly. Otherwise, they may be confused later why the result is not what they expected. Alert is better than silent fault. Just my thought. Open to discussion.

rukawata: The problem I see is Terraform does not have a proper way to throw an error.
In this specific case, if we do not provide a default value, the deployment will fail with some error that does not have anything to do with what really happens, unfortunately :(

PR #588, opened 2020-06-23
Allow multiple data disks for SCS and App VMs

### LINKED ISSUE #586: Application VM sizing should allow for multiple disks, with a default of 1.
# Problem statement

The `deploy/app_sizes.json` should allow for multiple storage disks, even though the common use case is one.

peterchoo:
# Problem

Disks for the Application Tier VM sizing should allow multiple disks to be added, even though the expected configuration is 1.  This is to allow customers the ability to update it if they so desire.

See #586 

## Description

Updates the app_sizes.json file to use an array for storage descriptions. Adds a name field. Updates the logic for the App and SCS VM disk creation and attachment.

Closes #586

### DIFF
--- a/deploy/app_sizes.json
+++ b/deploy/app_sizes.json
@@ -5,12 +5,13 @@
         "vm_size": "Standard_D4s_v3",
         "accelerated_networking": false
       },
-      "storage": {
+      "storage": [{
+        "name": "data",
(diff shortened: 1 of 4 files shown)


YunziZhangAzure:
Just a reminder, the creation of data disk and attachment of webdispatcher is not implemented yet.

PR #592, opened 2020-06-24
Add defaults to app tier

rukawata:
## Problem
Deployment fails when newly added parameter not set in the input JSON

## Solution
Create defaults for all the required parameters in input JSON
In this PR, focus on app tier

## Test
Remove all parameters from application section and terraform plan works fine.


YunziZhangAzure:
Thanks for your pr! 

This fix can solve the previous issue "missing webdispatcher_count in input JSON"
but I'm afraid more changes may be needed to support "newly added parameter not set in the input JSON"

Eg: I remove scs_high_availability field from template json. The following point fails:  https://github.com/Azure/sap-hana/blob/9b9093a906eb5cf25cea2f1a382fd4c0c3057a9a/deploy/terraform/modules/app_tier/infrastructure.tf#L31

I guess to implement "newly added parameter not set in the input JSON", all var variables should be changed to use local variables which you are creating in this pr.

rukawata:
> Thanks for your pr!
> 
> This fix can solve the previous issue "missing webdispatcher_count in input JSON"
> but I'm afraid more changes may be needed to support "newly added parameter not set in the input JSON"
> 
> Eg: I remove scs_high_availability field from template json. The following point fails:
> 
> https://github.com/Azure/sap-hana/blob/9b9093a906eb5cf25cea2f1a382fd4c0c3057a9a/deploy/terraform/modules/app_tier/infrastructure.tf#L31
> 
> I guess to implement "newly added parameter not set in the input JSON", all var variables which are specified in input JSON should be changed to use local variables which you are creating in this pr.

Actually this is my mistake! I should've updated to use local default variable in the infrastructure.tf.
Now I changed that part as well :)

YunziZhangAzure:
Looks good!

Issue #590, opened 2020-06-24
[BUG] Terraform Plan/Apply should fail if user specifies an incorrect sizing for app tier


peterchoo:
# Describe the bug

[If the user incorrectly specifies the `vm_sizing`, the plan/apply should fail instead of working with some defaults which give the user the impression their deployment has succeeded.](https://github.com/Azure/sap-hana/pull/583#discussion_r444450917)

## To reproduce

Steps to reproduce the behavior, from the project root:

1. Update the single_node_hana deployment template setting `application.vm_sizing` to an unknown value, e.g. "NotARealSizing"
1. Run `util/terraform_v2.sh plan single_node_hana`

## Expected behavior

Expect the plan to fail due to missing configuration.

peterchoo:
Closed by #593

PR #593, opened 2020-06-24
Default value of app scs web sizing

YunziZhangAzure:
With this pr, we don't have to change the variables_local.tf if the Default value in app_size.json gets changed in the future.


rukawata:
This closes #590.

PR #594, opened 2020-06-24
Set default values for HDB module

rukawata:
## Problem
It becomes a problem if certain parameters are not given in input JSON

## Solution
1. Assign default value to the variables passed under databases section.
1. Also make adjustment based on the fact only 1 HANA database will be deployed.

## Test
Below testcase has been done locally:
1. Deploy single HANA node, then change `high_availability` to true
1. Remove the whole `databases` section
1. Remove the dict inside databases section, meaning: `"databases": []`
1. Remove `platform`
1. Set `platform` to `NONE`
1. Remove `os`
1. Remove `size`
1. Remove `high_availability`
1. Remove `authentication`
1. Remove `instance`
1. Remove `sid`
1. Remove `credentials`
1. Remove `components`
1. Remove `xsa`, `shine` together
1. Remove `dbnodes`

### DIFF
--- a/deploy/terraform/modules/common_infrastructure/variables_local.tf
+++ b/deploy/terraform/modules/common_infrastructure/variables_local.tf
@@ -3,20 +3,28 @@ variable "is_single_node_hana" {
   default     = false
 }
 
+# Set defaults
 locals {
-
   # Filter the list of databases to only HANA platform entries
   hana-databases = [
     for database in var.databases : database
-    if database.platform == "HANA"
+    if try(database.platform, "NONE") == "HANA"
   ]
+  hdb    = try(local.hana-databases[0], {})
+  hdb_ha = try(local.hdb.high_availability, "false")
+  hdb_os = try(local.hdb.os,
+    {
+      "publisher" = "suse",
+      "offer"     = "sles-sap-12-sp5",
+      "sku"       = "gen1"
+  })
 
   # iSCSI target device(s) is only created when below conditions met:
   # - iscsi is defined in input JSON
   # - AND
   #   - HANA database has high_availability set to true
   #   - HANA database uses SUSE
-  iscsi_count = lookup(var.infrastructure, "iscsi", {}) != {} && (length(local.hana-databases) > 0 ? (local.hana-databases[0].high_availability && upper(local.hana-databases[0].os.publisher) == "SUSE") : false) ? var.infrastructure.iscsi.iscsi_count : 0
+  iscsi_count = lookup(var.infrastructure, "iscsi", {}) != {} && (length(local.hana-databases) > 0 ? (local.hdb_ha && upper(local.hdb_os.publisher) == "SUSE") : false) ? var.infrastructure.iscsi.iscsi_count : 0
 
   # Shortcut to iSCSI definition
   iscsi = merge(lookup(var.infrastructure, "iscsi", {}), { "iscsi_count" = "${local.iscsi_count}" })
--- a/deploy/terraform/modules/hdb_node/variables_local.tf
+++ b/deploy/terraform/modules/hdb_node/variables_local.tf
@@ -14,56 +14,114 @@ variable "ppg" {
   description = "Details of the proximity placement group"
 }
 
+# Set defaults
+locals {
+  hdb_list = [
+    for db in var.databases : db
+    if try(db.platform, "NONE") == "HANA"
+  ]
+  enable_deployment = (length(local.hdb_list) > 0) ? true : false
+
+  # Filter the list of databases to only HANA platform entries
+  hdb         = try(local.hdb_list[0], {})
+  hdb_version = try(local.hdb.db_version, "2.00.043")
+  hdb_os = try(local.hdb.os,
+    {
+      "publisher" = "suse",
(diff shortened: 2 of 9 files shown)


deploy/terraform/modules/common_infrastructure/variables_local.tf
```diff
@@ -3,20 +3,28 @@ variable "is_single_node_hana" {
...
+      "sku"       = "gen1"
+  })
 
   # iSCSI target device(s) is only created when below conditions met:
   # - iscsi is defined in input JSON
   # - AND
   #   - HANA database has high_availability set to true
   #   - HANA database uses SUSE
-  iscsi_count = lookup(var.infrastructure, "iscsi", {}) != {} && (length(local.hana-databases) > 0 ? (local.hana-databases[0].high_availability && upper(local.hana-databases[0].os.publisher) == "SUSE") : false) ? var.infrastructure.iscsi.iscsi_count : 0
+  iscsi_count = lookup(var.infrastructure, "iscsi", {}) != {} && (length(local.hana-databases) > 0 ? (local.hdb_ha && upper(local.hdb_os.publisher) == "SUSE") : false) ? var.infrastructure.iscsi.iscsi_count : 0
```
YunziZhangAzure, line 27: I think iSCSI devices are not only applied in hana database scenario? It can be used by NFS cluster, ASCS cluster and database cluster. Is it better to remove the condition about hana-database?

deploy/terraform/modules/common_infrastructure/variables_local.tf
```diff
@@ -3,20 +3,28 @@ variable "is_single_node_hana" {
...
+    if try(database.platform, "NONE") == "HANA"
   ]
+  hdb    = try(local.hana-databases[0], {})
+  hdb_ha = try(local.hdb.high_availability, "false")
+  hdb_os = try(local.hdb.os,
+    {
+      "publisher" = "suse",
+      "offer"     = "sles-sap-12-sp5",
+      "sku"       = "gen1"
+  })
```
YunziZhangAzure, line 21: This block about hdb can be removed if we decide to exclude the following condition(Please see the comment below). A benefit of doing so is to decouple the modules. This is the common module, I think ideally it doesn't depend on any other modules.

deploy/terraform/modules/jumpbox/variables_local.tf
```diff
@@ -26,6 +26,15 @@ variable "random-id" {
...
 
+# Set defaults
+locals {
+  hdb_list = [
+    for db in var.databases : db
+    if try(db.platform, "NONE") == "HANA"
+  ]
+  hana-sid = try(local.hdb_list[0].instance.sid, "")
+}
+
```
YunziZhangAzure, line 37: hdb_list and hana-sid seem not to be used anywhere in this module. Maybe remove this block? Please correct me if I'm wrong.

deploy/terraform/modules/hdb_node/variables_local.tf
```diff
@@ -14,56 +14,117 @@ variable "ppg" {
...
-# Imports HANA database sizing information
+# Set defaults
 locals {
-  sizes = jsondecode(file("${path.root}/../hdb_sizes.json"))
-}
+  hdb_list = [
+    for db in var.databases : db
+    if try(db.platform, "NONE") == "HANA"
+  ]
+  enable_deployment = (length(local.hdb_list) > 0) ? true : false
```
YunziZhangAzure, line 23: Not sure if TF can use such expression:

```suggestion
  enable_deployment = length(local.hdb_list) > 0
```

deploy/terraform/modules/common_infrastructure/variables_local.tf
rukawata: The condition is if no HA enabled, we do not deploy iscsi device.
Once we have other use case for iscsi, I will add logic to that.

deploy/terraform/modules/common_infrastructure/variables_local.tf
rukawata: We can't remove it, see comment [here](https://github.com/Azure/sap-hana/pull/594#discussion_r445288580). 
Also this will not impact if we want to separate the component because this is purely taken from the input JSON.

deploy/terraform/modules/jumpbox/variables_local.tf
rukawata: I leave it here is because we had similar definition in current [master](https://github.com/Azure/sap-hana/blob/07c69bdf4c8a60adbd3e760b8145cac0fd00bdad/deploy/terraform/modules/jumpbox/variables_local.tf#L70-L76).
I can remove it but will need to run additional tests to confirm it does not break anything.

deploy/terraform/modules/hdb_node/variables_local.tf
rukawata: The suggest expression will not work in TF unfortunately.

deploy/terraform/modules/jumpbox/variables_local.tf
rukawata: Tested and seems to be no problem. Removed this part from jumpbox local variable :)

PR #596, opened 2020-06-25
Always have 2 feip on the SCS load balancer

### LINKED ISSUE #581: [BUG] error occurs in the Application Tier when transitioning from HA to SA
**Describe the bug**
An error occurs in the Application Tier when transitioning from HA to SA.

```
Error: Error Creating/Updating Load Balancer "XX1_scs-alb" (Resource Group "modeegan-XX1-rg"): network.LoadBalancersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidResourceReference" Message="Resource /subscriptions/49dcf07f-3ea3-40f9-a87a-c75a49c32c5f/resourceGroups/modeegan-XX1-rg/providers/Microsoft.Network/loadBalancers/XX1_scs-alb/frontendIPConfigurations/XX1_ers-feip referenced by resource /subscriptions/49dcf07f-3ea3-40f9-a87a-c75a49c32c5f/resourceGroups/modeegan-XX1-rg/providers/Microsoft.Network/loadBalancers/XX1_scs-alb/loadBalancingRules/XX1_ERS_3310 was not found. Please make sure that the referenced resource exists, and that both resources are in the same region." Details=[]

  on ../../../../sap-hana/deploy/terraform/modules/app_tier/infrastructure.tf line 19, in resource "azurerm_lb" "scs":
  19: resource "azurerm_lb" "scs" {
```

**To reproduce**
1. terraform apply with scs_high_availability set to true
2. change scs_high_availability to false and perform terraform apply
3. See Error
4. Repeat terraform apply.
5. Completes successfully.

**Expected behavior**
It should complete successfully the first time and not require a repeat apply

peterchoo:
# Problem

Terraform fails first time if SCS HA is changed from true to false. See #581 

## Description

Updates the SCS Load Balancer to permanently have two front end IPs instead of a dynamic block based on the SCS HA variable.

Adding `depends_on: [azurerm_lb.scs]` to the rules did not allow Terraform to detect the dynamic `frontend_ip_configuration` block change.

This change allows the SCS HA components (e.g. ERS instance and all related configuration) bar the FrontEnd IP configuration chang to be added/destroyed by the SCS HA flag.

Closes #581

### DIFF
--- a/deploy/terraform/modules/app_tier/infrastructure.tf
+++ b/deploy/terraform/modules/app_tier/infrastructure.tf
@@ -27,14 +27,18 @@ resource "azurerm_lb" "scs" {
   resource_group_name = var.resource-group[0].name
   location            = var.resource-group[0].location
 
-  dynamic "frontend_ip_configuration" {
-    for_each = range(var.application.scs_high_availability ? 2 : 1)
-    content {
-      name                          = "${upper(var.application.sid)}_${frontend_ip_configuration.value == 0 ? "scs" : "ers"}-feip"
-      subnet_id                     = var.infrastructure.vnets.sap.subnet_app.is_existing ? data.azurerm_subnet.subnet-sap-app[0].id : azurerm_subnet.subnet-sap-app[0].id
-      private_ip_address_allocation = "Static"
-      private_ip_address            = cidrhost(var.infrastructure.vnets.sap.subnet_app.prefix, tonumber(frontend_ip_configuration.value) + local.ip_offsets.scs_lb)
-    }
+  frontend_ip_configuration {
+    name                          = "${upper(var.application.sid)}_scs-feip"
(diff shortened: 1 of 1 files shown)


peterchoo:
> Thanks for your pr! Nancy and I did talk about this point a little bit a few days ago. I was wondering why we didn't give scs and ers load balancer frontend ip respectively. Did you have any reason for the first version? Was it by design?

The original design attempted to only have the ERS Frontend IP configuration when required (i.e. the Application High Availability is true). The problem in #581 is evident when a HA deployment changes to a SA deployment, and Terraform does not see the dependency between the ERS Load Balancer Rules and the removal of the ERS Frontend IP Configuration.  This caused an error where the FEIP was removed before the rule and the `terraform apply` would fail.

Issue #492, opened 2020-05-15
Ansible: Change Inventory Structure


mkdeegan:
**Is your feature request related to a problem? Please describe.**
Currently there is no way for multiple inventories to aggregate.
Inventories should be keyed off an identifier (SID) and an Application Layer.
A structure like the one shown below allows Ansilble to execute against groups, and the members of the Groups to have a fact set to Identify the operational _tier_
Inventories should be output in a YAML format.

**Describe the solution you'd like**
A YAML formate file with servers grouped by SID and Tier
```
# /*---------------------------------------------------------------------------8
# |                                                                            |
# |                                  HOSTS                                     |
# |                                                                            |
# +------------------------------------4--------------------------------------*/
---
# +------------------------------------4--------------------------------------*/
Z00_HDB:
    hosts:
        z00hdb00:
            ansible_host: z00hdb00.scus.azure.mysapbasis.com
    vars:
        tier: hdb
        # ansible_ssh_common_args: '-o ProxyCommand="ssh -W %h:%p -q ansible@pub0scusdeploy00.southcentralus.cloudapp.azure.com"'
Z00_SCS:
    hosts:
        z00scs00:
            ansible_host: z00scs00.scus.azure.mysapbasis.com
    vars:
        tier: scs
        # ansible_ssh_common_args: '-o ProxyCommand="ssh -W %h:%p -q ansible@pub0scusdeploy00.southcentralus.cloudapp.azure.com"'
Z00_PAS:
    hosts:
        z00app00:
            ansible_host: z00app00.scus.azure.mysapbasis.com
    vars:
        tier: pas
        # ansible_ssh_common_args: '-o ProxyCommand="ssh -W %h:%p -q ansible@pub0scusdeploy00.southcentralus.cloudapp.azure.com"'
Z00_APP:
    hosts:
        z00app01:
            ansible_host: z00app01.scus.azure.mysapbasis.com
        z00app02:
            ansible_host: z00app02.scus.azure.mysapbasis.com
    vars:
        tier: app
        # ansible_ssh_common_args: '-o ProxyCommand="ssh -W %h:%p -q ansible@pub0scusdeploy00.southcentralus.cloudapp.azure.com"'
Z00_WEB:
    hosts:
        z00web00:
            ansible_host: z00web00.scus.azure.mysapbasis.com
    vars:
        tier: web
...
```

[...]

PR #597, opened 2020-06-26
Add SID to inventory yml file

YunziZhangAzure:
This pr is to include SID to Inventory file. See #492 

### Test
1. Currently we don't have the requirement that multiple Hana databases installed at one execution, but we want to make our codes extensible -- support multiple SIDs under one product. To test this feature, I manually change hosts.yml to the following configuration and test it. 
It shows what our hosts.yml may look like and proves the change is compatible with current Ansible script.
hosts.yml
image

PLAY [hanadbnodes] **************************************************************************************************************

TASK [Gathering Facts] **********************************************************************************************************
ok: [10.1.1.11]
ok: [10.1.1.10]
fatal: [10.1.1.13]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host 10.1.1.13 port 22: Connection timed out", "unreachable": true}
fatal: [10.1.1.12]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host 10.1.1.12 port 22: Connection timed out", "unreachable": true}

The reason why it fail on 13 and 12 is because these two hosts don't exist. This doesn't matter because this test case is only used to show what our inventory file look like and prove our Ansible script is working fine with the change.

### DIFF
--- a/deploy/ansible/roles/inventory-validation/tasks/repo_validation.yml
+++ b/deploy/ansible/roles/inventory-validation/tasks/repo_validation.yml
@@ -27,4 +27,3 @@
   block:
     - name: Check yum repo on HANA VMs
       command: yum repolist
-
--- a/deploy/terraform/module.tf
+++ b/deploy/terraform/module.tf
@@ -89,6 +89,7 @@ module "output_files" {
   loadbalancers                = module.hdb_node.loadbalancers
   hdb-sid                      = module.hdb_node.hdb-sid
   hana-database-info           = module.hdb_node.hana-database-info
+  app-sid                      = module.app_tier.app-sid
   nics-scs                     = module.app_tier.nics-scs
   nics-app                     = module.app_tier.nics-app
   nics-web                     = module.app_tier.nics-web
--- a/deploy/terraform/modules/app_tier/outputs.tf
+++ b/deploy/terraform/modules/app_tier/outputs.tf
@@ -9,3 +9,7 @@ output "nics-app" {
 output "nics-web" {
   value = azurerm_network_interface.web
 }
+
+output "app-sid" {
+  value = local.application_sid
+}
--- a/deploy/terraform/modules/output_files/ansible_inventory.yml.tmpl
+++ b/deploy/terraform/modules/output_files/ansible_inventory.yml.tmpl
@@ -5,11 +5,8 @@ all:
 %{~ for ip-iscsi in ips-iscsi }
         ${ip-iscsi}:
           ansible_connection:  "ssh"
-%{~ if iscsi.authentication.type == "key" }
           ansible_user:        "${iscsi.authentication.username}"
-%{~ endif }
 %{~ if iscsi.authentication.type == "password" }
-          ansible_user:        "${iscsi.authentication.username}"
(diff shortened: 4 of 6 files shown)


mkdeegan:
Yunzi,
what is the basis for this requirement?
_Currently we don't have the requirement that multiple Hana databases installed at one execution, but we want to make our codes extensible -- support multiple SIDs under one product. To test this feature, I manually change hosts.yml to the following configuration and test it.
It shows what our hosts.yml may look like and proves the change is compatible with current Ansible script.
hosts.yml_

If we are talking about installing multiple SAP SID's at one execution, that is an invalid use case.

If we are talking a Scale-Out scenario I could talk through a thought process.

There are some problems with the naming and structure.
`hananodes:` it too generic. If inventories were to be aggregated there would be collisions. Do not use this.

In every conversation about an SAP deployment the unique identifier will be the SAPSID.


First the Scale-Up:
HANA might be using either the SAPSID or another designator for the DBSID.
So we need to build a naming system like this:
`<SAPSID>_HDB_<DBSID>:`

and in a stanza, it would look like this:
```
<SAPSID>_HDB:
  hosts:
    <hostname>:
      ansible_host:       <ip or FQDN>
      ansible_connection: ssh
      ansible_user:       azureadm
  vars:
    scale: up
    dbsid: <DBSID>
    scale: UP
```
for a HA environment you would duplicate the hostname block.


Scale-Out:
Under a hypothetical Scale out scenario it would be exactly the same, but we set a parameter to indicate a different functionality.

```
<SAPSID>_HDB_<DBSID>:
  hosts:
    <hostname_node01>:
      ansible_host:       <ip or FQDN>
      ansible_connection: ssh
      ansible_user:       azureadm
    <hostname_node02>:
      ansible_host:       <ip or FQDN>
      ansible_connection: ssh
      ansible_user:       azureadm
    <hostname_node03>:
      ansible_host:       <ip or FQDN>
      ansible_connection: ssh
      ansible_user:       azureadm
    <hostname_node04>:
      ansible_host:       <ip or FQDN>
      ansible_connection: ssh
      ansible_user:       azureadm
  vars:
    scale: up
    dbsid: <DBSID>
    scale: OUT
```


hananodes and children construct is an erroneous thought process, because you will never take an action outside of a deployment unit.

In the deployment for a SDU, the structure I used is illustrated below. In this way, since one parameter that is passed to ansible is the SID, the correct groups can be targeted at the correct isolation level.

```
Z00_HDB:
    hosts:
        z00hdb00:
            ansible_host: <FQDN>
    vars:
        tier: hdb
Z00_SCS:
    hosts:
        z00scs00:
            ansible_host: <FQDN>
    vars:
        tier: scs
Z00_PAS:
    hosts:
        z00app00:
            ansible_host: <FQDN>
    vars:
        tier: pas
Z00_APP:
    hosts:
        z00app01:
            ansible_host: <FQDN>
        z00app02:
             ansible_host: <FQDN>
    vars:
        tier: app
Z00_WEB:
    hosts:
        z00web00:
            ansible_host: <FQDN>
    vars:
        tier: web

```

PR #631, opened 2020-07-13
Deploy NSG into resource group of the existing vnet.

KimForss:
### Problem 

The current design will create NSGs in the same resource group as the VMs. When deploying multiple SDUs and having the VNet in another resource group the subnets will end up with multiple NSGs all with the same name (as each SDU deployment creates a new one)

### Solution

Add the NSGs in the same Resource group as the VNet.


deploy/terraform/modules/hdb_node/infrastructure.tf
```diff
@@ -36,16 +36,16 @@ data "azurerm_subnet" "subnet-sap-db" {
 resource "azurerm_network_security_group" "nsg-admin" {
   count               = local.enable_deployment ? (local.sub_admin_nsg_exists ? 0 : 1) : 0
   name                = local.sub_admin_nsg_name
-  location            = var.resource-group[0].location
```
rukawata, line 39: location is missing:
```suggestion
  virtual_network_name = var.vnet-sap[0].location
```

deploy/terraform/modules/hdb_node/infrastructure.tf
```diff
@@ -36,16 +36,16 @@ data "azurerm_subnet" "subnet-sap-db" {
...
-  resource_group_name = var.resource-group[0].name
+  resource_group_name  = var.vnet-sap[0].resource_group_name
+  virtual_network_name = var.vnet-sap[0].name
```
rukawata, line 40: virtual_network_name is not part of the nsg attribute.
You can use [azurerm_subnet_network_security_group_association](https://www.terraform.io/docs/providers/azurerm/r/subnet_network_security_group_association.html) to link the nsg and subnet.

KimForss: fixed

deploy/terraform/modules/hdb_node/infrastructure.tf
KimForss: fixed

PR #638, opened 2020-07-14
[HOTFIX] Save hostname in output.json

rukawata:
## Problem
When hostname is different than VM resource name, resource name get to saved in output.json and used by ansible.

## Solution
Pass hostname which is without '_' to output.json

### DIFF
--- a/deploy/terraform/modules/output_files/inventory.tf
+++ b/deploy/terraform/modules/output_files/inventory.tf
@@ -46,9 +46,10 @@ resource "local_file" "output-json" {
       xsa               = database.xsa,
       shine             = database.shine,
       nodes = [for ip-dbnode-admin in local.ips-dbnodes-admin : {
-        dbname       = local.hdb_vms[index(local.ips-dbnodes-admin, ip-dbnode-admin)].name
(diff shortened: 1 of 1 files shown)


deploy/terraform/modules/output_files/inventory.tf
```diff
@@ -46,9 +46,10 @@ resource "local_file" "output-json" {
       xsa               = database.xsa,
       shine             = database.shine,
       nodes = [for ip-dbnode-admin in local.ips-dbnodes-admin : {
-        dbname       = local.hdb_vms[index(local.ips-dbnodes-admin, ip-dbnode-admin)].name
+        // Hostname is required for Ansible, therefore set dbname from resource name to hostname
+        dbname       = replace(local.hdb_vms[index(local.ips-dbnodes-admin, ip-dbnode-admin)].name, "_", "")
```
tniek, line 50: are there any other characters that could be part of the resource name, but are not allowed in the host name?

rukawata: This is about being consistency.
We use the same value for computer_name which is the hostname of VM.

PR #641, opened 2020-07-15
Basic resource creation for sap deployer

rukawata:
## Problem
The former RTI which is a VM where ansible execution happens is part of the deployment.
We need to separate it into it's own terraform deployment.

## Solution
1. Separate former RTI into a terraform unit `sap_deployer`.
1. Separate definition of `sap_deployer` from `jumpboxes` to `deployers`.
1. Rename RTI to deployer because it deploys resources (terraform + ansible)
1. Update default values according to resolve issue #644 (partially)
1. This deployment will deploy:
1.1 network if > 1 deployer will be deployed
1.2 storage account (for diag only. remote tfstate SA will be deployed with `SAP_Libarary`) if > 1 deployer will be deployed
1.3 1..n deployer

## Test:
1. cd `sap-hana/deploy/terraform/bootstrap/sap_deployer`
2. create `input.json` with below content:
```
{
  "infrastructure": {
    "region": "westus2"
  },
  "sshkey": {
    "path_to_public_key": "~/.ssh/id_rsa.pub",
    "path_to_private_key": "~/.ssh/id_rsa"
  },
  "options": {
    "enable_secure_transfer": true
  }
}
```
3. `terraform init`
4. `terraform apply -auto-approve -var-file=input.json`
5. Search for resource `deployer-xxx` in portal

## Notes
Below are items (not limited) will be added later in separate PRs:

1. - [x] Support custom image for deployer (we provide an image with things preconfigured). PR #650 
1. - [x]  Add MSI, so deployer can access storage account as backend config storage. PR #651
1. - [x] Add outputs in tfstate. PR #652 
1. - [x] Prepare script to upload tfstate. PR #652
1.  Add KV to deployer (not for this project)
1.  Use KV to save SSH key pair for deployer (not for this project)
1.  Prepare backend config hcl on deployer and save in SA (to make deployer stateless, done by SA PR)
...


rukawata:
> The test instruction may need to be updated.

Updated :)

deploy/terraform/run/providers.tf
```diff
@@ -0,0 +1,34 @@
+/*-----------------------------------------------------------------------------8
+|                                                                              |
```
YunziZhangAzure, line 2: Do we need this file? there are several folders underneath /run, they are the root module I think.

deploy/terraform/run/sap_deployer/module.tf
```diff
@@ -0,0 +1,13 @@
+/*
+Description:
```
YunziZhangAzure, line 2: This file may not be needed. The directory /run/sap_deployer is used to init azurem as backend. Sync with you offline about this point.

deploy/terraform/run/providers.tf
rukawata: You are right, removed :)

PR #648, opened 2020-07-20
No SAP vnet gets deployed when no application or database is deployed

rukawata:
## Problem
SAP vent gets deployed regardless or application/database is deployed or not.

## Solution
Define `enable_sap_vnet` to switch on/off SAP vnet deployment

## Test
1. Deploy with `application.enable_deployment` set to `false` and `database.*.platform` set to `NONE`. No `vnet-sap` gets deployed.
2. Set `application.enable_deployment` set to `true`, see `vnet-sap` gets deployed now.

### DIFF
--- a/deploy/terraform-units/modules/sap_library/infrastructure.tf
+++ b/deploy/terraform-units/modules/sap_library/infrastructure.tf
@@ -1,48 +1,36 @@
-/*-----------------------------------------------------------------------------8
-|                                                                              |
-|                                RESOURCE GROUP                                |
-|                                                                              |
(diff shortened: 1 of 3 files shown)


rukawata:
> I'm thinking this solution definitely works for the current scenario, but when we start to do task "Phase II – Modularize the common network infrastructure as separate from the repeatable SDU"
> When should vnet-sap be deployed? at deploying sap_landscape time or at deploying sap_system time?

With SDU, vnet-sap will be deployed with sap_landscape :)
This is just to fix a incorrect behavior with current codebase.

PR #654, opened 2020-07-21
Add module to bootstrap

rukawata:
## Problem
Need module calls under bootstrap

## Solution
1. Copy all tf under run into bootstrap without `modules`
1. Correct the link in module.tf to `run/modules`

## Test
1. cd `sap-hana/deploy/terraform/bootstrap`
1. `terraform plan -var-file=../../template_samples/<pick you working sample>`


rukawata:
> Just want to confirm, at last, the boostrap/module.tf will invoke deploy_platform module right?
> We won't invoke any sap_landscape and sap_system module from here right?

Yes that is correct!

PR #660, opened 2020-07-23
Add instruction to mount dedicated lun for kdump core dump files

prtyag:
Kdump core dump files take significant amount of space when dump files are written during kernel crash. If we don't use a dedicated lun for kdump then the root file system will be exhausted. So in order to solve this problem we have decided to attach a dedicated lun for kdump at the lun_id=3. This change detects the dedicated lun attach with lun_id=3 and mount that to the /var/crash directory.


prtyag:
Decided to use lun_id=3 for kdump dedicated lun. so instead of size identifier this PR use lun_id as a identifier.

tools/enable-kdump.sh
```diff
@@ -24,6 +26,40 @@ supported_version=( "12-SP2"
     "15-SP2"
 )
 
+# scan the added lun and mount that to /var/crash
+# A race condition requires that rescan-scsi-bus.sh be run twice
+# if logical units are mapped for the first time. During the first scan,
+# rescan-scsi-bus.sh only adds LUN0; all other logical units are added in the second scan.
+# refer https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/online_storage_reconfiguration_guide/rescan-scsi-bus
+for i in {0..1}
```
jaiawasthi, line 38: sometimes i have seen it doesnt work at all, what do we do to handle that case.
if you are always successful its fine[during your testing], else add a echo during start of rescan and add a warning that if it doesnt comeplte within 3-4 mins, reboot the system.

tools/enable-kdump.sh
```diff
@@ -24,6 +26,40 @@ supported_version=( "12-SP2"
...
+# format a disk
+mkfs.xfs -f $luns_dev_mapper
+ExitIfFailed $? "Unable format kdump dedicated lun"
+
+# add entry in the /etc/fstab to mount the kdump lun
+echo "$luns_dev_mapper /var/crash xfs defaults 0 0" >> /etc/fstab
+ExitIfFailed $? "Unable to add lun mount entry in /etc/fstab"
+
+# mount the file system mentioned in the /etc/fstab file
+mount -a
```
jaiawasthi, line 86: just mount this device, because incase storage devices were failing, it would fail the complete operation

tools/enable-kdump.sh
prtyag: This happens with fdisk -l as discussed offline with you but here i'm using sg_scan so the hang problem i've not see in my testing. So i think we are good with this.

PR #675, opened 2020-07-29
Output all info about resource group and storage accounts

YunziZhangAzure:
1. Output storage accounts, containers and file share
2. Support blob container for sap bits

### Test
See the sap_library.json as input
Basically,
If enable_deployment = true, the corresponding resource will be tracked 
If enable_deployment = false, the corresponding resource(the block) will be ignored, no matter if the resource exists or not.
If is_existing = true, the resource will be imported and will not be created again.
If is_existing = false, the resource will be created, no matter if the resource exists or not.

There are a bunch of use cases, just list some of them
Use case:
1. Customer has existing sapbits storage account but doesn't have blob container and file share:

 "storage_account_sapbits": {
        "is_existing": "true",
        "arm_id": "xx/xx...",
...
        "file_share": {
            "enable_deployment": true,
            "is_existing": "false",
            "name": "bits"
        },
        "sapbits_blob_container": {
            "enable_deployment": true,
            "is_existing": "false",
            "name": "bits-container"
        }
}
Result:
file share and sapbits_blob_container will be created in the existing sapbits storage account

2. Customer dosn't have existing sapbits storage account, file share and blob container yet, wants to have them all.

 "storage_account_sapbits": {
        "is_existing": "false",
        "arm_id": "xx/xx...",
...
        "file_share": {
            "enable_deployment": true,
            "is_existing": "true",
            "name": "bits"
        },
        "sapbits_blob_container": {
            "enable_deployment": true,
            "is_existing": "false",
            "name": "bits-container"
        }
}
Result:
storage account, sapbits_blob_container and file share will be created.

3. However, there are some invalid combination,
 eg: storage_account_sapbits.is_existing = false, sapbits_blob_container.is_existing = true


deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
```diff
@@ -9,11 +9,11 @@ function validate_deployer_tfstate(){
 
     echo "Check if terraform.tfstate of deployer exists:"
 
-    if [ ! -f "${deployer_terraform_tfstate_path}" ]; then
-        echo "${deployer_terraform_tfstate_path} does not exist."
+    if [ ! -f "${deployer-terraform-tfstate-path}" ]; then
```
rukawata, line 12: What is the reason for this change from `_` to `-`?

deploy/terraform/bootstrap/sap_library/output.tf
```diff
@@ -1,11 +1,11 @@
-output "sap-library-resource-group-name" {
-  value = module.sap_library.rgName
+output "sap-library-resource-group" {
+  value = module.sap_library.sap-library-resource-group
```
rukawata, line 2: This information can be pulled from `sapbits-storage-account.resource_group_name`, therefore it is not necessary to save it to output.

deploy/terraform/terraform-units/modules/sap_library/outputs.tf
```diff
@@ -1,11 +1,11 @@
...
-    value = local.sa_tfstate_exists ? data.azurerm_storage_account.storage-tfstate[0].name : azurerm_storage_account.storage-tfstate[0].name 
+output "tfstate-storage-account" {
+    value = local.sa_tfstate_exists ? data.azurerm_storage_account.storage-tfstate[0] : azurerm_storage_account.storage-tfstate[0] 
 }
 
-output "sapbits-storage-account-name" {
-    value = local.sa_sapbits_exists ? data.azurerm_storage_account.storage-sapbits[0].name : azurerm_storage_account.storage-sapbits[0].name 
+output "sapbits-storage-account" {
+    value = local.sa_sapbits_exists ? data.azurerm_storage_account.storage-sapbits[0] : azurerm_storage_account.storage-sapbits[0] 
 }
```
rukawata, line 11: Can we add `azurerm_storage_share.fileshare-sapbits` and `azurerm_storage_container.storagecontainer-sapbits` to output? We need the `file_share_name`/`blob_container_name` in `output.json`.

deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
YunziZhangAzure: hmm... just want to be same with other variables in this file. Sometimes we use underscore, sometimes we use hyphen, I'm thinking we can have a discussion about this and put some instruction in our code guideline.

deploy/terraform/bootstrap/sap_library/output.tf
YunziZhangAzure: Good point, will remove it.

deploy/terraform/terraform-units/modules/sap_library/outputs.tf
YunziZhangAzure: I can add azurerm_storage_share.fileshare-sapbits and azurerm_storage_container.deployer, azurerm_storage_container.tfstate, azurerm_storage_container.saplibrary in this pr. 

I want to implement blob container of sapbits in next pr. I will add azurerm_storage_container.storagecontainer-sapbits in that pr.

Will make the output.tf name consistent.

deploy/terraform/terraform-units/modules/sap_library/storage_accounts.tf
```diff
@@ -4,98 +4,131 @@
...
-  resource_group_name       = local.rg_exists ? data.azurerm_resource_group.library[0].name : azurerm_resource_group.library[0].name
-  location                  = local.rg_exists ? data.azurerm_resource_group.library[0].location : azurerm_resource_group.library[0].location
-  account_replication_type  = "LRS"
+  resource_group_name       = local.rg_library.name
+  location                  = local.rg_library.location
+  account_replication_type  = local.sa_tfstate_account_replication_type
   account_tier              = local.sa_tfstate_account_tier
   account_kind              = local.sa_tfstate_account_kind
   enable_https_traffic_only = local.sa_tfstate_enable_secure_transfer
+  allow_blob_public_access  = true
```
rukawata, line 23: Any reason to set it to true?

YunziZhangAzure: https://docs.microsoft.com/en-us/answers/questions/40660/azure-storage-account-blob-service-sas-connectivit.html
If we don't set it as true, the blob container can not use 'blob' as type and it will only support 'private', i.e. no public access.

YunziZhangAzure:
> Overall looks good!
> One more question other than the comments, do we want to support to deploy both blob and azure file for sapbits? My assumption is we only deploy/require one of them.

Most use cases I can think of is to require only one of these two. Only the one with enable_deployment = true will be tracked. The other one can be set enable_deployment = false. This depends on the configuration in sap_library.json

rukawata:
> > Overall looks good!
> > One more question other than the comments, do we want to support to deploy both blob and azure file for sapbits? My assumption is we only deploy/require one of them.
> 
> Most use cases I can think of is to require only one of these two. Only the one with enable_deployment = true will be tracked. The other one can be set enable_deployment = false. This depends on the configuration in sap_library.json

That makes sense. Can we set one of them to be false by default in sap_library.json? This way we minimize the modification of the json before the run :)

YunziZhangAzure:
> > > Overall looks good!
> > > One more question other than the comments, do we want to support to deploy both blob and azure file for sapbits? My assumption is we only deploy/require one of them.
> > 
> > 
> > Most use cases I can think of is to require only one of these two. Only the one with enable_deployment = true will be tracked. The other one can be set enable_deployment = false. This depends on the configuration in sap_library.json
> 
> That makes sense. Can we set one of them to be false by default in sap_library.json? This way we minimize the modification of the json before the run :)

Done! By default, the file share will be provisioned.

If file share is to be provisioned, the output of its name is the name specified in the sap_library.json; The output of storagecontainer_sapbits is null.

If blob container is to be provisioned, the output of file share's name is ""(empty string), the output of storagecontainer_sapbits is the resource.

I think the downstream can use this information to determine where is the sapbits.

PR #676, opened 2020-07-30
Import sap bits information from sap_library tfstate

rukawata:
## Problem
The SA for sapbits currently is provisioned with sap_library

## Solution
Read SA information from remote tfstate

## Test
- Test with Azure file:
1. Deploy SA with Azure file
1. Deploy sap_system with just infrastructure
1. Check storage part from output.json:
```
"storage_account_sapbits": {
      "blob_container_name": null,
      "file_share_name": "bits",
      "name": "sapbits8198c957",
      "storage_access_key": "xxx"
    }
```

- Test with Azure blob:
1. Deploy SA with blob
1. Deploy sap_system with just infrastructure
1. Check storage part from output.json:
```
"storage_account_sapbits": {
      "blob_container_name": "bits-container",
      "file_share_name": null,
      "name": "sapbitsb1bf5933",
      "storage_access_key": "xxx"
    }
```

- Test with both Azure file and blob:
1. Deploy SA with both azure file and blob
1. Deploy sap_system with just infrastructure
1. Check storage part from output.json
```
"storage_account_sapbits": {
      "blob_container_name": "bits-container",
      "file_share_name": "bits",
      "name": "sapbitsb4cea599",
      "storage_access_key": "xxx"
    }
```

### DIFF
--- a/deploy/terraform/run/sap_system/module.tf
+++ b/deploy/terraform/run/sap_system/module.tf
@@ -7,6 +7,10 @@ module "deployer" {
   source = "../../terraform-units/modules/sap_system/deployer"
 }
 
+module "saplibrary" {
+  source = "../../terraform-units/modules/sap_system/saplibrary"
+}
+
 module "common_infrastructure" {
   source              = "../../terraform-units/modules/sap_system/common_infrastructure"
   is_single_node_hana = "true"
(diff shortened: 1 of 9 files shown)


deploy/terraform/run/sap_system/module.tf
```diff
@@ -7,6 +7,10 @@ module "deployer" {
   source = "../../terraform-units/modules/sap_system/deployer"
 }
 
+module "saplibrary" {
+  source = "../../terraform-units/modules/sap_system/saplibrary"
+}
+
```
YunziZhangAzure, line 13: Why do we need to invoke these two modules here? deployer and saplibrary.

rukawata: deployer is to import resources got deployed with deployer, mainly the admin vnet.

PR #679, opened 2020-08-03
Make *_tfstate_scrpt.sh idempotent

rukawata:
## Problem
The `deployer_tfstate_script.sh` and `saplibrary_tfstate_script.sh` can't be run repeatedly.

## Solution
Add validations/conditions to make it idempotent, includes:
- cleanup `.terraform` (this is especially reuiqred when switch to a different backend config)
- check if tfstate file(s) already exists remotely.

## Test
1. Deploy a sap_libaray.
1. Repeatedly run `deployer_tfstate_script.sh` and `saplibrary_tfstate_script.sh`

### DIFF
--- a/deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
+++ b/deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
@@ -39,6 +39,16 @@ function validate_deployer_tfstate(){
     fi
 }
 
+function validate_deployer_remote_state(){
+
+    az login --identity > /dev/null
+    remote_state_exists=$(az storage blob exists -c ${storagecontainer_deployer_name} --name ${deployer_tfstate_name} --account-name ${tfstate_storage_account_name} | jq -r .exists)
+    if $remote_state_exists; then
+        echo "remote state ${deployer_tfstate_name} already exists"
+        exit 0
+    fi
+}
+
 function deployer_remote_backend_init(){
 
     echo "Start initialize remote backend for deployer"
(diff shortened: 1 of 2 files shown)


deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
```diff
@@ -47,13 +57,19 @@ function deployer_remote_backend_init(){
...
 
+    rm -r .terraform
+
     terraform init -force-copy \
     -backend-config "resource_group_name=${saplibrary_resource_group_name}" \
     -backend-config "storage_account_name=${tfstate_storage_account_name}" \
     -backend-config "container_name=${storagecontainer_deployer_name}" \
     -backend-config "key=${deployer_tfstate_name}"
 
+    validate_deployer_remote_state
```
YunziZhangAzure, line 68: I'm thinking these tfstate file should be able to be override if it is uploaded again.

rukawata: If the tfstate already exists, terraform push errors out:
```
Acquiring state lock. This may take a few moments...
Failed to write state: cannot import state with serial 24 over newer state with serial 25
```

YunziZhangAzure: I see. 

What do you think of the following logic?
Before terraform init, check if the tfstate file exists in the SA, if exists, then remove it from SA.

In this way, the script is idempotent, every execution a new one.

rukawata: IMO, once the tfstate is in the SA, it should be only updated by terraform apply with remote backend.

Thinking about below situation:
A just updated `saplibrary.tfstate` stored in SA, with running terraform apply with remote backend, from one deployer.
B tries to run `saplibrary_tfstate_script.sh` from a **different** deployer, with an outdated `saplibrary.tfstate`.

We probably don't want the outdated saplibrary.tfstate got updated into the SA.
And another thing to consider, if at all, for B, is to sync the local saplibrary.tfstate under ~/.config with remote.

PR #681, opened 2020-08-04
Copy saplibrary module from bootstrap to run

YunziZhangAzure:
sap_library under run will be used to manage the tfstate of sap_library

### Test
1. Under ~/sap-hana/deploy/terraform/run/sap_library/default/, execute ./saplibrary_tfstate_script.sh 
2. ~/sap-hana/deploy/terraform/run/sap_library$ terraform apply -var-file="../../bootstrap/sap_library/saplibrary.json"
Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

### DIFF
--- a/deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
+++ b/deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
@@ -35,6 +35,7 @@ function validate_deployer_tfstate(){
         exit 2
     else 
         echo "${deployer_terraform_tfstate_path} exists."
+        validate_deployer_remote_state
         deployer_remote_backend_init
     fi
 }
@@ -57,6 +58,8 @@ function deployer_remote_backend_init(){
 
     cd ..
 
+    echo "Clean .terraform if exists:"
+
     rm -r .terraform
 
     terraform init -force-copy \
@@ -64,8 +67,6 @@ function deployer_remote_backend_init(){
     -backend-config "storage_account_name=${tfstate_storage_account_name}" \
     -backend-config "container_name=${storagecontainer_deployer_name}" \
     -backend-config "key=${deployer_tfstate_name}"
-
-    validate_deployer_remote_state
 
     terraform state push "${deployer_tfstate_name}"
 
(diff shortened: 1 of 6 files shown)


deploy/terraform/bootstrap/sap_library/deployer_tfstate_script.tmpl
```diff
@@ -35,6 +35,7 @@ function validate_deployer_tfstate(){
         exit 2
     else 
         echo "${deployer_terraform_tfstate_path} exists."
+        validate_deployer_remote_state
```
rukawata, line 38: Assume below scenario:
1. A deployed saplibrary and stored remotely.
1. B did a fresh clone of the repo.

Now remote backend is not configured for B.
I assume we will have another pre-deployment script where we initialize the remote backend? 
If that is the case, then it is not a problem :)

YunziZhangAzure: Yes, I plan to have another script under run/sap_library. 

I think there are at least two use cases here.
For role A, the person who is responsible for initializing the whole saplibrary(create sa...). After role A executes saplibrary_tfstate_script.sh, the saplibrary.terraform.tfstate will exist in the storage account. same thing for deployer.terraform.tfstate.

After this step, no one can override tfstate file.s They can only 'modify' the tfstate files through making changes to the input(saplibrary.json) and after the resources are changed, the tfstate files will reflect these changes.

For Role B, users should have a local file saplibrary.json under run/saplibrary first, (This file can be downloaded from the existing storage account or copy from other place, as long as there is a such file), users can edit this file. 
Then users can execute a new script under run/saplibrary(my next pr), which will read storage accounts info from .config/sa_config.json, initialize the backend and execute terraform apply. This is how B can set up a terraform working directory if he has a fresh clone of the repo.

If A wants to modify the saplibrary resources (like role B), A should do the same thing like B which is to have saplibrary.json local and use the new script.

YunziZhangAzure: So basically saplibrary_tfstate_script.sh and default_tfstate_script.sh will be a one time use script I think.

PR #689, opened 2020-08-07
Add scripts to help users download input json from SA

YunziZhangAzure:
## Problem
Users need script to download input json from SA

## Solution
Shell scripts read storage account info from sa_config.json and use az cli to download file to the local

## Tests

- To test download_input.sh of sapsystem: go to /run/sap_system/ execute util/download_input.sh QA Z01; then check if the QA-Z01.json shows at the working directory

- To test download_input.sh of saplibrary: go to /run/sap_library/ execute util/download_input.sh; then check if the saplibrary.json shows at the working directory

- To test download_input.sh of sapdeployer: go to /run/sap_deployer/ execute util/download_input.sh; then check if the deployer.json shows at the working directory

## Notes

### DIFF
--- a/deploy/terraform/run/sap_deployer/util/download_input.sh
+++ b/deploy/terraform/run/sap_deployer/util/download_input.sh
@@ -24,9 +24,10 @@ function main(){
     local storage_account_name=$(read_json .deployer.storage_account_name)
     local container_name=$(read_json .deployer.container_name)
     local remote_file_name="deployer.json"
+    local remote_file_path="deployer.json"
     local local_file_path="${local_file_dir}${remote_file_name}"
 
-    json_download ${local_file_path} ${storage_account_name} ${container_name} ${remote_file_name}    
+    json_download ${local_file_path} ${storage_account_name} ${container_name} ${remote_file_path}    
 }
 
 function read_json(){
@@ -75,13 +76,23 @@ function json_download(){
     local local_file_path=$1
     local storage_account_name=$2
     local container_name=$3
-    local remote_file_name=$4
-
-    echo "Start downloading file: ${remote_file_name}"
+    local remote_file_path=$4
 
     az login --identity > /dev/null
 
-    local cmd="az storage blob download --container-name ${container_name} --file ${local_file_path} --name ${remote_file_name} --account-name ${storage_account_name}"
+    echo "Check if ${remote_file_path} exists in storage accounts"
+
+    remote_state_exists=$(az storage blob exists -c ${container_name} --name ${remote_file_path} --account-name ${storage_account_name} | jq -r .exists)
+    if [ $remote_state_exists = true ]; then
+        echo "remote file ${remote_file_path} exists"
+    else
+        echo "remote file ${remote_file_path} does not exist"
+        exit 1
+    fi
+
+    echo "Start downloading file: ${remote_file_path}"
+
+    local cmd="az storage blob download --container-name ${container_name} --file ${local_file_path} --name ${remote_file_path} --account-name ${storage_account_name}"
     eval "$cmd"
 }
 
--- a/deploy/terraform/run/sap_deployer/util/download_input.sh
+++ b/deploy/terraform/run/sap_deployer/util/download_input.sh
@@ -24,6 +24,7 @@ function main(){
     local storage_account_name=$(read_json .deployer.storage_account_name)
     local container_name=$(read_json .deployer.container_name)
     local remote_file_name="deployer.json"
+    # The path of remote file can be updated based on actual needs
     local remote_file_path="deployer.json"
     local local_file_path="${local_file_dir}${remote_file_name}"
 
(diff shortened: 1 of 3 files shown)


deploy/terraform/run/sap_system/util/download_input.sh
```diff
@@ -0,0 +1,101 @@
...
+    local remote_file_name="${workspace}-${sid}.json"
+    local remote_file_path="${workspace}/${sid}/${remote_file_name}"
+    local local_file_path="${local_file_dir}${remote_file_name}"
+
+    json_download ${local_file_path} ${storage_account_name} ${container_name} ${remote_file_path}    
+}
+
+function validate_arguments(){
+
+    if ["$#" -ne 2 ]: then
```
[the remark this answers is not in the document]
YunziZhangAzure: oops! Good catch!

deploy/terraform/run/sap_system/util/download_input.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "Start downloading file:"
+    local local_file_path=$1
+    local storage_account_name=$2
+    local container_name=$3
+    local remote_file_path=$4
+
+    az login --identity > /dev/null
+
+    local cmd="az storage blob download --container-name ${container_name} --file ${local_file_path} --name ${remote_file_path} --account-name ${storage_account_name}"
+    eval "$cmd"
```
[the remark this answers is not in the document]
YunziZhangAzure: ok! Will add it.

deploy/terraform/run/sap_system/util/download_input.sh
```diff
@@ -0,0 +1,101 @@
...
+    local workspace=$1
+    local sid=$2
+    
+    check_jq_installed
+
+    check_file_exists ${target_json}
+
+    local storage_account_name=$(read_json .saplibrary.storage_account_name)
+    local container_name="sapsystem"
+    local remote_file_name="${workspace}-${sid}.json"
```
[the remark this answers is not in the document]
YunziZhangAzure: Like our discussion before, I thought this json is like a resource instead of local variables, so I use "-". What do you think?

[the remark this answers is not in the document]
rukawata: This is the name of the blob file, do we consider this as a resource name?
The reason I'm asking is that, I would like to have the folder name on deployer to be consist with the json file name (the folder for ansible inventory). It seems somehow odd to have a folder on a linux VM with naming `PROD-HN1`?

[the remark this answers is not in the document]
YunziZhangAzure: I see. I can change it to underscore, as long as we make it consistent.

deploy/terraform/run/sap_deployer/util/download_input.sh
```diff
@@ -0,0 +1,99 @@
...
+    local file_path="$1"
+    
+    if [ ! -f "${file_path}" ]; then
+        printf "%s\n" "ERROR: File ${file_path} does not exist. Please follow guidance to recover it" >&2
+        # TODO: create a guidance about file/env recovery
+        exit 1
+    fi
+}
+
+function check_jq_installed(){
```
ben-moss, line 52: Rather than redefine this function in all 3 scripts, why not just source the original and pass in specific the args?: https://github.com/Azure/sap-hana/blob/beaa7b49207951f94e3b77fafee3c488e58d3c85/util/common_utils.sh#L66

ben-moss: It seems like a maintenance headache to copy and paste the same function into several scripts, but maybe there's a good reason? 😄

YunziZhangAzure: LOL, I know. I have thought about this, but because I thought our structure of repo may have a big change (very soon), and the util/* files may be distributed to different directory. I just want to decouple the scripts I'm working on and make them functionally correct at this time but we can definitely refine these scripts later. 

Thanks for comment!

deploy/terraform/run/sap_deployer/util/download_input.sh
```diff
@@ -0,0 +1,99 @@
...
+function main(){
+    
+    check_jq_installed
+
+    check_file_exists ${target_json}
+
+    local storage_account_name=$(read_json .deployer.storage_account_name)
+    local container_name=$(read_json .deployer.container_name)
+    local remote_file_name="deployer.json"
+    local remote_file_path="deployer.json"
```
rukawata, line 27: what is the difference between `remtoe_file_name` and `remote_file_path`?
It make sense in deploy/terraform/run/sap_system/util/download_input.sh, but may be not so much here?

deploy/terraform/run/sap_library/util/download_input.sh
```diff
@@ -75,13 +76,23 @@ function json_download(){
...
     az login --identity > /dev/null
 
-    local cmd="az storage blob download --container-name ${container_name} --file ${local_file_path} --name ${remote_file_name} --account-name ${storage_account_name}"
+    echo "Check if ${remote_file_path} exists in storage accounts"
+
+    remote_state_exists=$(az storage blob exists -c ${container_name} --name ${remote_file_path} --account-name ${storage_account_name} | jq -r .exists)
+    if [ $remote_state_exists = true ]; then
+        echo "remote file ${remote_file_path} exists"
+    else
+        echo "remote file ${remote_file_path} does not exist"
```
rukawata, line 89: Maybe print out more information? eg. SA name, container name.
This can be helpful in trouble shooting.

deploy/terraform/run/sap_system/util/download_input.sh
```diff
@@ -85,14 +85,25 @@ function check_jq_installed(){
...
+    if [ $remote_state_exists = true ]; then
+        echo "remote file ${remote_file_path} exists"
+    else
+        echo "remote file ${remote_file_path} does not exist"
+        exit 1
+    fi
+
+    echo "Start downloading file:"
 
     local cmd="az storage blob download --container-name ${container_name} --file ${local_file_path} --name ${remote_file_path} --account-name ${storage_account_name}"
```
rukawata, line 108: Considering add --output none to az cli command. This way we limit the output to only show error.
Can be done in a separate PR since there are several places.

deploy/terraform/run/sap_deployer/util/download_input.sh
rukawata: I see this can be done in the future when the folder structure with remote state gets stable :)

deploy/terraform/run/sap_library/util/download_input.sh
YunziZhangAzure: Sure! Will add more info.

deploy/terraform/run/sap_deployer/util/download_input.sh
YunziZhangAzure: The path of remote file can be updated based on actual needs.
Currently the name and the path are same. They can be different if needed in the future.

deploy/terraform/run/sap_system/util/download_input.sh
YunziZhangAzure: Will be done in a separate pr.

PR #695, opened 2020-08-11
force a new deployment of json if the content of input json changed

YunziZhangAzure:
## Problem
Currently, using "source" at azurerm_storage_blob will not lead to the redeployment of the input json even if the json is changed

## Solution
Change "source" to "source_content", then after the content of json is changed, it will be redeployed.

## Tests

Edit .config/deployer.json
Then under /run/sap_library, execute util/terraform.sh
image

Edit /run/sap_library/saplibrary.json
Then under /run/sap_library, execute util/terraform.sh
image

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/remote_file.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/remote_file.tf
@@ -9,7 +9,10 @@ resource "azurerm_storage_blob" "deployer_json" {
   storage_account_name   = local.sa_tfstate.name
   storage_container_name = local.storagecontainer_deployer.name
   type                   = "Block"
-  source                 = pathexpand("~/.config/${local.storagecontainer_deployer.name}.json")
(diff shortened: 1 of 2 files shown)


deploy/terraform/terraform-units/modules/sap_library/remote_file.tf
```diff
@@ -9,7 +9,7 @@ resource "azurerm_storage_blob" "deployer_json" {
   storage_account_name   = local.sa_tfstate.name
   storage_container_name = local.storagecontainer_deployer.name
   type                   = "Block"
-  source                 = pathexpand("~/.config/${local.storagecontainer_deployer.name}.json")
+  source                 = file(pathexpand("~/.config/${local.storagecontainer_deployer.name}.json"))
```
rukawata, line 12: Also `source_content`? With comments?

YunziZhangAzure: Good catch, changed!

PR #697, opened 2020-08-12
Set default location to eastus for deployer

rukawata:
## Problem
`sap_library` and `sap_system` use `eastus` as default region.

## Solution
Change `deployer.json` to use `eastus` as default.

## Tests
N/A

## Notes
N/A


YunziZhangAzure:
Maybe change the default value as well?
https://github.com/Azure/sap-hana/blob/4cd2c987492fbe431c19dbe5b3151b765bd466e6/deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf#L41

PR #703, opened 2020-08-17
Mask output date for pipeline

rukawata:
## Problem
output contains sensitive data

## Solution
set `sensitive = true`

## Tests
1. cd ~/sap-hana/deploy/terraform/bootstrap/sap_deployer
1. terraform init && terraform apply -auto-approve -var-file=deployer.json
1. Check output:
```
Outputs:

deployer = <sensitive>
deployer_id = <sensitive>
deployer_uai = <sensitive>
nsg_mgmt = <sensitive>
subnet_mgmt = <sensitive>
vnet_mgmt = <sensitive>

```
## Notes


rukawata:
> If setting output as sensitive, does it mean users can only read tfstate file if they want to know the ip of deployer, storage account...?



> If setting output as sensitive, does it mean users can only read tfstate file if they want to know the ip of deployer, storage account...?

No, it only means they are masked during terraform apply ([more info](https://www.terraform.io/docs/configuration/outputs.html)).
The main reason is I want those information to be masked in pipeline.

PR #714, opened 2020-08-21
Add naming convention to sap_library

KimForss:
## Problem
Apply the default naming convention

## Tests
Deploy and validate

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
@@ -1,21 +1,62 @@
 // Input arguments 
+variable "region_mapping" {
+  type        = map(string)
+  description = "Region Mapping: Full = Single CHAR, 4-CHAR"
+
+  # 28 Regions 
+
+  default = {
+    westus             = "weus"
+    westus2            = "wus2"
+    centralus          = "ceus"
+    eastus             = "eaus"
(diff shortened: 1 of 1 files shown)


deploy/terraform/bootstrap/sap_library/backend.tf
```diff
@@ -5,6 +5,6 @@
 */
 terraform {
   backend "local" {
-    path = "../../../../../.config/saplibrary.terraform.tfstate"
+    path = "~/.config/saplibrary.terraform.tfstate"
```
rukawata, line 8: This is removed in PR #716, hence not in scope of this PR. Please revert this change.

deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
```diff
@@ -1,21 +1,62 @@
...
 locals {
 
     // Infrastructure
     var_infra = try(var.infrastructure, {})
 
     // Region
-    region = try(local.var_infra.region, "eastus")
-
+    region         = try(local.var_infra.region, "")
+    environment    = try(var.infrastructure.environment, "")
```
rukawata, line 47: Please keep the key landscape for the time being.
This is used in other PRs as well as existing code in feature branch.
If there is a reason to change that and we all agree, let's do that in a separate PR.

PR #731, opened 2020-08-28
Remove jumpbox stanza

YunziZhangAzure:
## Problem
Current input json contain jumpbox stanza

## Solution
Remove the jumpbox stanza;

Modify the local variable, make it contain an empty default

## Tests
Plan: 82 to add, 0 to change, 0 to destroy.

## Notes

### DIFF
--- a/deploy/terraform/run/sap_system/variables_global.tf
+++ b/deploy/terraform/run/sap_system/variables_global.tf
@@ -16,8 +16,8 @@ variable "infrastructure" {
 variable "jumpboxes" {
   description = "Details of the jumpboxes"
   default     = {
-    windows = {},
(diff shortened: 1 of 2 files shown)


deploy/terraform/run/sap_system/variables_global.tf
```diff
@@ -15,6 +15,10 @@ variable "infrastructure" {
 
 variable "jumpboxes" {
   description = "Details of the jumpboxes"
+  default     = {
+    windows = {},
+    linux = {}
```
rukawata, line 20: My bad, those two should be list:
```suggestion
    windows = [],
    linux = []
```

PR #744, opened 2020-09-03
Adjust ansible pipeline to use uppercase

rukawata:
## Problem
Required to use uppercase in workspaces.

## Solution
Change to use upper case in ansible test as well.

## Tests
https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=9666&view=results

## Notes
RHEL fail due to a problem with image.


scites-ms (commented):
Although I am not familiar with this code base, it looks good to me.

YunziZhangAzure (approved):

PR #752, opened 2020-09-06
Adding the new naming module

KimForss:
## Problem
Create a naming convention module to keep resource naming central
## Solution
Create a separate module to manage naming conventions

## Tests

## Notes


deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+  location_short = upper(try(var.region_mapping[var.location], "unkn"))
+
+  environment_length = 5
+  sap_vnet_length    = 7
+
+  env_verified         = upper(substr(var.environment, 0, local.environment_length))
+  vnet_verified        = upper(substr(var.sap_vnet_name, 0, local.sap_vnet_length))
+  dep_vnet_verified    = upper(substr(var.deployer_vnet_name, 0, local.sap_vnet_length))
+  random-id_verified   = upper(substr(var.random-id, 0, 4))
+  random-id_verifiedvm = lower(substr(var.random-id, 0, 3))
```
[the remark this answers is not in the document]
rukawata: I had the same discussion with Yunzi, see [here](https://github.com/Azure/sap-hana/pull/763#discussion_r486538881).

rukawata (commented):
Thanks for the PR!
First round of review and some comments :)

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+}
+
+variable codename {
+  description = "Code name of application (optional)"
+  default     = ""
+}
+
+variable deployer_vnet_name {
+  description = "Name of Deployer VNet"
+}
```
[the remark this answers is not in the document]
KimForss: fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+
+variable sap_vnet_name {
+  description = "Name of SAP VNet"
+}
+
+variable sap_sid {
+  description = "SAP SID"
+}
```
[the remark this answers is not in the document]
KimForss: sap_sid is mandatory so that we can create the SDU resource group name.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+
+variable db_sid {
+  description = "Database SID"
+}
```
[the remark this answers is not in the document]
KimForss: For anydb the dbsid becomes ora, db2, sql, ase etc - It is mandatory for the any db naming

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+
+variable db_server_max_count {
+  type    = number
+  default = 10
+}
+
+variable azlimits {
+  description = "Name length for resources"
+  default = {
```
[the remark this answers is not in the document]
KimForss: Added a ToDo

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+    "ukwest"             = "ukwe"
+    "westcentralus"      = "wcus"
+    "westeurope"         = "weeu"
+    "westindia"          = "wein"
+    "westus"             = "weus"
+    "westus2"            = "wus2"
+  }
+}
+
+variable resource-extension {
```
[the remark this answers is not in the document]
KimForss: Fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+    "web-subnet"          = "_web-subnet"
+    "web-subnet-nsg"      = "_webSubnet-nsg"
+  }
+}
+
+locals {
+  location_short = upper(try(var.region_mapping[var.location], "unkn"))
+
+  environment_length = 5
+  sap_vnet_length    = 7
```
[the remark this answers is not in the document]
KimForss: Good one, I'll add one

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,205 @@
...
+
+variable app_server_max_count {
+  type    = number
+  default = 25
+}
+
+variable db_server_max_count {
+  type    = number
+  default = 10
+}
```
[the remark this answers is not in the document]
KimForss: This controls how many names the naming module will put in the arrays defined in vm.tf, I agree the variable name is misleading. We could also pass in the app server count, web dispatch count, scs count and db count so that the name arrays would have the exact number

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
KimForss: Fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -0,0 +1,42 @@
+locals {
+
+  db_oscode       = upper(var.db_ostype) == "LINUX" ? "l" : "w"
+  app_oscode      = upper(var.app_ostype) == "LINUX" ? "l" : "w"
+  db_platformcode = substr(var.db_platform, 0, 3)
+
+
+  //Deployer
+  deployer_names = [for idx in range(var.app_server_max_count) :
```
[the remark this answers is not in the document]
KimForss: maybe best is to pass in the parameter from the input parameters

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
KimForss: The naming standard today has both 3 and 4 - I agree having 3 across the board would be more consistent

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
KimForss: I changed the code to have defaults with the ability to override. As we will have the server count in input that will resolve this

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
rukawata: Can you add the ToDo in the code as comment?
eg. //Todo: add reference doc for resource abbreviations.

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
```diff
@@ -0,0 +1,19 @@
+output prefixes {
```
rukawata, line 1: One thought, can we just output one as a nested map? This way, when you pass those to modules, only have to pass one variable.

```
{
	"prefixes": {
		"SDU": local.sdu_name,
		"DEPLOYER": local.deployer_name,
		...
	},
	"sa_name": {
		"SDU": local.sdu_name,
		"DEPLOYER": local.deployer_name,
		...
	},
	...
}
```

rukawata (commented):
Thanks for addressing most of the comments!

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
KimForss: I prefer passing variables of the base type ie primitives (string, list etc) and not complex variables - I believe that it makes the code more readable and type checking easier for non developers, ie if we pass a complex type then the code using it would need to understand the structure, for instance using your example above the code for getting the SDU name would need to refer to var.prefixes["SDU"] instead of just var.prefixes.

Having one complex variable with x primitive items (string, list, number) is not more efficient than having the x primitive items as parameters/variables

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
KimForss: done

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
YunziZhangAzure: Map is not a complex type. If you look at some other projects, when service A calls service B, service should provide an interface doc which describes the arguments and the return type. Almost every big project uses such design. Same for us. Think of the naming component is service B, sap_deployer, Sap_library, Sap_system... are service A which trys to call service B.

There are several disadvantages not using this interface design.
1. In modult.tf, the number of fields is doubled. Whenever there is a resource, a corresponding naming-resource field is needed.
2. It mix the functional components(sap_deployer, Sap_library, Sap_system) with util component(naming module). Ideally, we hope they are loose coupling.

KimForss: For real software where the components are loosely coupled and built as services and the parameters are serialised to get efficiency I 100% agree - however Terraform is not that :) Terraform is essentially just a parsed language that gets read into memory at run time - there is no loose coupling between modules.

I agree that the map is a not a complex type but if the map has values that are of different types it is somewhat complex, If the prefix is a string but the servernames is a list then the consumer (the code in the module) of it needs to know that. 

I feel we compromise the readability of the code for reducing the number of parameters. When all of this becomes hidden in the RP that is our responsibility but now when we have customers/partners using the repo I would rather have clarity even if it requires multiple fields

However I am not adamant on my design - if you both agree we can change it

mkdeegan: It will always be important to preserve readability. Sometimes this is a judgement call, but if the code obfuscates what is happening and makes it difficult to understand, thin it must be accompanied by a text block that spells it out in a way understandable to a lay person.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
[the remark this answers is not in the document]
KimForss: After discussion with Morgan I changed it to 3 across the board

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
KimForss: changed to a map

KimForss: Changing the variable to a map caused two issues.

1. Terraform crashes, when looking through the crash log this is due to the map elements being of different types. The prefix is a string, the virtual machine, key vault and storage account names are lists and the resource extension is a map

2. The naming convention is using the random id in the virtual machine, key vault and storage account names. The "keeper" of the random id is the prefix, i.e. as long as the prefix remains the same terraform does not recalculate the unique id. But if we define the keeper to be the prefix (which is a part of the map) we get a circular reference.

I propose we revert back to use the separate output variables instead of the map

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -0,0 +1,45 @@
+locals {
+
+  db_oscode       = upper(var.db_ostype) == "LINUX" ? "l" : "w"
+  app_oscode      = upper(var.app_ostype) == "LINUX" ? "l" : "w"
+  db_platformcode = substr(var.db_platform, 0, 3)
+
+
+  //Deployer
+  deployer_names = [for idx in range(var.app_server_count) :
```
rukawata, line 9: nit: to have both `deployer_name` and `deployer_names` can be quite confusing.
I assume this is for vm, so maybe `deployer_vm_names` would be more appropriate?

deploy/terraform/terraform-units/modules/sap_namegenerator/keyvault.tf
```diff
@@ -0,0 +1,12 @@
+locals {
+
+  sdu_kvname = [format("%s%s%s%sp%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid, local.random-id_verified),
```
rukawata, line 3: nit: I would suggest to use eg. `sdu_kv_name`, `deployer_kv_name`,...to follow the rhythm of other variables.

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
```diff
@@ -0,0 +1,19 @@
...
+  value = tomap({ "SDU" = local.sdu_name, "DEPLOYER" = local.deployer_name, "VNET" = local.vnet_name, "LIBRARY" = local.library_name })
+}
+
+output storageaccount_names {
+  value = tomap({ "SDU" = [local.sdu_sa_name], "DEPLOYER" = [local.deployer_sa_name], "VNET" = [local.vnet_sa_name], "LIBRARY" = [local.library_sa_name,local.state_sa_name] })
+}
+
+output resource_extensions {
+  value = var.resource_extension
+}
```
rukawata, line 11: nit: In module this is used as `resource_suffixes`, which makes more sense. Maybe change this variable also to `resource_suffixes` to be consistent?

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -0,0 +1,233 @@
...
+
+variable sap_sid {
+  description = "SAP SID"
+}
+
+variable db_sid {
+  description = "Database SID"
+}
+
+variable random-id {
```
rukawata, line 31: this is not related to this PR: now that random-id is no longer generated inside the terraform-unit module, but in the root module, please remember to remove that code form the terraform-unit module :)

rukawata (commented):
Some nit comments. Thanks!

rukawata (approved):
lgtm! Thanks for addressing the comments.

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
```diff
@@ -0,0 +1,34 @@
...
+      DEPLOYER = local.deployer_name,
+      SDU = local.sdu_name,
+      VNET = local.vnet_name, 
+      LIBRARY = local.library_name
+    }
+    storageaccount_names = {
+      DEPLOYER = [local.deployer_sa_name],
+      SDU = [local.sdu_sa_name],
+      VNET = [local.vnet_sa_name], 
+      LIBRARY = [local.library_sa_name,local.state_sa_name]
```
rukawata, line 13: Actually while I reviewing the deployer PR, another comment, can this be a nested map instead of list?
This way we avoid referring using a ambiguous index:
`name                      = local.storageaccount_names[0]`

Eg.
```
storageaccount_names = {
      DEPLOYER = {
        deployer_sa_name = local.deployer_sa_name
      }
      ...
      LIBRARY = {
        library_sa_name = local.library_sa_name,
        state_sa_name = local.state_sa_name
      }
```
And the above line can rewritten into:
`name                      = local.storageaccount_names.deployer_sa_name`

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
```diff
@@ -0,0 +1,38 @@
...
+      APP      = local.app_server_names,
+      DEPLOYER = local.deployer_vm_names,
+      HANA     = local.hana_server_names
+      HANA_HA  = local.hana_server_names_ha,
+      ISCSI    = local.iscsi_server_names
+      SCS      = local.scs_server_names,
+      WEB      = local.web_server_names,
+    }
+    resource_suffixes = var.resource_suffixes
+
```
rukawata, line 36: nit: empty line.
```suggestion
```

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
```diff
@@ -0,0 +1,38 @@
...
+    keyvault_names = {
+      DEPLOYER = local.deployer_keyvault_name,
+      LIBRARY  = local.library_keyvault_name
+      SDU      = local.sdu_keyvault_name,
+      VNET     = local.vnet_keyvault_name
+    }
+    virtualmachine_names = {
+      ANYDB    = local.anydb_server_names
+      ANYDB_HA = local.anydb_server_names_ha
+      APP      = local.app_server_names,
```
rukawata, line 27: Some key value pair has comma in the end, some don't, can you make sure it is consistent in this file?

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
```diff
@@ -0,0 +1,14 @@
...
+  deployer_name = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.dep_vnet_verified))
+  vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
+  library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
+
+  sdu_storageaccount_name      = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.vnet_verified, local.random-id_verified))
+  library_storageaccount_name  = lower(format("%s%s%ssaplib%s", local.env_verified, local.location_short, local.vnet_verified, local.random-id_verified))
+  terraformstate_storageaccount_name    = lower(format("%s%s%stfstate%s", local.env_verified, local.location_short, local.vnet_verified, local.random-id_verified))
+  deployer_storageaccount_name = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random-id_verified))
+  vnet_storageaccount_name     = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random-id_verified))
```
rukawata, line 12: nit: equal alignment.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
rukawata, line 31: Always use _ instead of - to be consistent with variable naming.

rukawata (commented):
Hopefully this is last round of comments :)
All are nit comments, please take a look.

deploy/terraform/terraform-units/modules/sap_namegenerator/output.tf
KimForss: fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
KimForss: fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -0,0 +1,45 @@
+locals {
+
+  db_oscode       = upper(var.db_ostype) == "LINUX" ? "l" : "w"
+  app_oscode      = upper(var.app_ostype) == "LINUX" ? "l" : "w"
+  db_platformcode = substr(var.db_platform, 0, 3)
+
+
+  //Deployer
```
rukawata, line 8: what is this comment for?

KimForss: removed it

rukawata (approved):

rukawata (commented):
Some additional thinking the naming for different modules can cause confusion.
Now we have to sets of names:
set 1: [`sap_deployer`, `sap_library`, `sap_landscape`, `sap_system`]
set 2: [`DEPLOYER`, `LIBRARY`, `VNET`, `SDU`]

This is confusing for people who reads the code.

We may or may not change the module name in the future, but to make more sense out of the code, I would suggest to use the name of the module in `sap_namegenerator`, basically replace set 2 with set 1. This way, one could clearly know which part the naming module is applied to.

The same applies to `virtualmachine_names` - should be consistent with the module names, which are `anydb_node`, `app_tier`, `hdb_node`. And I would suggest to have HA as input, just like the vm counts, and only produce one set of names for each module, meaning not to have `ANYDB` and `ANYDB_HA` - if this is possible.

This can be a new PR and will merge the current one.

Thanks,
Nnacy

PR #766, opened 2020-09-09
Supporting custom disk sizing

KimForss:
## Problem
Currently customers are not able to deploy custom disk configurations using our automation. We need to be able to support Ultra Disks for Log disks for AnyDB deployments. This will also require the we can specify zonal deployments.

## Solution
Provide a way for the customer to provide a variable with the path to a custom json file for disk configuration. This will introduce Ultra Disk support as well as Availability Zone support.

## Tests
Create a copy of the disk configuration json into the workspace directory and run

image

-var="disk_sizes=mydisksize.json"

## Notes


deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -216,10 +228,29 @@ locals {
...
+    for vm_counter, anydb_vm in local.anydb_vms : [
+      for storage_type in lookup(local.sizes, local.anydb_size).storage : [
+        for disk_count in range(storage_type.count) : {
+          name = format("%s-%s%02d", anydb_vm.name, storage_type.name, disk_count)
+        }
+      ]
+      if storage_type.disk_type == "UltraSSD_LRS"
+  ]])
+
+  enable_ultradisk = length(local.ultra_data_disks) > 0 ? true : false
```
rukawata, line 253: To create an `ultra_data_disks` seems over kill for the purpose to check if ultradisk is used.
Probably can directly check the disk_type in sizing json.

eg:
```
  storage_list = lookup(local.sizes, local.anydb_size).storage
  enable_ultradisk = try(compact([
    for storage in local.storage_list :
    storage.disk_type == "UltraSSD_LRS" ? true : ""
  ])[0], false)
```
From the logic, it seems if Ultra disk is used, it will be used for all VMs? Can you confirm that is the case?

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -241,4 +252,18 @@ locals {
       }
     ]
   ])
+
+  disk_count = length(local.data-disk-list) / length(local.hdb_vms)
```
rukawata, line 256: it can be simpler:
```suggestion
  disk_count =  length(local.data-disk-per-dbnode)
```
Another option would be to have the lun number in `local.data-disk-list`, so don't need to calculate later in the resource block.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf
```diff
@@ -184,12 +179,26 @@ resource "azurerm_linux_virtual_machine" "vm-dbnode" {
...
   managed_disk_id           = azurerm_managed_disk.data-disk[count.index].id
   virtual_machine_id        = azurerm_linux_virtual_machine.vm-dbnode[floor(count.index / length(local.data-disk-per-dbnode))].id
   caching                   = local.data-disk-list[count.index].caching
   write_accelerator_enabled = local.data-disk-list[count.index].write_accelerator_enabled
-  lun                       = count.index
-}
+
+  //Make sure the LUNs start from 0 for each VM
+  lun                       = count.index - local.disk_count * floor(count.index / length(local.data-disk-per-dbnode))
+}
```
rukawata, line 204: nit: newline.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -216,10 +228,29 @@ locals {
           name                      = format("%s-%s%02d", anydb_vm.name, storage_type.name, disk_count)
           storage_account_type      = storage_type.disk_type
           disk_size_gb              = storage_type.size_gb
+          disk-iops-read-write      = try(storage_type.disk-iops-read-write, 0)
```
rukawata, line 231: I only see this defined for anydb. Is this intended? Do we need that for hdb or app servers?
Probably would be helpful if you can add some comment here so other people understand the reason behind it :)

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -241,4 +252,18 @@ locals {
...
+    // Check if any disk is a Ultra Disk
+  ultra_data_disks = flatten([
+    for hdb_vm in local.hdb_vms : [
+      for datadisk in local.data-disk-per-dbnode : {
+        name                      = format("%s-%s", hdb_vm.name, datadisk.suffix)
+        }
+      if datadisk.storage_account_type == "UltraSSD_LRS"
+  ]])
+
+  enable_ultradisk = length(local.ultra_data_disks) > 0 ? true : false
```
rukawata, line 267: Same as above, overkill :)

deploy/terraform/run/sap_system/variables_local.tf
```diff
@@ -13,6 +13,17 @@ variable "scenario" {
   default     = "HANA Database"
 }
 
+variable "db_disk_sizes" {
```
rukawata, line 16: the example uses `disk_sizes`, but I assume the input variable should be either `db_disk_sizes` and/or `app_disk_sizes`?

rukawata (commented):
Thanks for working on this!

One big question I have is, to pass the customized json with -var option, can this be supported in TFC/TFE?
Other than that, this review is only on the disk sizing part. As discussed, will review the AV zone in a separate one :)

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -18,7 +18,18 @@ variable naming {
   description = "Defines the names for the resources"
 }
 
+variable "custom_disk_sizes" {
```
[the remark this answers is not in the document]
KimForss: Good catch - changed it to custom_disk_sizes_filename

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
KimForss: No this code checks if Ultra is defined and if so it will enable the VMs in this module to use Ultra

KimForss: I did not know how to query it simpler :)

rukawata: > No this code checks if Ultra is defined and if so it will enable the VMs in this module to use Ultra

Right, I guess my point is, if `enable_ultradisk` set to true, all VMs in this module will have this enabled? Sorry I did not express it clearly.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
KimForss: I was thinking of that but did not know how to do that as there is a nested loop

deploy/terraform/run/sap_system/variables_local.tf
KimForss: yes, fixed

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
KimForss: Will add for Hana as well

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -216,10 +228,29 @@ locals {
           name                      = format("%s-%s%02d", anydb_vm.name, storage_type.name, disk_count)
           storage_account_type      = storage_type.disk_type
           disk_size_gb              = storage_type.size_gb
+          disk-iops-read-write      = try(storage_type.disk-iops-read-write, 0)
+          disk-mbps-read-write      = try(storage_type.disk-mbps-read-write, 0)
```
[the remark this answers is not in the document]
KimForss: good call

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
rukawata: For hdb is it very straight forward since it has two nested loops:

```
data-disk-list = flatten([
    for hdb_vm in local.hdb_vms : [
      for idx, datadisk in local.data-disk-per-dbnode : {
        vm_name              = hdb_vm.name
        disk_name            = datadisk.disk_name
        storage_account_type = datadisk.storage_account_type
        lun                  = idx
      }
    ]
  ])
```

I can take a look at anydb case.

KimForss:
maybe we should change anydb to be the same structure

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -220,10 +227,21 @@ locals {
...
       if storage_type.name != "os"
   ]])
+
+  disk_count = length(local.anydb_disks) / length(local.anydb_vms)
+  
+  storage_list = lookup(local.sizes, local.anydb_size).storage
+  enable_ultradisk = try(compact([
+    for storage in local.storage_list :
+    storage.disk_type == "UltraSSD_LRS" ? true : ""
+  ])[0], false)
```
rukawata, line 246: This would be possible in 0.13 (sum only comes in 0.13):

```
  anydb_disks = flatten([
    for vm_counter, anydb_vm in local.anydb_vms : [
      for storage_counter, storage_type in local.storage_types : [
        for disk_counter, disk_count in range(storage_type.count) : {
          vm_index             = vm_counter
          name                 = format("%s-%s%02d", anydb_vm.name, storage_type.name, disk_count)
          storage_account_type = storage_type.disk_type
          lun                  = disk_counter + try(sum(slice(local.storage_types, 0, storage_counter)[*].count), 0)
        }
      ]
      if storage_type.name != "os"
  ]])
```

Before that, maybe 
1. separate into 2 the loops like in hdb module
2. Add comment with the potential change come in 0.13.

KimForss: Moved it to two looks like hdb module

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
```diff
@@ -33,15 +33,14 @@ resource "azurerm_linux_virtual_machine" "dbserver" {
   location            = var.resource-group[0].location
 
   //If more than one servers are deployed into a single zone put them in an availability set and not a zone
-  proximity_placement_group_id = local.zonal_deployment ? var.ppg[count.index % local.db_zone_count].id : var.ppg[0].id
+  proximity_placement_group_id = local.enable_ultradisk || local.zonal_deployment ? var.ppg[count.index % local.db_zone_count].id : var.ppg[0].id
```
rukawata, line 36: Sorry, I may need more help here to understand this condition.

Let's assume no zonal for anydb, then:
    - `db_zone_count` = 1
    - ppg count is 1
If ultra disk is enabled, we use ppg[0]
If ultra disk is not enabled, we still use ppg[0]

So if we have zonal, 
    - `db_zone_count` = x
    - ppg count is x
Regardless the value of `enable_ultradisk`, we use ppg[count.index]

It seems that bool `local.enable_ultradisk` has no impact on which ppg to use and is not required here.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
```diff
@@ -33,15 +33,14 @@ resource "azurerm_linux_virtual_machine" "dbserver" {
...
   //If more than one servers are deployed into a single zone put them in an availability set and not a zone
-  availability_set_id = local.zonal_deployment ? (
-    local.db_server_count == local.db_zone_count ? null : azurerm_availability_set.anydb[count.index % local.db_zone_count].id) : (
-    azurerm_availability_set.anydb[0].id
+  availability_set_id = local.enable_ultradisk || (local.zonal_deployment && (local.db_server_count == local.db_zone_count)) ? (
+    null) : (
+    azurerm_availability_set.anydb[count.index % local.db_zone_count].id
```
rukawata, line 40: I think I may need some help (maybe some docs) to understand this together with the logic for zone:
Does this mean that if ultradisk is enabled, we need zonal and no avset?

rukawata (commented):
Thanks! Just some questions on the conditions.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
KimForss: you are right

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
KimForss: Ultradisk requires zonal: https://docs.microsoft.com/en-us/azure/virtual-machines/disks-enable-ultra-ssd?tabs=azure-portal#ga-scope-and-limitations

rukawata (approved):
Thanks. Looks good now :)

PR #767, opened 2020-09-09
First iteration: Support password with deployer

### LINKED ISSUE #772: KeyVault: Deployer
**Introduction of KeyVault for the Deployer**
Create a two-tier secure store of credentials, following a DevOps & AppOps model.

**Describe the solution you'd like**
DevOps Secure Store is used for secrets that are not appropriate to be human known, except through a break-glass type access. Access to these credentials should be reserved for the Tools, Terraform and Ansible.

AppOps Secure Store is used for passwords that are used during application installation, and can be retrieved by Administrators.

The Deployer should be able to access the DevOps KV, using the MSI Credential, to retrieve a customer stored SPN, that is Mapped the the ENVIRONMENT. The SPN is loaded into the DevOps KV, after bootstrap and before re-initialization, typically by the InfoSec, IAM, or Orchestration Teams (Subscription Owner). The SSH Key pair to be used by Ansible is also stored here. 

There will also be an AppOps KV that will store any credentials that should be accessible to the administrators. At this time the same SSH Key will be stored here as well.

**Additional context**
Details provided in the KeyVault Specification Document

rukawata:
## Problem
Support generate and save password for deployer.
Resolves #772.

## Solution
By default, if `authentication.type` is set to `key`. If it is set to `password`, we will:
- by default generate and use random password.
- if `password` is provided in the authentication section, we use the provided password.

## Tests
Deploy with below section added in deployer.json - still uses SSH key pair.
```
  "deployers": [
    {
      "authentication": {
        "type": "key"
      }
    }
  ]
```

Deploy with below section - it creates random password and use that for the VM (also in post_deployment.sh)
```
  "deployers": [
    {
      "authentication": {
        "type": "password"
      }
    }
  ]
```

Deploy with below section - it uses the password provided for the VM (also in post_deployment.sh)
```
  "deployers": [
    {
      "authentication": {
        "type": "password",
       "password"： "2020Hey!"
      }
    }
  ]
```
## Notes
Limitation:

1. For security, the password will not be saved in post_deployment.sh, user have to retrieve the password from portal, then enter when execute the script.
2. Just like we always use one set  of SSH keypair for all deployer(s), we only support to take one password for all deployer(s) as well.

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/deployer_scp.tmpl
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer_scp.tmpl
@@ -5,26 +5,55 @@ ssh_timeout_s=10
 
 temp_file=$(mktemp)
 vault_name=${user_vault_name}
+
 ppk_name=${ppk_name}
-ppk=$(az keyvault secret show --vault-name $${vault_name} --name $${ppk_name} | jq -r .value)
-echo "$${ppk}" > $${temp_file}
+if [ ! -z $${ppk_name} ]
+then
+  printf "%s\n" "Collecting secrets from KV"
+  ppk=$(az keyvault secret show --vault-name $${vault_name} --name $${ppk_name} | jq -r .value)
+  echo "$${ppk}" > $${temp_file}
+fi
 
 printf "%s\n" "Create remote workspace if not exists"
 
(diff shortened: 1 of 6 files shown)


YunziZhangAzure (approved):
LGTM

PR #782, opened 2020-09-14
KV for saplandscape

YunziZhangAzure:
## Problem
Currently, only one key pair is used anywhere which is unable to meet the requirement of security and extensibility.

## Solution
1. Two keyvaults per sap landscape are created.

2. By default, Ansible ssh key for SID uses random generated public key. Provide sshkey.path_to_public_key and path_to_private_key overides it. They are stored in user kv.

3. Ansible ssh key for iSCSI uses random generated public key. Provide sshkey.path_to_public_key and path_to_private_key overides it. They are stored in user kv.

## Tests

## Notes
1. The output of the saplandscape hasn't been fully decided yet.

2. This pr contains some change on deployer module, thus please use this pr to provision deployer on devbox, so that the deployer-uai and deployer_user are included in deployer's tfstate file.

### DIFF
--- a/deploy/terraform/run/sap_landscape/saplandscape.json
+++ b/deploy/terraform/run/sap_landscape/saplandscape.json
@@ -21,6 +21,20 @@
                     "prefix": "10.1.0.0/24"
                 }
             }
+        },
+        "iscsi": {
+            "iscsi_count": 3,
+            "size": "Standard_D2s_v3",
+            "os": {
+                "publisher": "SUSE",
+                "offer": "sles-sap-12-sp5",
+                "sku": "gen1",
+                "version": "latest"
+            },
+            "authentication": {
+                "type": "key",
+                "username": "azureadm"
+            }
         }
     },
     "jumpboxes": {
--- a/deploy/terraform/run/sap_landscape/saplandscape.json
+++ b/deploy/terraform/run/sap_landscape/saplandscape.json
@@ -23,18 +23,7 @@
             }
         },
         "iscsi": {
-            "iscsi_count": 3,
-            "size": "Standard_D2s_v3",
-            "os": {
-                "publisher": "SUSE",
-                "offer": "sles-sap-12-sp5",
-                "sku": "gen1",
-                "version": "latest"
-            },
-            "authentication": {
-                "type": "key",
-                "username": "azureadm"
-            }
+            "iscsi_count": 3
         }
     },
     "jumpboxes": {
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
@@ -1,0 +1,151 @@
+/*
+  Description:
+  Set up key vault for sap landscape
+*/
+
+data "azurerm_client_config" "deployer" {}
+
+// Create private KV with access policy
+resource "azurerm_key_vault" "kv_prvt" {
+  name                       = local.kv_private_name
+  location                   = local.region
+  resource_group_name        = local.rg_exists? data.azurerm_resource_group.resource-group[0].name : local.rg_name
+  tenant_id                  = data.azurerm_client_config.deployer.tenant_id
+  soft_delete_enabled        = true
+  soft_delete_retention_days = 7
+  purge_protection_enabled   = true
+  sku_name                   = "standard"
+}
+
+resource "azurerm_key_vault_access_policy" "kv_prvt_msi" {
+  key_vault_id = azurerm_key_vault.kv_prvt.id
+
+  tenant_id = data.azurerm_client_config.deployer.tenant_id
+  object_id = var.deployer-uai.principal_id
+
+  secret_permissions = [
+    "get",
+  ]
+}
+
+// Create user KV with access policy
+resource "azurerm_key_vault" "kv_user" {
+  name                       = local.kv_user_name
+  location                   = local.region
+  resource_group_name        = local.rg_exists? data.azurerm_resource_group.resource-group[0].name : local.rg_name
+  tenant_id                  = data.azurerm_client_config.deployer.tenant_id
+  soft_delete_enabled        = true
+  soft_delete_retention_days = 7
+  purge_protection_enabled   = true
+
+  sku_name = "standard"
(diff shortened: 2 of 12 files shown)


deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
```diff
@@ -0,0 +1,149 @@
+/*
```
rukawata, line 1: This is good that you have a separate file for KV. 
Not in this PR, but let's do the same in other modules as well :)

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
```diff
@@ -0,0 +1,149 @@
...
+  name         = format("%s-iscsi-sshkey-pub", local.prefix)
+  value        = local.iscsi_public_key
+  key_vault_id = azurerm_key_vault.kv_user.id
+}
+
+/*
+ To force dependency between kv access policy and secrets. Expected behavior:
+ https://github.com/terraform-providers/terraform-provider-azurerm/issues/4971
+*/
+resource "azurerm_key_vault_secret" "iscsi_username" {
```
rukawata, line 98: Do we store username as well? This does not block this PR, but open for discussion :)

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
```diff
@@ -0,0 +1,149 @@
...
+  && local.iscsi_auth_type == "password" 
+  && try(local.var_iscsi.authentication.password, "") == "" ) ? 1 : 0
+  length           = 16
+  special          = true
+  override_special = "_%@"
+}
+
+// Using TF tls to generate SSH key pair for SID and store in user KV
+resource "tls_private_key" "sid" {
+  count     = try(file(var.sshkey.path_to_public_key), "") == "" ? 1 : 0
```
rukawata, line 127: should only generate key when` local.iscsi_auth_type == "key"`

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -124,6 +138,13 @@ locals {
   iscsi_auth_username = try(local.var_iscsi.authentication.username, "azureadm")
   iscsi_nic_ips       = local.sub_iscsi_exists ? try(local.var_iscsi.iscsi_nic_ips, []) : []
 
+  // By default, ssh key for iSCSI uses generated public key. Provide sshkey.path_to_public_key and path_to_private_key overides it
+  iscsi_public_key  = (local.iscsi_count > 0 && local.iscsi_auth_type == "key") ? try(file(var.sshkey.path_to_public_key), tls_private_key.iscsi[0].public_key_openssh) : null
+  iscsi_private_key = (local.iscsi_count > 0 && local.iscsi_auth_type == "key") ? try(file(var.sshkey.path_to_private_key), tls_private_key.iscsi[0].private_key_pem) : null
+
+  // Currently, only Linux VM is used as iSCSI target. The following password of Windows VM is a placeholder for potential future use.
+  iscsi_auth_password = (local.iscsi_count > 0 && local.iscsi_auth_type == "password") ? try(local.var_iscsi.authentication.password, random_password.iscsi_password[0].result) : ""
```
rukawata, line 146: can also use `null` if we do not need password, so it is consistent with key.

rukawata (commented):
Thanks for the PR!
Some comments that we can also discuss tomorrow :)
The pipeline did not pass the validate because provider not updated? although it is in the [code](https://github.com/Azure/sap-hana/blob/c23849ec3212e92aeddc27ca6a076854636709f3/.pipeline/templates/sapsystem/validate-sapsystem-steps.yml#L32) - let's check what went wrong.

And for the error, will require a rebuild of the release-pipeline form your branch, so we have this field in the outputs of deployer tfstate:

```
  on ../../../../U0419-EAUS-SAP-PRD/sap-hana/deploy/terraform/terraform-units/modules/sap_system/deployer/output.tf line 22, in output "deployer_user":
  22:   value = data.terraform_remote_state.deployer.outputs.deployer_user
    |----------------
    | data.terraform_remote_state.deployer.outputs is object with 6 attributes

```

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
YunziZhangAzure: This one is for SID(sapsystem). When provisioning landscape, the authentication of SID isn't known yet, thus provisioning it here, even if the SID will use username/password to logon.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
YunziZhangAzure: Will do. Will use null in all the similar case to keep them consistent

deploy/terraform/run/sap_landscape/saplandscape.json
```diff
@@ -21,6 +21,20 @@
                     "prefix": "10.1.0.0/24"
                 }
             }
+        },
+        "iscsi": {
+            "iscsi_count": 3,
```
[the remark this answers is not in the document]
YunziZhangAzure: Will remove redundant fields.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/keyvault.tf
rukawata: Ah I see. Because there will be only one key pair for the landscape and all SID. Now it makes sense.
But how do we handle the situation where only windows application server is generated in all SIDs under the same landscape?

YunziZhangAzure: In this case, there will still be a key pair stored in landscape's kv, because even if at some point, there are only windows application server deployed, there may be more Linux VMs using key pair as authentication type to be deployed based on this landscape.

rukawata (approved):
Approved!

Issue #775, opened 2020-09-10
KeyVault: SDU


mkdeegan:
**Introduction of KeyVault for the SDU**
Create a two-tier secure store of credentials, following a DevOps & AppOps model.

**Describe the solution you'd like**
DevOps Secure Store is used for secrets that are not appropriate to be human known, except through a break-glass type access. Access to these credentials should be reserved for the Tools, Terraform and Ansible.

AppOps Secure Store is used for passwords that are used during application installation, and can be retrieved by Administrators.

**Additional context**
Details provided in the KeyVault Specification Document

PR #784, opened 2020-09-16
Provision KVs for applications tier, store logon username/password in KV

YunziZhangAzure:
## Problem
#775

## Solution
1. Create two KVs for applications.

2. If authentication type = password, a password will be generated unless user specifies one.

## Tests

## Notes
Currently ansible_inventory in output_files module doesn't have password printed out. Will have another PR enable this feature to print the password to hosts.yml and hosts.

This pr also modifies the output of hdb module. The reason is because when trying to deploy only application tier, some of the output of hdb module is not empty list which causes the output_files module fail. Will do more tests about this point.

Retrieving the ssh key from saplandscape is not included in this pr.

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
@@ -28,7 +28,8 @@ resource "azurerm_linux_virtual_machine" "app" {
   ]
   size                            = local.app_sizing.compute.vm_size
   admin_username                  = local.sid_auth_username
-  disable_password_authentication = true
+  disable_password_authentication = ! local.enable_auth_password
+  admin_password                  = local.sid_auth_password
 
   os_disk {
     name                 = format("%s_%sapp%02dl%s-osdisk", local.prefix, lower(local.sid), count.index, substr(var.random-id.hex,0,3))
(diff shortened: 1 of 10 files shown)


deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
```diff
@@ -48,9 +49,12 @@ resource "azurerm_linux_virtual_machine" "app" {
...
   }
 
-  admin_ssh_key {
-    username   = local.authentication.username
-    public_key = file(var.sshkey.path_to_public_key)
+  dynamic "admin_ssh_key" {
+    for_each = range(local.enable_auth_password ? 0 : 1)
+    content {
+      username   = local.sid_auth_username
+      public_key = file(var.sshkey.path_to_public_key)
```
rukawata, line 56: Same question as hdb. This should use the keypair from sap_landscape.

rukawata (commented):
Overall looks good! Same question as the other.

deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
YunziZhangAzure: Will be included in a future pr

rukawata (approved):
lgtm!

PR #783, opened 2020-09-15
Provision KVs for hdb and support username/password logon to hdb VM

YunziZhangAzure:
## Problem
<Please describe what problem this PR is trying to resolve>
#775

## Solution
<Please elaborate the solution for the problem>
1. Create two KVs for hdb.

2. Although HANA database only supports Linux VM, using password as authentication type should be supported. 
If authentication type = password, a password will be generated unless user specifies one.

3. The password will be output to the hosts.yml and hosts for Ansible use

## Tests
<Please provide steps to test the PR>

## Notes
<Additional comments for the PR>

Retrieving the ssh key from saplandscape is not included in this pr.

PR #785, opened 2020-09-17
Add credentials of hana database to kv

YunziZhangAzure:
## Problem
Hana database credentials should be stored in KV

## Solution

1. Store passwords in KV.

2. If users don't specify some password field, a random generated password will be used.

3. To lower the complexity of codes, six random passwords will be generated regardless, but they will not be used if customers input the corresponding password.

## Tests

## Notes
Before reviewing this pr, please review pr #783 first.


deploy/terraform/terraform-units/modules/sap_system/hdb_node/key_vault.tf
```diff
@@ -104,3 +104,55 @@ resource "azurerm_key_vault_secret" "auth_password" {
   value        = local.sid_auth_password
   key_vault_id = azurerm_key_vault.kv_user.id
 }
+
+
+// Generate random passwords as hana database credentials
+resource "random_password" "credentials" {
```
rukawata, line 110: This should only generate when password is not provided.
If that is the case, actually it creates a layer of complexity - how many credential should I generate? And how do I know which one is for what :)

rukawata (commented):
Overall looks good! One comment to think about :)

deploy/terraform/terraform-units/modules/sap_system/hdb_node/key_vault.tf
YunziZhangAzure: My thought is to generate 6 passwords for hana database credentials regardless  how many passwords customers populate in credentials block. If some of them is empty, it will use one of these pre-generated passwords with a fixed index. 

This approach here could lead to some redundant resources, e.g: customers populate every password, but this approach can lower the complexity of codes. We can discuss about it.

rukawata: One approach what I used in deployer is:
- Always generate the same password for all deployers.
- And if there is at least one password is required, enable the password generation.

This does not apply to this situation here since we are talking about password used in different occasions whereas the other one only sudo user password. Let's start with what you have, can discuss and iterate later.

rukawata (approved):

PR #787, opened 2020-09-18
Add object_id for users to be consistent with sap_library

rukawata:
## Problem
No `object_id` on deployer since it uses msi.

## Solution
Add field `object_id` if user wants to add their user/spn back to the KV.
Otherwise, the KV policy for that user will only be used during bootstrap phase.

## Tests
1. Deploy deployer using bootstrap.
1. Deploy sap_library on deployer.
1. Re-init deployer.
1. Plan/apply deployer using ~/Azure_SAP_Automated_Deployment/sap-hana/deploy/terraform/run/sap_deployer/, see that user got removed:

  ```
# module.sap_deployer.azurerm_key_vault_access_policy.kv_user_portal[0] will be destroyed
  - resource "azurerm_key_vault_access_policy" "kv_user_portal" {
      - certificate_permissions = [] -> null
      - id                      = "/subscriptions/XXX/resourceGroups/GLOBAL-EAUS-MGMT-INFRASTRUCTURE/providers/Microsoft.KeyVault/vaults/GLOBAEAUSMGMTuser01D/objectId/XXX" -> null
      - key_permissions         = [] -> null
      - key_vault_id            = "/subscriptions/XXX/resourceGroups/GLOBAL-EAUS-MGMT-INFRASTRUCTURE/providers/Microsoft.KeyVault/vaults/GLOBAEAUSMGMTuser01D" -> null
      - object_id               = "XXX" -> null
      - secret_permissions      = [
          - "delete",
          - "get",
          - "list",
          - "set",
        ] -> null
      - storage_permissions     = [] -> null
      - tenant_id               = "XXX" -> null
    }
```
5. Add object_id in the JSON like below, apply, see the right user added again.
```
  "deployers":[{
    "users": {
      "object_id": ["<object_id>"]
    }
  }],
```

## Notes
Also sync the json under bootstrap and run.
Will have a separate PR to remove the sshkey section since by default we generate key pair.

### DIFF
--- a/.pipeline/templates/deployer/create-deployer-steps.yml
+++ b/.pipeline/templates/deployer/create-deployer-steps.yml
@@ -32,6 +32,12 @@ steps:
         nsg: {
           allowed_ips: ["67.160.0.0/16", $allowed_ip]
         }
+      }' \
+      | jq '. += {
+        sshkey: {
+          path_to_public_key: "sshkey.pub",
+          path_to_private_key: "sshkey"
+        }
       }' > ${input}
 
       cat ${input}
--- a/deploy/terraform/bootstrap/sap_deployer/deployer.json
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer.json
@@ -12,10 +12,6 @@
       }
     }
   },
-  "sshkey": {
-    "path_to_public_key": "sshkey.pub",
-    "path_to_private_key": "sshkey"
-  },
   "options": {
     "enable_secure_transfer": true
   }
--- a/deploy/terraform/run/sap_deployer/deployer.json
+++ b/deploy/terraform/run/sap_deployer/deployer.json
@@ -12,10 +12,6 @@
       }
     }
   },
-  "sshkey": {
-    "path_to_public_key": "sshkey.pub",
-    "path_to_private_key": "sshkey"
-  },
   "options": {
     "enable_secure_transfer": true
   }
--- a/deploy/terraform/run/sap_deployer/deployer.json
+++ b/deploy/terraform/run/sap_deployer/deployer.json
@@ -4,6 +4,7 @@
     "environment": "global",
     "vnets": {
       "management": {
+        "name" : "MGMT",
         "address_space": "10.0.0.0/26",
         "subnet_mgmt": {
           "prefix": "10.0.0.16/28"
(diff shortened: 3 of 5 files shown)


YunziZhangAzure (commented):
I'm thinking if it is better to keep the logic of bootstrap/sap_deployer same as run/sap_deployer, which both need user's object_id as input. If not specified, no users will have access to the kv.
In this way, there will be no resources changes between bootstrap/sap_deployer and run/sap_deployer.
Any reasons to retrieve the user object_id (data.azurerm_client_config.deployer.object_id) in bootstrap/sap_deployer?

rukawata:
> I'm thinking if it is better to keep the logic of bootstrap/sap_deployer same as run/sap_deployer, which both need user's object_id as input. If not specified, no users will have access to the kv.
> In this way, there will be no resources changes between bootstrap/sap_deployer and run/sap_deployer.
> Any reasons to retrieve the user object_id (data.azurerm_client_config.deployer.object_id) in bootstrap/sap_deployer?

Yes, to authorize the SPN who deployed the KV to access the sshkey/password, so that the post_deployment.sh would work (to upload json and tfstate to deployer).
Another option to keep it consistent behavior would be make object_id mandatory - will require manually work in preparing the JSON.
But one could also argue that after deployer is created, by default only msi should have access to the KV, so the behavior in current PR would guarantee that.

YunziZhangAzure:
> > I'm thinking if it is better to keep the logic of bootstrap/sap_deployer same as run/sap_deployer, which both need user's object_id as input. If not specified, no users will have access to the kv.
> > In this way, there will be no resources changes between bootstrap/sap_deployer and run/sap_deployer.
> > Any reasons to retrieve the user object_id (data.azurerm_client_config.deployer.object_id) in bootstrap/sap_deployer?
> 
> Yes, it is necessary to authorize the SPN, who deployed the KV, to have access to the sshkey/password. Only this way that the post_deployment.sh would work (to upload json and tfstate to deployer).
> 
> I see another two options option if we want keep it consistent behavior:
> 
> 1. Make object_id mandatory - will require manually work in preparing the JSON.
> 2. Add manual step to add SPN to access KV, before running post_deployment.
> 
> But one could also argue that after deployer is created, by default only msi should have access to the KV, so the behavior in current PR would guarantee that, that is why I'm in favor of the current implementation.

Ok! Let's keep the current version and collect more feedbacks at integration tests.

YunziZhangAzure (dismissed):

YunziZhangAzure (approved):

PR #792, opened 2020-09-22
Store SPN of Azure fence agent in KV

YunziZhangAzure:
## Problem

The SPN of Azure fence agent needs to be stored in SID's KV.

## Solution
When Azure fence agent is specified in input json, the value will be stored in user's KV.

## Tests

## Notes


deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -96,6 +96,13 @@ locals {
   sid_auth_username    = try(local.hdb.authentication.username, "azureadm")
   sid_auth_password    = local.enable_auth_password ? try(local.hdb.authentication.password, random_password.password[0].result) : null
 
+  // SPN of Azure Fence Agent for Hana Database
+  enable_fence_agent          = local.enable_deployment && try(local.hdb.fence_agent, null) != null
+  fence_agent_subscription_id = local.enable_fence_agent ? local.hdb.fence_agent.subscription_id : null
+  fence_agent_tenant_id       = local.enable_fence_agent ? local.hdb.fence_agent.tenant_id : null
```
rukawata, line 102: I think the sub as well as tenant that is used to deploy the resources should be the same as the ones used in fencing agent. If that is the case, can we directly pull those info from terraform config instead of entering via input JSON?

deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf
```diff
@@ -12,9 +12,9 @@ HANA DB Linux Server private IP range: .10 -
 
 # Creates the admin traffic NIC and private IP address for database nodes
 resource "azurerm_network_interface" "nics-dbnodes-admin" {
-  count                         = local.enable_deployment ? length(local.hdb_vms) : 0
-  name                          = local.customer_provided_names ? format("%s-admin-nic",local.hdb_vms[count.index].name) : format("%s_%s-adminnic", local.prefix,local.hdb_vms[count.index].name)
-  
+  count = local.enable_deployment ? length(local.hdb_vms) : 0
```
rukawata, line 15: nit: format

rukawata (commented):
Overall logic looks correct.
The change in pipeline to add fence agent info is not included in this PR (it would probably still run?), will there be a separate one?

deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf
YunziZhangAzure: Will format here.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
YunziZhangAzure: It seems there are no requirements about which subscription and tenant the Azure fence agent should belong to?

In V2, creating Azure fence agent is a separate step by using script(az cli). I think technically hdb resources and Azure fence agent can be in different subscription? so I didn't want to add restrictions here.

YunziZhangAzure:
> Overall logic looks correct.
> The change in pipeline to add fence agent info is not included in this PR (it would probably still run?), will there be a separate one?

I'm not sure if we want to add this part into our test pipeline. Technically we can, but then we may need more test cases in test pipeline, e.g: set high_availability = true and deploy more hdb nodes, which may take longer to complete a test case in test pipeline.

Please let me know your opinions.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf
YunziZhangAzure: Hmm.. looks like this is a formatted version already by using terraform fmt

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
rukawata: So the question is, can an SPN belongs to sub A manage resources in sub B? My guess would be no. But let's check.

rukawata:
> > Overall logic looks correct.
> > The change in pipeline to add fence agent info is not included in this PR (it would probably still run?), will there be a separate one?
> 
> I'm not sure if we want to add this part into our test pipeline. Technically we can, but then we may need more test cases in test pipeline, e.g: set high_availability = true and deploy more hdb nodes, which may take longer to complete a test case in test pipeline.
> 
> Please let me know your opinions.

We do deploy a HA scenario for ansible :)

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
YunziZhangAzure: Here's the test I did:
I created a service principal in Subscription A;
Used our fencing_agent_role.json as a template, set the AssignableScopes = subscription B.
Assigned this role to the service principal. 
There's no error.
Here's the result:(http://sp-test-0924 is in our AzLinux SAP Landscapes, "90f511b-ada5-4d0c-8fab-73edd0ed75e2" is my own DevYunzzha subscription)

azureadm@devbox:~/Azure_SAP_Automated_Deployment/sap-hana/util$ az role assignment list --assignee "http://sp-test-0924"
[
  {
    "canDelegate": null,
    "id": "/subscriptions/790f511b-ada5-4d0c-8fab-73edd0ed75e2/providers/Microsoft.Authorization/roleAssignments/2f2853af-d382-4e29-8239-1213b6549490",
    "name": "2f2853af-d382-4e29-8239-1213b6549490",
    "principalId": "9278894c-ecc4-4e9f-958c-b0fe2c727c91",
    "principalName": "http://sp-test-0924",
    "principalType": "ServicePrincipal",
    "roleDefinitionId": "/subscriptions/790f511b-ada5-4d0c-8fab-73edd0ed75e2/providers/Microsoft.Authorization/roleDefinitions/d648e11c-1dff-485d-9c0c-a494e9dcc344",
    "roleDefinitionName": "SAP HANA Linux Fencing Agent 0924",
    "scope": "/subscriptions/790f511b-ada5-4d0c-8fab-73edd0ed75e2",
    "type": "Microsoft.Authorization/roleAssignments"
  }
]


And I find some description in Azure official doc. From the words, it seems service principal is limited to the tenant level, not he subscription level.
https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals

image

YunziZhangAzure:
> > > Overall logic looks correct.
> > > The change in pipeline to add fence agent info is not included in this PR (it would probably still run?), will there be a separate one?
> > 
> > 
> > I'm not sure if we want to add this part into our test pipeline. Technically we can, but then we may need more test cases in test pipeline, e.g: set high_availability = true and deploy more hdb nodes, which may take longer to complete a test case in test pipeline.
> > Please let me know your opinions.
> 
> We do deploy a HA scenario for ansible test pipeline :)

OK! Will add a test case for this feature.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
rukawata: I see. Potentially we could pop the tenant id, but for now just keep all from inputs :)

rukawata (approved):

PR #796, opened 2020-09-24
[hotfix]  Use empty string instead of null 

YunziZhangAzure:
## Problem
Using 'null' will lead to output_file  module behave incorrectly

## Solution
Change it to empty string ""

## Tests
See the test pipeline below

## Notes


rukawata (approved):
Thanks for fixing it!

PR #800, opened 2020-09-24
Fix several issues in KV

YunziZhangAzure:
## Problem

## Solution
1. Add users.object_id into template sap_deployer.json, so that users can know how to add the users' object_id.

2. Fix the output of sap_deployer, so that it includes all the users' object_id.

3. Fix the resource group of common_infrastructure, so that it won't cause dependency issue.

4. When provisioning SID, ssh key pair will be retrieved from sap landscape, if authentication type is "key", so I remove ssh block in sapsystem.json which is not used at this step.

## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/deployer.json
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer.json
@@ -12,13 +12,6 @@
       }
     }
   },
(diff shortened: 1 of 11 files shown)


rukawata (approved):
thanks for fixing the bugs!

PR #806, opened 2020-09-28
Produce only one pair of KVs per SID

rukawata:
## Problem
1. Application and hdb generates a pair of KVs for both, hence we get 4 KVs from sap_system.
2. When deploy sap_system, sap_landscape KV pairs got deployed again.
3. SID in key names are not replaced with real SID.

## Solution
1. Create only one pair of KVs for one SID.
2. Do not deploy sap_landscape KV when a sap_landscape KV is imported.
3. Use SID in key naming.

## Tests
Deploy sap_system with existing sap_landscape KV, see only two KVs created:
image

## Notes


deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -181,12 +175,12 @@ locals {
...
   hdb_nr                 = try(local.hdb_ins.instance_number, "01")
   hdb_cred               = try(local.hdb.credentials, {})
-  db_systemdb_password   = local.enable_deployment ? try(local.hdb_cred.db_systemdb_password, random_password.credentials[0].result) : null
-  os_sidadm_password     = local.enable_deployment ? try(local.hdb_cred.os_sidadm_password, random_password.credentials[1].result) : null
-  os_sapadm_password     = local.enable_deployment ? try(local.hdb_cred.os_sapadm_password, random_password.credentials[2].result) : null
-  xsa_admin_password     = local.enable_deployment ? try(local.hdb_cred.xsa_admin_password, random_password.credentials[3].result) : null
-  cockpit_admin_password = local.enable_deployment ? try(local.hdb_cred.cockpit_admin_password, random_password.credentials[4].result) : null
-  ha_cluster_password    = local.enable_deployment && local.hdb_ha ? try(local.hdb_cred.ha_cluster_password, random_password.credentials[5].result) : null
+  db_systemdb_password   = local.enable_deployment ? var.hdb_cred[0].result : null
```
YunziZhangAzure, line 178: These six passwords will be included in hana-database-info and passed to output_file module to generate output.json.
If customers specify the passwords in input json, the KV will store the input passwords, but the var.hdb_cred here still contains random passwords which are not same as the ones stored in KV, which will lead to the inconsistency between output.json and the real secret values in KV.

so I think maybe var.hdb_cred should be the real combined passwords instead of the random.credential.

 Another thing I notice is these six passwords are not used anywhere in module hdb_node, except being passed to local.hana-database.credentials. Maybe we don't need to list them here, just assign the variable var.hdb_cred
to the local.hana-database.credentials directly.
https://github.com/Azure/sap-hana/blob/5d191c890983cd1f06dfed780817369ad1f4d440/deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf#L223


Or I'm thinking we can restructure the Keyvault of sapsystem in this way:
The logic about creating KVs happen in common_infrastructure module and KVs are passed to the app tier/hdb module.
The logic about generating secrets and storing secrets happen in app tier/hdb module

rukawata: Customer input passwords are incorporated [here](https://github.com/Azure/sap-hana/blob/5d191c890983cd1f06dfed780817369ad1f4d440/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf#L151-L156)

rukawata: With current scope, meaning no password retrieval with ansible, we still need the credentials block to be passed to output module, so will show up in the output.json. Hence credentials can't be removed :)

rukawata: Change the logic to pass sap_system KV instead of secrets.

YunziZhangAzure (approved):
Thanks for the pr! LGTM

PR #807, opened 2020-09-29
Support anydb with KV

rukawata:
## Problem
Currently anydb module does not store secrets in KV.

## Solution
Add supportability to store anydb credentials in KV:
1. Use SSH key pair from sap_landscape.
1. Store VM password if it is used and defined.
1. Generate and store anydb credential in SID KV, look like below:
image
And if password is used, looks like this:
image

## Tests
https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=11204&view=results

## Notes


YunziZhangAzure (approved):
LGTM!

YunziZhangAzure (approved):

PR #832, opened 2020-10-09
Correct saplibrary variables when conditional expression is used

YunziZhangAzure:
## Problem
1. There's error when using existing storage accounts.

```
{
    "infrastructure": {
        "region": "eastus",
        "environment": "np",
        "resource_group":{
            "is_existing": "true",
            "arm_id": "/subscriptions/xx/resourceGroups/UNIT-EAUS-SAP_LIBRARY"
        }
    },
"storage_account_sapbits": {
            "is_existing": true,
                "arm_id": "/subscriptions/xx/resourceGroups/NP-EAUS-SAP_LIBRARY/providers/Microsoft.Storage/storageAccounts/npeaussaplib2b33",
                    "file_share": { "is_existing": true },
                        "sapbits_blob_container": {
                                      "is_existing": true
                                          }
                                            },
                                              "storage_account_tfstate": {
                                                          "is_existing": true,
                                                              "arm_id": "/subscriptions/xx/resourceGroups/NP-EAUS-SAP_LIBRARY/providers/Microsoft.Storage/storageAccounts/npeaustfstate2b33",
                                                                  "saplibrary_blob_container": {
                                                                                "is_existing": true
                                                                                    }
                                                                                      }
}

```

image

## Solution
1. Root cause:
A data source is a special type of resource, which is not necessarily same as the corresponding resource type.

https://github.com/hashicorp/terraform/issues/22450

"While it is true that Terraform only evaluates the value on the chosen side of the conditional, Terraform does still require both sides to be convertible to some common type so that Terraform can know what the result type will be even if it doesn't yet know which of the expressions will be selected. "

Thus, in this pr, the type in conditional expression is updated to be primitive type -- String, instead of object.

## Tests

## Notes

"saplibrary_blob_container" in input json is changed to "tfstate_blob_container" for consistent reason

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
@@ -52,11 +52,11 @@ locals {
   var_rg    = try(local.var_infra.resource_group, {})
   rg_exists = try(local.var_rg.is_existing, false)
   rg_arm_id = local.rg_exists ? try(local.var_rg.arm_id, "") : ""
-  rg_name   = local.rg_exists ? "" : try(local.var_rg.name, format("%s-SAP_LIBRARY", local.prefix))
+  rg_name   = local.rg_exists ? split("/", local.rg_arm_id)[4] : try(local.var_rg.name, format("%s-SAP_LIBRARY", local.prefix))
 
   // Storage account for sapbits
   sa_sapbits_exists                   = try(var.storage_account_sapbits.is_existing, false)
-  sa_sapbits_name                     = lower(format("%s%ssaplib%s", substr(local.environment, 0, 5), local.location_short, substr(random_id.post_fix.hex, 0, 4)))
+  sa_sapbits_name                     = local.sa_sapbits_exists ? split("/", local.sa_sapbits_arm_id)[8] : lower(format("%s%ssaplib%s", substr(local.environment, 0, 5), local.location_short, substr(random_id.post_fix.hex, 0, 4)))
   sa_sapbits_account_tier             = local.sa_sapbits_exists ? "" : try(var.storage_account_sapbits.account_tier, "Standard")
   sa_sapbits_account_replication_type = local.sa_sapbits_exists ? "" : try(var.storage_account_sapbits.account_replication_type, "LRS")
   sa_sapbits_account_kind             = local.sa_sapbits_exists ? "" : try(var.storage_account_sapbits.account_kind, "StorageV2")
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
```diff
@@ -91,10 +91,11 @@ locals {
 
 // Output objects 
 locals {
-  rg_library               = local.rg_exists ? data.azurerm_resource_group.library[0] : azurerm_resource_group.library[0]
-  sa_tfstate               = local.sa_tfstate_exists ? data.azurerm_storage_account.storage_tfstate[0] : azurerm_storage_account.storage_tfstate[0]
-  sa_sapbits               = local.sa_sapbits_exists ? data.azurerm_storage_account.storage_sapbits[0] : azurerm_storage_account.storage_sapbits[0]
-  storagecontainer_tfstate = local.sa_tfstate_container_exists ? data.azurerm_storage_container.storagecontainer_tfstate[0] : azurerm_storage_container.storagecontainer_tfstate[0]
-  storagecontainer_sapbits = ! local.sa_sapbits_blob_container_enable ? null : (local.sa_sapbits_blob_container_exists ? data.azurerm_storage_container.storagecontainer_sapbits[0] : azurerm_storage_container.storagecontainer_sapbits[0])
+  rg_library_name          = local.rg_exists ? data.azurerm_resource_group.library[0].name : azurerm_resource_group.library[0].name
```
rukawata, line 94: Can we use rg_name instead? If existing rg is used, it pulls the rg name from the arm_id.

deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
```diff
@@ -91,10 +91,11 @@ locals {
...
+  rg_library_location      = local.rg_exists ? data.azurerm_resource_group.library[0].location : azurerm_resource_group.library[0].location
+  sa_tfstate               = local.sa_tfstate_exists ? data.azurerm_storage_account.storage_tfstate[0].name : azurerm_storage_account.storage_tfstate[0].name
+  sa_sapbits               = local.sa_sapbits_exists ? data.azurerm_storage_account.storage_sapbits[0].name : azurerm_storage_account.storage_sapbits[0].name
+  storagecontainer_tfstate = local.sa_tfstate_container_exists ? data.azurerm_storage_container.storagecontainer_tfstate[0].name : azurerm_storage_container.storagecontainer_tfstate[0].name
+  storagecontainer_sapbits = ! local.sa_sapbits_blob_container_enable ? null : (local.sa_sapbits_blob_container_exists ? data.azurerm_storage_container.storagecontainer_sapbits[0].name : azurerm_storage_container.storagecontainer_sapbits[0].name)
```
rukawata, line 99: Can we rename them to `sa_tfstate_name`, `sa_sapbits_name`,... so it's more descriptive?

rukawata (commented):
Thanks for fixing this! Some small questions :)

rukawata (approved):
lgtm!

PR #834, opened 2020-10-10
correct variable type when using conditional expression 

YunziZhangAzure:
## Problem
See https://github.com/Azure/sap-hana/pull/832
Merge this hotfix commit to feature/keyvault as well.

## Solution

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/output.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/output.tf
@@ -1,13 +1,13 @@
 output "tfstate_storage_account" {
-  value = local.sa_tfstate
+  value = local.sa_tfstate_name
 }
 
 output "sapbits_storage_account" {
-  value = local.sa_sapbits
+  value = local.sa_sapbits_name
 }
 
 output "storagecontainer_tfstate" {
-  value = local.storagecontainer_tfstate
+  value = local.sa_tfstate_container_name
 }
 
 output "storagecontainer_sapbits" {
--- a/deploy/terraform/terraform-units/modules/sap_library/storage_accounts.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/storage_accounts.tf
@@ -14,8 +14,8 @@ data "azurerm_storage_account" "storage_tfstate" {
 resource "azurerm_storage_account" "storage_tfstate" {
   count                     = local.sa_tfstate_exists ? 0 : 1
   name                      = local.sa_tfstate_name
-  resource_group_name       = local.rg_library.name
-  location                  = local.rg_library.location
+  resource_group_name       = local.rg_name
+  location                  = local.rg_library_location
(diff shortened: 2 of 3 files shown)


rukawata (commented):
Sorry I did not catch this earlier in PR #832.
By not using the object, the container creation lost the dependency to the SA creation.
Therefore we see error in test [here](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12071&view=results). 
Same problem with feature/remote-tfstate branch, see test result [here](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12082&view=logs&j=d764d31b-fb26-5fc3-93b2-7fc094a2e94a&t=8b66a7ac-67d7-5b53-2853-18ed51fe316e).

I feel to force the dependency using depends_on is probably not a good idea - if possible TF should handle the dependency.
There are some creative ideas in the thread [here](https://github.com/hashicorp/terraform/issues/22405), maybe take a look. Or another option is to do the logic in the resource block instead of having a variable.

Another thing I don't have an answer yet is, when we do not reuse existing storage account, this should be the same condition mismatch between the two object, why it was not a problem?

YunziZhangAzure:
> Sorry I did not catch this earlier in PR #832.
> By not using the object, the container creation lost the dependency to the SA creation.
> Therefore we see error in test [here](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12071&view=results).
> Same problem with feature/remote-tfstate branch, see test result [here](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12082&view=logs&j=d764d31b-fb26-5fc3-93b2-7fc094a2e94a&t=8b66a7ac-67d7-5b53-2853-18ed51fe316e).
> 
> I feel to force the dependency using depends_on is probably not a good idea - if possible TF should handle the dependency.
> There are some creative ideas in the thread [here](https://github.com/hashicorp/terraform/issues/22405), maybe take a look. Or another option is to do the logic in the resource block instead of having a variable.
> 
> Another thing I don't have an answer yet is, when we do not reuse existing storage account, this should be the same condition mismatch between the two object, why it was not a problem?

Thanks for catching this issue! I have checked the post, but it seems the scenario of them (map, list, variable) is a little different from us. One option may be applicable is to transfer the object jsoncode, which may make our codes more complex. 
so I'm including the logic in the resource block to introduce an implicit dependency between storage account and container.
Here's the test after a new commit. 
https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12193&view=results

rukawata (approved):
lgtm! Thanks for fixing this!

rukawata:
> lgtm! Thanks for fixing this!

Ah, please also fix the same in feature/remote-tfstate :)

PR #838, opened 2020-10-15
Fix release pipeline

rukawata:
## Problem
Release pipeline has problem after sap_library deployment.

## Solution
There are several problem fixed in this PR which fixed the release pipeline:
     - add sap_land info into release pipeline, otherwise ansible test does not know where to pull the SSH keypair for SID VMs.
     - fix output change from pr #834
     - change default ssh key path for sap_system - it should not point to local files. 
     - download ssh key pair to deployer in ansible test
 
## Tests
https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12321&view=results

## Notes
Why we need the key pair?

For the first attempt to ssh from deployer to a target VM, it requires to have the public key so this can be added to the `authorized_keys` of target VM. Otherwise, the first attempt to connect the target will fail although private key is given. Then it will ask for password as second logon method.

After the first connection, the public key is no longer required to establish an ssh connection with the target VM.

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/output.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/output.tf
@@ -2,16 +2,24 @@ output "tfstate_storage_account" {
   value = local.sa_tfstate_name
 }
 
-output "sapbits_storage_account" {
+output "sapbits_storage_account_name" {
   value = local.sa_sapbits_name
+}
+
+output "sapbits_sa_resource_group_name" {
+  value = local.rg_name
 }
 
 output "storagecontainer_tfstate" {
   value = local.sa_tfstate_container_name
 }
 
-output "storagecontainer_sapbits" {
-  value = local.storagecontainer_sapbits
+output "storagecontainer_sapbits_name" {
+  value = local.storagecontainer_sapbits_name
+}
+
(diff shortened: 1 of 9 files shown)


deploy/terraform/terraform-units/modules/sap_library/output.tf
```diff
@@ -2,16 +2,24 @@ output "tfstate_storage_account" {
...
   value = local.sa_tfstate_container_name
 }
 
-output "storagecontainer_sapbits" {
-  value = local.storagecontainer_sapbits
+output "storagecontainer_sapbits_name" {
+  value = local.storagecontainer_sapbits_name
+}
+
+output "storagecontainer_sapbits_sa_name" {
```
YunziZhangAzure, line 21: Does "storagecontainer_sapbits_sa_name" have same value as "sapbits_storage_account_name"?

rukawata: You are right! Changed and tested： https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12356&view=results

YunziZhangAzure (approved):

PR #842, opened 2020-10-17
Read scs_count from json file

KimForss:
## Problem
The customers may need more than two servers in the scs tier

## Solution
Add a "scs_server_count" property to the application section of the json file

## Tests
Change the value in the json file to "scs_server_count": 3 deploy and validate that 3 scs servers are created

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-scs.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-scs.tf
@@ -81,7 +81,7 @@ resource "azurerm_linux_virtual_machine" "scs" {
 
 # Create the SCS Windows VM(s)
 resource "azurerm_windows_virtual_machine" "scs" {
-  count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? (local.scs_high_availability ? 2 : 1) : 0) : 0
+  count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? local.scs_server_count : 0) : 0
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-scs.tf
```diff
@@ -81,7 +81,7 @@ resource "azurerm_linux_virtual_machine" "scs" {
 
 # Create the SCS Windows VM(s)
 resource "azurerm_windows_virtual_machine" "scs" {
-  count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? (local.scs_high_availability ? 2 : 1) : 0) : 0
+  count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? local.scs_server_count : 0) : 0
```
rukawata, line 84: Check this :)
`local.enable_deployment && upper(local.app_ostype) == "WINDOWS" ? local.scs_server_count : 0`

And I think we may be missing the same logic for linux scs VMs.

rukawata (commented):
Thanks for adding this!

deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-scs.tf
KimForss: you are right - my bad

rukawata (commented):
Thanks for addressing the comment!
I missed another place, the `scs-data-disks` in `deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf` needs to be adjusted as well.

rukawata (approved):
LGTM!

PR #844, opened 2020-10-17
Kimforss add anchor v ms

KimForss:
## Problem
For HA zonal deployments all VMs may be in availability sets attached to proximity placement groups. For this scenario a zonal VM is needed to anchor the PPG to the right availability zone

## Solution
Add a section 
   "anchor_vms": {
      "sku": "Standard_D4s_v3",
      "authentication": {
        "type": "key",
        "username": "azureadm"
      },
      "support_ultra": [true, true, false],
      "os": {
        "publisher": "Oracle",
        "offer": "Oracle-Linux",
        "sku": "81",
        "version": "latest"
      }
    },
to the infrastructure node in the input json. This will instruct the automation code to deploy the anchor VMs
## Tests
Add the section mentioned above to the json file and deploy

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
@@ -52,7 +52,8 @@ locals {
   storageaccount_name         = var.naming.storageaccount_names.SDU
   keyvault_names              = var.naming.keyvault_names.SDU
   virtualmachine_names        = var.naming.virtualmachine_names.ISCSI_COMPUTERNAME
-  anchor_virtualmachine_names = var.naming.virtualmachine_names.ANCHOR_COMPUTERNAME
+  anchor_virtualmachine_names = var.naming.virtualmachine_names.ANCHOR_VMNAME
+  anchor_computer_names       = var.naming.virtualmachine_names.ANCHOR_COMPUTERNAME
   resource_suffixes           = var.naming.resource_suffixes
 
   //Filter the list of databases to only HANA platform entries
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
@@ -48,11 +48,12 @@ locals {
   zones            = distinct(concat(local.db_zones, local.app_zones, local.scs_zones, local.web_zones))
   zonal_deployment = length(local.zones) > 0 ? true : false
 
-  vnet_prefix          = var.naming.prefix.VNET
-  storageaccount_name  = var.naming.storageaccount_names.SDU
-  keyvault_names       = var.naming.keyvault_names.SDU
-  virtualmachine_names = var.naming.virtualmachine_names.ISCSI_COMPUTERNAME
-  resource_suffixes    = var.naming.resource_suffixes
+  vnet_prefix                 = var.naming.prefix.VNET
+  storageaccount_name         = var.naming.storageaccount_names.SDU
+  keyvault_names              = var.naming.keyvault_names.SDU
+  virtualmachine_names        = var.naming.virtualmachine_names.ISCSI_COMPUTERNAME
+  anchor_virtualmachine_names = var.naming.virtualmachine_names.ANCHOR_COMPUTERNAME
+  resource_suffixes           = var.naming.resource_suffixes
 
   //Filter the list of databases to only HANA platform entries
   hana-databases = [
@@ -72,6 +73,27 @@ locals {
   }
 
   var_infra = try(var.infrastructure, {})
+
+  //Anchor VM
+  anchor      = try(local.var_infra.anchor_vms, {})
+  anchor_size = try(local.anchor.sku, "Standard_D8s_v3")
+  anchor_authentication = try(local.anchor.authentication,
+    {
+      "type"     = "key"
+      "username" = "azureadm"
+  })
+
+  anchor_custom_image = try(local.anchor.os.source_image_id, "") != "" ? true : false
+
+  anchor_os = {
+    "source_image_id" = local.anchor_custom_image ? local.anchor.os.source_image_id : ""
+    "publisher"       = try(local.anchor.os.publisher, local.anchor_custom_image ? "" : "suse")
+    "offer"           = try(local.anchor.os.offer, local.anchor_custom_image ? "" : "sles-sap-12-sp5")
+    "sku"             = try(local.anchor.os.sku, local.anchor_custom_image ? "" : "gen1")
+    "version"         = try(local.anchor.os.version, local.anchor_custom_image ? "" : "latest")
+  }
+  anchor_ostype           = upper(try(local.anchor.os.os_type, "LINUX"))
+  anchor_enable_ultradisk = try(local.anchor.support_ultra, [false, false, false])
 
   //Resource group
   var_rg    = try(local.var_infra.resource_group, {})
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
@@ -39,7 +39,7 @@ resource "azurerm_linux_virtual_machine" "anchor" {
   disable_password_authentication = true
 
   os_disk {
-    name                 = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.os_disk)
+    name                 = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.osdisk)
     caching              = "ReadWrite"
     storage_account_type = "Standard_LRS"
   }
@@ -89,7 +89,7 @@ resource "azurerm_windows_virtual_machine" "anchor" {
   admin_password = local.anchor_authentication.password
 
   os_disk {
-    name                 = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.os_disk)
+    name                 = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.osdisk)
     caching              = "ReadWrite"
     storage_account_type = "Standard_LRS"
   }
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
@@ -1,0 +1,117 @@
+data "azurerm_subnet" "anchor" {
+  count                = local.zonal_deployment ? (local.sub_db_exists ? 1 : 0) : 0
+  name                 = split("/", local.sub_db_arm_id)[10]
+  resource_group_name  = split("/", local.sub_db_arm_id)[4]
+  virtual_network_name = split("/", local.sub_db_arm_id)[8]
+}
+
+# Create Anchor VM
+resource "azurerm_network_interface" "anchor" {
+  count                         = local.zonal_deployment ? length(local.zones) : 0
+  name                          = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.nic)
+  resource_group_name           = local.rg_exists ? data.azurerm_resource_group.resource-group[0].name : azurerm_resource_group.resource-group[0].name
+  location                      = local.rg_exists ? data.azurerm_resource_group.resource-group[0].location : azurerm_resource_group.resource-group[0].location
+  enable_accelerated_networking = false
+
+  ip_configuration {
+    name                          = "IPConfig1"
+    subnet_id                     = data.azurerm_subnet.anchor[0].id
+    private_ip_address            = try(local.anchor.nic_ips[count.index], cidrhost(data.azurerm_subnet.anchor[0].address_prefixes[0], (count.index + 5)))
+    private_ip_address_allocation = "static"
+  }
+}
+
+# Create the Linux Application VM(s)
+resource "azurerm_linux_virtual_machine" "anchor" {
+  count                        = local.zonal_deployment && (local.anchor_ostype == "LINUX") ? length(local.zones) : 0
+  name                         = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.vm)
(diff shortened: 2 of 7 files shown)


deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
```diff
@@ -0,0 +1,117 @@
+data "azurerm_subnet" "anchor" {
+  count                = local.zonal_deployment ? (local.sub_admin_exists ? 1 : 0) : 0
+  name                 = split("/", local.sub_admin_arm_id)[10]
+  resource_group_name  = split("/", local.sub_admin_arm_id)[4]
+  virtual_network_name = split("/", local.sub_admin_arm_id)[8]
+}
+
+# Create Anchor VM
+resource "azurerm_network_interface" "anchor" {
+  count                         = local.zonal_deployment ? length(local.zones) : 0
```
rukawata, line 10: Should we add logic so we only deploy anchor VM when DB VMs uses avset?
Meaning `local.zonal_deployment && local.db_server_count == local.db_zone_count`

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
```diff
@@ -0,0 +1,117 @@
...
+    name                          = "IPConfig1"
+    subnet_id                     = data.azurerm_subnet.anchor[0].id
+    private_ip_address            = try(local.anchor.nic_ips[count.index], cidrhost(data.azurerm_subnet.anchor[0].address_prefixes[0], (count.index + 5)))
+    private_ip_address_allocation = "static"
+  }
+}
+
+# Create the Linux Application VM(s)
+resource "azurerm_linux_virtual_machine" "anchor" {
+  count                        = local.zonal_deployment && (local.anchor_ostype == "LINUX") ? length(local.zones) : 0
```
rukawata, line 26: Same concern as above.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -73,6 +75,27 @@ locals {
...
+
+  anchor_os = {
+    "source_image_id" = local.anchor_custom_image ? local.anchor.os.source_image_id : ""
+    "publisher"       = try(local.anchor.os.publisher, local.anchor_custom_image ? "" : "suse")
+    "offer"           = try(local.anchor.os.offer, local.anchor_custom_image ? "" : "sles-sap-12-sp5")
+    "sku"             = try(local.anchor.os.sku, local.anchor_custom_image ? "" : "gen1")
+    "version"         = try(local.anchor.os.version, local.anchor_custom_image ? "" : "latest")
+  }
+  anchor_ostype           = upper(try(local.anchor.os.os_type, "LINUX"))
+  anchor_enable_ultradisk = try(local.anchor.support_ultra, [false, false, false])
```
rukawata, line 97: Two questions:
1. Why is this a list of 3?
2. Why ultradisk needs to be enabled? The anchor VM only has os disk. What is the impact of this?

rukawata (commented):
Thanks for helping me to understand better the relationship with avset, zonal, ppg and anchor VM :) Based on our conversation, I still have some question and comments and I put them here.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
```diff
@@ -0,0 +1,117 @@
+data "azurerm_subnet" "anchor" {
+  count                = local.zonal_deployment ? (local.sub_admin_exists ? 1 : 0) : 0
```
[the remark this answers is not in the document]
KimForss: I would not add another subnet for this - customers don't want to have many subnets unless they have to - so I would use the admin one

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
KimForss: I would keep it optional and driven by the anchor_vms section in json

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
KimForss: the ultra list is for the 3 zones - the ultra flag is to make sure that we don't deploy in a DC that does not have Ultra capability

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -73,6 +75,27 @@ locals {
 
   var_infra = try(var.infrastructure, {})
 
+  //Anchor VM
+  anchor      = try(local.var_infra.anchor_vms, {})
+  anchor_size = try(local.anchor.sku, "Standard_D8s_v3")
```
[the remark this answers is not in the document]
KimForss: to read the sku for the db vm means that we need to read the disk json from common infrastructure module as well as query the json so that logic is quite complex. I believe customers are ok with providing the SKU they want for the anchor

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -73,6 +75,27 @@ locals {
 
   var_infra = try(var.infrastructure, {})
 
+  //Anchor VM
```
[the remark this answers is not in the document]
KimForss: anchor is for db tier - but it is deployed from common_module - if we want to derive the SKU etc we need to duplicate the code from both hanadb and anydb as we read the DB sku from their json files.

I don't think it is required as the customers who want to deploy and anchor will typically know what SKU they want for it

rukawata (commented):
I still have some concerns on the default value for anchor vm. Please check my comments.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
KimForss: Yes - but as it would only be deployed if that section is specified we should make the decision based on that, and not overwrite it

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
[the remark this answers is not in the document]
KimForss: Option 2 is best - I'll do that

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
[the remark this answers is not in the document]
KimForss: os and auth read from db section

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
KimForss: The design for Azure is that there will be (max) 3 zones. 

The reason for this flag is that we can control that the PPG will be in a dc that supports ultra, so in your example we can quarantee that the first PPG will be created in a DC that support UItra, for the two other ones we don't know if theu land in a DC that supports Ultra or not so it may be a problem for a VM that lands in Zone 3

It may be that this is an unneccessary check but I added it as a safety check in case we have DCs that do not (yet) support ultra.

rukawata: Sorry what I meant was, we only need to ensure anchor VMs land in DC supporting ultra is when we need to deploy DB VMs that requires ultra. That means,
`local.enable_ultradisk` is true => 
we set `additional_capabilities` with `ultra_ssd_enabled` to true => 
and this should hold for true for all anchor VMs for all zones, not just one of them, this way we make sure all DB VMs land in a zone where ultra is supported.

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
rukawata: But I do not see the condition for anchor_vms...maybe it should be `local.deploy_anchor` instead of `local.zonal_deployment`?

rukawata (commented):
Still some points need to be clarified :)

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
KimForss: yes :)

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
KimForss: Actually this can be removed as Ultra is available in all the Availability Zones in all the DCs

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
```diff
@@ -0,0 +1,112 @@
...
+  disable_password_authentication = true
+
+  os_disk {
+    name                 = format("%s_%s%s", local.prefix, local.anchor_virtualmachine_names[count.index], local.resource_suffixes.osdisk)
+    caching              = "ReadWrite"
+    storage_account_type = "Standard_LRS"
+  }
+
+  additional_capabilities {
+    ultra_ssd_enabled = local.anchor_enable_ultradisk[count.index]
```
rukawata, line 48: This is only removed from windows vm :)

rukawata (commented):
Almost good! One thing you may have forgotten to remove :)

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
KimForss: addressed

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
```diff
@@ -0,0 +1,108 @@
+data "azurerm_subnet" "anchor" {
+  count                = local.deploy_anchor && local.sub_admin_exists ? 1 : 0
```
rukawata, line 2: If admin subnet exists meaning we import existing admin subnet, anchor VMs will be deployed into imported admin subnet. But if the admin subnet do not exist (eg. no dual nics and anydb scenario)?

KimForss: deploy anchor is true if it is either dual nic anydb or dual nic app or hana so the if statement should be ok

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
KimForss: doh

KimForss: I cannot see it`?

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
rukawata: deploy_anchor         = length(local.anchor) > 0 ? true : false
deployer_anchor is only decided by the anchor_vms section :( Am I missing anything？

KimForss:
That is the intention, we will only deploy if it is explicitly specified by the customer by adding the section to the json

________________________________

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/vm-anchor.tf
rukawata: so what if we have `anchor_vms` defined, no hana deploys, app deploys **without** dual nics? This situation, no admin subnet is deployed. Plus if anchor vm are deployed into an existing admin subnet deployed together with the SID, it should directly refer to that subnet resource instead of using arm_id.

KimForss:
What if we define that if an anchor vm is defined we will deploy the admin subnet even if dual nic is not specified. Or alternatively we will deploy it to the db subnet. I don’t think there would ever be a case where we need an anchor without the database

________________________________

rukawata:
> What if we define that if an anchor vm is defined we will deploy the admin subnet even if dual nic is not specified. Or alternatively we will deploy it to the db subnet. I don’t think there would ever be a case where we need an anchor without the database

We do not really need to access the anchor VM, correct? Maybe it is fine to just always deploy anchor VM into db subnet?
Whichever way we go with, need to reflect in the code :)
I'm looking again at this [doc](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-proximity-placement-scenarios), but fail to find any subnet related information. We should go check with Jurgen's team.

KimForss:
Subnets don’t really matter here for proximity, that is why the document does not mention them.

You are right we don’t need to access the anchor, in fact in almost all cases it will be turned off to save costs. Let’s change it to use the db subnet, that way it is more logical

________________________________

rukawata (approved):
Thanks. Please make sure you address below in a separate PR:
1. deploy anchor vm into db subnet
1. move db subnet into common_infra

deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -44,31 +44,63 @@ locals {
...
+  vnet_prefix                 = var.naming.prefix.VNET
+  storageaccount_name         = var.naming.storageaccount_names.SDU
+  keyvault_names              = var.naming.keyvault_names.SDU
+  virtualmachine_names        = var.naming.virtualmachine_names.ISCSI_COMPUTERNAME
+  anchor_virtualmachine_names = var.naming.virtualmachine_names.ANCHOR_VMNAME
+  anchor_computer_names       = var.naming.virtualmachine_names.ANCHOR_COMPUTERNAME
+  resource_suffixes           = var.naming.resource_suffixes
+
+
+  databases = [
```
rukawata, line 56: Due to this name change, you may need to change the code for dual_nics:

```
on ../../../../U3567-EAUS-SAP-PRD/sap-hana/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf line 119, in locals:
 119:   enable_admin_subnet = try(var.application.dual_nics, false) || try(var.databases[0].dual_nics, false) || length(local.hana-databases) > 0
```

rukawata (approved):

PR #845, opened 2020-10-18
Separate computername and resource name in apptier naming module

KimForss:
## Problem
The current naming module does not cater for zonal information in the VM resource name

## Solution
Separate the computer name and the VM resource name to two different lists, this allows us to keep the 14 character computername length but provide additional information in the VM resource name

## Tests

## Notes

### DIFF
--- a/deploy/terraform/run/sap_system/module.tf
+++ b/deploy/terraform/run/sap_system/module.tf
@@ -69,7 +69,10 @@ module "sap_namegenerator" {
   app_server_count = local.app_server_count
   web_server_count = local.webdispatcher_count
   scs_server_count = local.scs_server_count
-  zones            = local.zones
+  app_zones        = local.app_zones
+  scs_zones        = local.scs_zones
+  web_zones        = local.web_zones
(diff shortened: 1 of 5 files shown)


rukawata (commented):
Overall looks good. One comment on the variable names.
In HDB, we have `hana_server_vm_names` and `hana_computer_names`.
Can we have the same for app tier, so we have `web_server_vm_names` and `web_computer_names`/`web_server_computer_names`? The idea is so whoever reads the code understand this is computer name whereas `web_server_names` can be unspecific :)

KimForss:
> Overall looks good. One comment on the variable names.
> In HDB, we have `hana_server_vm_names` and `hana_computer_names`.
> Can we have the same for app tier, so we have `web_server_vm_names` and `web_computer_names`/`web_server_computer_names`? The idea is so whoever reads the code understand this is computer name whereas `web_server_names` can be unspecific :)

That is in the next PR

rukawata (approved):
lgtm!

PR #847, opened 2020-10-18
Add Observer VM for HA Oracle deployments

KimForss:
## Problem
HA oracle deployments require an observer VM

## Solution
Deploy an additional VM in each zone that has HA Oracle deployment. 

## Tests


deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
```diff
@@ -0,0 +1,128 @@
...
+}
+
+# Create the Linux Application VM(s)
+resource "azurerm_linux_virtual_machine" "observer" {
+  count                        = local.deploy_observer && upper(local.anydb_ostype) == "LINUX" ? length(local.zones) : 0
+  name                         = format("%s_%s%s", local.prefix, local.observer_virtualmachine_names[count.index], local.resource_suffixes.vm)
+  computer_name                = local.observer_computer_names[count.index]
+  resource_group_name          = var.resource-group[0].name
+  location                     = var.resource-group[0].location
+  //If more than one servers are deployed into a single zone put them in an availability set and not a zone
```
YunziZhangAzure, line 31: It would be helpful if you can provide the reference doc describing this logic "If more than one servers are deployed into a single zone put them in an availability set and not a zone".

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
```diff
@@ -0,0 +1,128 @@
+data "azurerm_subnet" "observer" {
```
YunziZhangAzure, line 1: I'm looking at some of the reference architecture docs, it seems observer vm and oracle database node reside in the same subnet. We already have subnet anydb resource in this module, see: https://github.com/Azure/sap-hana/blob/08ad124bf0c6bc910b93d1e05aa1022103b22339/deploy/terraform/terraform-units/modules/sap_system/anydb_node/infrastructure.tf#L60

https://github.com/Azure/sap-hana/blob/08ad124bf0c6bc910b93d1e05aa1022103b22339/deploy/terraform/terraform-units/modules/sap_system/anydb_node/infrastructure.tf#L69

YunziZhangAzure (commented):
Thanks for your pr and thanks for sharing the docs with me today!
I have read this doc: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/oracle/oracle-reference-architecture#high-availability-for-oracle-databases

I'm wondering which architecture in this doc you want to implement in the current PR? Can you provide more details?

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
KimForss: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-proximity-placement-scenarios#combine-availability-sets-and-availability-zones-with-proximity-placement-groups

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
rukawata: Agree. Should use the db subnet resource directly instead of importing it with arm_id.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
```diff
@@ -0,0 +1,128 @@
...
+  admin_username = local.observer_authentication.username
+  admin_password = local.observer_authentication.password
+
+  os_disk {
+    name                 = format("%s_%s%s", local.prefix, local.observer_virtualmachine_names[count.index], local.resource_suffixes.osdisk)
+    caching              = "ReadWrite"
+    storage_account_type = "Standard_LRS"
+  }
+
+  source_image_id = local.observer_custom_image ? local.observer_os.source_image_id : null
```
rukawata, line 113: `observer_os.source_image_id` is not defined in `deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf,` please check.

rukawata (commented):
Thanks for the PR. See comments.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
KimForss: fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -4,7 +4,7 @@ locals {
   app_oscode      = upper(var.app_ostype) == "LINUX" ? "l" : "w"
   db_platformcode = substr(var.db_platform, 0, 3)
 
-  anchor_server_names = [for idx in range(length(local.zones)) :
+  anchor_server_names = [for idx in range(length(var.zones)) :
```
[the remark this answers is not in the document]
KimForss: local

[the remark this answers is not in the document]
rukawata: The current code in the branch uses var :) please check.
image

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-observer.tf
KimForss: fixed

YunziZhangAzure:
Thanks for introducing ppg, avset and av zone to our project!

While I'm looking through our codes, I found the following codes which I want to verify with you:

https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf#L141
https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf#L138

If local.db_zone_count = 0, count.index % local.db_zone_count may lead to "divided by 0" exceptions.

https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf#L98

Since local.zones is empty array by default, db_zone_count will never get a chance to be 1. 

Can you verify if the codes above are correct?

Also, I would recommend to use local.db_server_count <= local.db_zone_count, instead of "==" to cover when the number of db nodes is less than the number of zone.

KimForss:
> Thanks for introducing ppg, avset and av zone to our project!
> 
> While I'm looking through our codes, I found the following codes which I want to verify with you:
> 
> https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf#L141
> 
> 
> https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/vm-hdb.tf#L138
> 
> If local.db_zone_count = 0, count.index % local.db_zone_count may lead to "divided by 0" exceptions.
> 
> https://github.com/Azure/sap-hana/blob/ae68be96a818583e88f74e6aad8f5916f64edc48/deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf#L98
> 
> Since local.zones is empty array by default, db_zone_count will never get a chance to be 1.
> 
> Can you verify if the codes above are correct?
> 
> Also, I would recommend to use local.db_server_count <= local.db_zone_count, instead of "==" to cover when the number of db nodes is less than the number of zone.

% 0 will not give a divide by zero exceptions 
image

 local.db_server_count <= local.db_zone_count, - good catch - but as it is not related to observer I will fix it in another PR

YunziZhangAzure:
> azurerm_availability_set.hdb[count.index % local.db_zone_count].id

Heh, looks like Terraform doesn't see "divided by zero" as an exception. 

However, there's still risk here. azurerm_availability_set.hdb[count.index % local.db_zone_count].id,
e.g.: count.index = 1,  local.db_zone_count = 0, count.index % local.db_zone_count = 1. but the length of azurerm_availability_set.hdb is 1. so azurerm_availability_set.hdb[count.index % local.db_zone_count].id-> azurerm_availability_set.hdb[1] may lead to "array index out of bound". 

So we better rethink the logic here, but this can be done in another pr.

rukawata (approved):
Please make sure the logic get corrected in a separate PR.

PR #848, opened 2020-10-19
Add support for specifying different images for the components in the application tier

### LINKED ISSUE #758: Supporting a mixture of Linux and Windows
**Is your feature request related to a problem? Please describe.**
Currently the deployment only supports one operating system for all components in the application tier

**Describe the solution you'd like**
Having the ability to provide different OS settings for the different components scs, web dispatcher

**Describe alternatives you've considered**

**Additional context**

KimForss:
## Problem
Customers want to deploy the application tier components (app, scs, web) using different operating systems, templates

## Solution
Add the option to specify different OS/template for scs and web VMs. If these are not specified deploy all the tier components using the same configuration as the application tier

## Tests
Add the following section to the json file 
```
 "os": {
      "publisher": "Oracle",
      "offer": "Oracle-Linux",
      "sku": "81",
      "version": "latest"
    },
    "scs_os": {
      "publisher": "suse",
      "offer": "sles-sap-15-sp1",
      "sku": "gen1",
      "version": "latest"
    },
    "web_os": {
      "publisher": "suse",
      "offer": "sles-sap-15-sp1",
      "sku": "gen1",
      "version": "latest"
    },
```

## Notes
Close #758

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
@@ -140,7 +140,7 @@ locals {
   // If custom image is used, we do not overwrite os reference with default value
   // If no publisher or no custom image is specified use the custom image from the app if specified
   scs_publisher_empty = try(var.application.scs_os.publisher, "") == "" ? true : false
-  scs_custom_image    = try(var.application.scs_os.source_image_id, "") && local.scs_publisher_empty && local.app_custom_image ? true : false
+  scs_custom_image    = try(var.application.scs_os.source_image_id, "") == "" && ! local.app_custom_image ? false : true
 
   scs_os = {
     "source_image_id" = local.scs_custom_image ? try(var.application.scs_os.source_image_id, var.application.os.source_image_id) : ""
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
```diff
@@ -136,6 +136,31 @@ locals {
     "version"         = try(var.application.os.version, local.app_custom_image ? "" : "latest")
   }
 
+  // OS image for all SCS VMs
+  // If custom image is used, we do not overwrite os reference with default value
+  scs_custom_image = try(var.application.scs_os.source_image_id, "") != "" ? true : false
+
+  scs_os = {
+    "source_image_id" = local.scs_custom_image ? var.application.scs_os.source_image_id : ""
```
rukawata, line 144: How would this handle situation where custom image is provided in `app_os.source_image_id` but not `scs_os.source_image_id`? Two options:
1. use `scs_custom_image` as a flag regardless using `application.os.source_image_id` or `application.scs_os.source_image_id`.
2. use `scs_custom_image` as a flag when scs uses it's own custom image. Will have to change the logic in the VM section.

rukawata (commented):
One concern about the logic. Please check. Thanks!

deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
KimForss: Added som additional logic - if the scs node has no image id nor a publisher then use the custom image from app if specified otherwise not

rukawata: I think to use publisher as condition for using custom image for scs may not be the best idea, what if we just want to override the offer/sku/version and use the default publisher - meaning no need to provide publisher. 
Here is what I propose:

`scs_custom_image    = try(var.application.scs_os.source_image_id, "") == "" &&  ! local.app_custom_image ? false : true
`

This means, if no app_custom_image provided or no `scs_os.source_image_id` provided, we assume scs do not use custom image. The rest code should be fine stay the way they are.

KimForss: changed

rukawata (approved):
LGTM!

rukawata (approved):

PR #858, opened 2020-10-20
Clean software block and the related logic

YunziZhangAzure:
## Problem
This pr is to address comments from previous discussion. Also fix a bug in previous pr. See Solution section down below, for which task items are included in this pr

## Solution

- Update test pipeline to support tfstate_resource_id in tf-saplandscape
- clean software block
- Get rid of importing saplibrary logic in sap system module 
- Merge KeyVault access policy into one Keyvault resource
- Clean the "depends on" keyvault access policy resource
- Fix the existing bug in the usage of Keyvault access policy.
See the erro here:https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=12850&view=logs&j=26691db3-61d7-5524-8085-39e615892bb3&t=bd477c5e-eed1-5ccb-a423-ea4312ca6e75

See official doc from Terraform: https://www.terraform.io/docs/providers/azurerm/r/key_vault.html#object_id
The object id of keyvault access policy should be populated with the object id of service principal instead of the application id which is stored in Keyvault of deployer right now. To obtain the object id, a new provider azuread and the data resource azuread_service_principal are used.
sap library has similar bug, fixed here: https://github.com/Azure/sap-hana/pull/861

- Change the approach when retrieving value of "sid_public_key_secret_name" in create-sapsystem-steps.yml. Using command "az secret" doesn't work because it is run on deployer and the deployer-msi may not have permission to list all the secrets in the keyvault of saplandscape.

## Tests

See the results in the test pipeline down below.

## Notes


rukawata (approved):
thanks! LGTM!

PR #861, opened 2020-10-21
Keyvault access policy should use object id of service principal

YunziZhangAzure:
## Problem
Fix the existing bug in the usage of Keyvault access policy.

The object id of keyvault access policy should be populated with the object id of service principal instead of the application id which is stored in Keyvault of deployer right now. 

## Solution
To obtain the object id, a new provider azuread and the data resource azuread_service_principal are used.

## Tests
See the result in the test pipeline down below

image

## Notes
sap-hana-pipeline-app 
The object id used in configuring access policy is the object id of service principal instead of the object id of application.


rukawata (approved):
Thanks for addressing this issue.

PR #864, opened 2020-10-26
[NO DEPLOYER VM] Ability to deploy the deployer infrastructure without the deployer VM

KimForss:
## Problem
Customers who use Azure DevOps or other execution environments do not want a deployment with an VM that they will not use. This case applies for Terraform Enterprise as well

## Solution
Add an option in the deployers section that blocks the deployment of the VM. The default value for the setting is true, if the customer wants to deploy without the deployer VM they will need to explicitly add the section to the parameters.

 "deployers": [{
    "deploy_vm" : "false"
  }],

## Tests
Add the section to the json file and deploy the deployer (either bootstrap or run)

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf
@@ -52,6 +52,8 @@ locals {
 
   // Deployer(s) information with default override
   enable_deployers = length(local.deployer_input) > 0 ? true : false
+
+  // Provide the ability to deploy without the VM
   enable_vm        = try(local.deployer_input.deploy_vm, true)
 
   deployers = [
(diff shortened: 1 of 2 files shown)


YunziZhangAzure (commented):
I want to know more context of the requirement. 
"Customers who use Azure DevOps or other execution environments do not want a deployment with an VM that they will not use." Do customers want the deployer's vnet and subnet?

KimForss:
> I want to know more context of the requirement.
> "Customers who use Azure DevOps or other execution environments do not want a deployment with an VM that they will not use." Do customers want the deployer's vnet and subnet?
If they do not want the deployer they do not want the vnet either

YunziZhangAzure:
> > I want to know more context of the requirement.
> > "Customers who use Azure DevOps or other execution environments do not want a deployment with an VM that they will not use." Do customers want the deployer's vnet and subnet?
> > If they do not want the deployer they do not want the vnet either

I see. I think we already have such logic:

https://github.com/Azure/sap-hana/blob/08ad124bf0c6bc910b93d1e05aa1022103b22339/deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf#L54

We don't need to add more field e.g: "deploy_vm" : "false" in input json. We can make changes to the existing codes to make it support this feature.

PR #865, opened 2020-10-26
Add the ability to specify the separator used between the resource prefix and the resource name

KimForss:
## Problem
The current solution has a hardcoded separator "_" that separates the resource prefix and the resource name. For customers who implement their naming convention there is no way of overriding the separator

## Solution
Make the separator to be part of the naming module with the default value "_", if customers want to override the naming module they can specify the separator to another string or keep it empty

 //The separator to use between the prefix and resource name
  separator = "_"

and return it from the naming module
output.tf
....
 separator = local.separator

the naming logic becomes
name                = format("%s%s%s%s", local.prefix, **var.naming.separator**, local.app_virtualmachine_names[count.index], local.resource_suffixes.vm)

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
+++ b/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
@@ -1,6 +1,9 @@
 locals {
 
-  sdu_name      = length(var.codename) > 0 ? upper(format("%s-%s-%s_%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.codename, var.sap_sid)) : upper(format("%s-%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid))
+  sdu_name = length(var.codename) > 0 ? (
+    upper(format("%s-%s-%s_%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.codename, var.sap_sid))) : (
+    upper(format("%s-%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid))
+  )
   deployer_name = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.dep_vnet_verified))
   vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
   library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
--- a/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
@@ -74,7 +74,7 @@ resource "azurerm_linux_virtual_machine" "app" {
 # Create the Windows Application VM(s)
 resource "azurerm_windows_virtual_machine" "app" {
   count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? local.application_server_count : 0) : 0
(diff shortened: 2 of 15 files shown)


deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
```diff
@@ -74,7 +74,7 @@ resource "azurerm_linux_virtual_machine" "app" {
 # Create the Windows Application VM(s)
 resource "azurerm_windows_virtual_machine" "app" {
   count               = local.enable_deployment ? (upper(local.app_ostype) == "WINDOWS" ? local.application_server_count : 0) : 0
-  name                = format("%s_%s%s", local.prefix, local.app_virtualmachine_names[count.index], local.resource_suffixes.vm)
+  name                = format("%s%s%s", local.prefix, var.naming.separator, local.app_virtualmachine_names[count.index], local.resource_suffixes.vm)
```
YunziZhangAzure, line 77: Missing one %s?

deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
```diff
@@ -126,7 +126,7 @@ resource "azurerm_windows_virtual_machine" "app" {
 # Creates managed data disk
 resource "azurerm_managed_disk" "app" {
   count                = local.enable_deployment ? length(local.app-data-disks) : 0
-  name                 = format("%s_%s%s", local.prefix, local.app_virtualmachine_names[count.index], local.app-data-disks[count.index].suffix)
+  name                 = format("%s%s%s", local.prefix, var.naming.separator, local.app_virtualmachine_names[count.index], local.app-data-disks[count.index].suffix)
```
YunziZhangAzure, line 129: missing one %s?

deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
KimForss: yes :(

deploy/terraform/terraform-units/modules/sap_system/app_tier/vm-app.tf
KimForss: yes, fixed

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
```diff
@@ -1,6 +1,9 @@
 locals {
 
-  sdu_name      = length(var.codename) > 0 ? upper(format("%s-%s-%s_%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.codename, var.sap_sid)) : upper(format("%s-%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid))
+  sdu_name = length(var.codename) > 0 ? (
+    upper(format("%s-%s-%s_%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.codename, var.sap_sid))) : (
```
YunziZhangAzure, line 4: Will the underscore here be replaced by local.separator?

rukawata (commented):
Same question as Yunzi, otherwise looks good.

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
KimForss: I would say no - as this separator is only for the case when the codename is specified. And also if the customer wants to change the separator they will need to make the changes in the module so thay can change this one as well.

rukawata (approved):
lgtm!

PR #866, opened 2020-10-26
Remove role assignment to deployer/ Modify input JSON

YunziZhangAzure:
## Problem
1. User Access Administrator role assignment to deployer msi is not being used. 
2. We need a way to track the input JSON with full set of parameters and input JSON with minimum set of parameters.

## Solution
1. Add input JSON to track the full set of parameters
2. remove User Access Administrator role assignment to deployer msi

## Tests

## Notes
{deployer/saplibrary/saplandscape/sapsystem}_full.json in this pr are added as placeholders, they are not consolidated.
Will keep updating them in the future.

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/deployer_full.json
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer_full.json
@@ -1,0 +1,18 @@
+{
+    "infrastructure": {
+        "region": "eastus",
+        "environment": "np",
+        "vnets": {
+            "management": {
+                "name": "DEP00",
+                "address_space": "10.0.0.0/26",
+                "subnet_mgmt": {
+                    "prefix": "10.0.0.16/28"
+                }
+            }
+        }
+    },
+    "options": {
+        "enable_secure_transfer": true
+    }
+}
--- a/deploy/terraform/run/sap_landscape/saplandscape.json
+++ b/deploy/terraform/run/sap_landscape/saplandscape.json
@@ -2,9 +2,6 @@
     "infrastructure": {
         "environment": "np",
         "region": "eastus",
-        "resource_group": {
-            "name": "azure-test-saplandscape-rg"
-        },
         "vnets": {
             "management": {
                 "tfstate_resource_id": "",
@@ -21,14 +18,6 @@
         "iscsi": {
             "iscsi_count": 3
         }
-    },
-    "databases": [
-        {
-            "platform": "NONE"
-        }
-    ],
-    "application": {
-        "enable_deployment": false
     },
     "sshkey": {},
     "options": {}
--- a/deploy/terraform/run/sap_landscape/saplandscape_full.json
+++ b/deploy/terraform/run/sap_landscape/saplandscape_full.json
@@ -3,6 +3,8 @@
         "environment": "np",
         "region": "eastus",
         "resource_group": {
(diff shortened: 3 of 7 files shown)


deploy/terraform/bootstrap/sap_deployer/deployer_full.json
```diff
@@ -0,0 +1,18 @@
+{
+    "infrastructure": {
```
rukawata, line 2: It is great that we have this full version that track all possible configurations.
We can also include things like deployer VM username, SSH key override, etc...But we can address this in a new PR.

deploy/terraform/run/sap_landscape/saplandscape.json
```diff
@@ -1,10 +1,8 @@
 {
     "infrastructure": {
+        "environment": "np",
         "region": "eastus",
-        "environment": "global",
         "resource_group": {
-            "is_existing": "false",
-            "arm_id": "",
             "name": "azure-test-saplandscape-rg"
```
rukawata, line 6: rg name will be generated by default, so this can be in the full version but not in the minimal version.

deploy/terraform/run/sap_landscape/saplandscape_full.json
```diff
@@ -0,0 +1,37 @@
...
+        },
+        "iscsi": {
+            "iscsi_count": 3
+        }
+    },
+    "databases": [
+        {
+            "platform": "NONE"
+        }
+    ],
```
rukawata, line 31: Since default we pass empty list to databases, I think this can also be removed since we do not deploy database with saplandscape. Same applies to the min version.

deploy/terraform/run/sap_landscape/saplandscape_full.json
```diff
@@ -0,0 +1,37 @@
...
+    "application": {
+        "enable_deployment": false
+    },
```
rukawata, line 34: Same, by default is will not be deployed.

rukawata (commented):
Thanks for making this effort!
Most part looks good, but have some comment on what should be in the default min version, what should be in the full version. We can discuss offline.

deploy/terraform/bootstrap/sap_deployer/deployer_full.json
YunziZhangAzure: Ok! Will do this in the future pr.

YunziZhangAzure:
> Thanks for making this effort!
> Most part looks good, but have some comment on what should be in the default min version, what should be in the full version. We can discuss offline.

My initial thought is to have a place track

deploy/terraform/run/sap_landscape/saplandscape.json
YunziZhangAzure: updated.

deploy/terraform/run/sap_landscape/saplandscape_full.json
YunziZhangAzure: Done.

deploy/terraform/run/sap_landscape/saplandscape_full.json
YunziZhangAzure: done

rukawata (approved):
lgtm!

PR #891, opened 2020-11-05
Add the ostype variables for scs and web

KimForss:
Add the ostype for scs and web

## Problem
The scs and web ostype variables are missing

## Solution
 scs_ostype = try(var.application.scs_os.os_type, "Linux")
 web_ostype = try(var.application.web_os.os_type, "Linux")

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
@@ -196,6 +196,7 @@ locals {
   // If custom image is used, we do not overwrite os reference with default value
   // If no publisher or no custom image is specified use the custom image from the app if specified
   scs_custom_image = try(var.application.scs_os.source_image_id, "") == "" && ! local.app_custom_image ? false : true
+  scs_ostype       = try(var.application.scs_os.os_type, "Linux")
 
   scs_os = {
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
```diff
@@ -196,6 +196,7 @@ locals {
   // If custom image is used, we do not overwrite os reference with default value
   // If no publisher or no custom image is specified use the custom image from the app if specified
   scs_custom_image = try(var.application.scs_os.source_image_id, "") == "" && ! local.app_custom_image ? false : true
+  scs_ostype       = try(var.application.scs_os.os_type, "Linux")
```
rukawata, line 199: in the `vm-scs.tf` and `vm-webdisp.tf`, we use `local.app_ostype` in the condition. So this variable is actually not used.
One option is set default to app_ostype instead of Linux, then replace the condition in the vm-*.tf to use scs_ostype and web_ostype.

rukawata (commented):
Please check the comments. Thanks!

deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
KimForss: good catch

rukawata (approved):
Thanks for addressing the comments!

PR #895, opened 2020-11-05
Add ultradisk flag and accelerated networking flag to anchor

KimForss:
## Problem
The anchor needs to be able to control if the cluster supports ultra and accelerated networking

## Solution
Control these with parameters

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
@@ -114,15 +114,16 @@ locals {
 
   //Anchor VM
 
-  anchor                      = try(local.var_infra.anchor_vms, {})
-  deploy_anchor               = length(local.anchor) > 0 ? true : false
-  anchor_size                 = try(local.anchor.sku, "Standard_D8s_v3")
-  anchor_authentication       = try(local.anchor.authentication, local.db_auth)
-  anchor_auth_type            = try(local.anchor.authentication.type, "key")
-  enable_anchor_auth_password = local.deploy_anchor && local.anchor_auth_type == "password"
-  enable_anchor_auth_key      = local.deploy_anchor && local.anchor_auth_type == "key"
-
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_system/common_infrastructure/variables_local.tf
```diff
@@ -114,15 +114,16 @@ locals {
...
-
-  anchor_nic_ips = local.sub_admin_exists ? try(local.anchor.nic_ips, []) : []
+  anchor                        = try(local.var_infra.anchor_vms, {})
+  deploy_anchor                 = length(local.anchor) > 0 ? true : false
+  anchor_size                   = try(local.anchor.sku, "Standard_D8s_v3")
+  anchor_authentication         = try(local.anchor.authentication, local.db_auth)
+  anchor_auth_type              = try(local.anchor.authentication.type, "key")
+  enable_anchor_auth_password   = local.deploy_anchor && local.anchor_auth_type == "password"
+  enable_anchor_auth_key        = local.deploy_anchor && local.anchor_auth_type == "key"
+  enable_anchor_ultra           = try(local.anchor.enable_ultra, [false, false, false])
```
rukawata, line 124: What's the reason of adding this parameter back?

KimForss: To give the ability to control which zone requires Ultra disk - without that granularity I got allocation errors as the anchor got deployed to a cluster that does not support ultra and accelerated networking at the same time

rukawata: But if we know that the VM needs ultra disks (controlled by local.enable_ultradisk), then all the zones should have ultradisk support, meaning the anchor will be a VM with ultradisk.

This way we make sure when ultra is required, the anchor deploys with it, rather than manually control a list that maps to the zones.

rukawata (approved):

PR #899, opened 2020-11-06
fix a bug about storage_list when deploying saplandscape

YunziZhangAzure:
## Problem
A bug when deploying saplandscape, var.database is empty, which leads to an error when referecing.

## Solution
Add a condition. If var.database is empty list, return []

## Tests

See test pipeline:

https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=14445&view=results

## Notes


KimForss (approved):
Good catch

rukawata (approved):
Thanks for catching this bug!

PR #901, opened 2020-11-06
reference azure resource group implicitly to add dependency

YunziZhangAzure:
## Problem
When provisioning kv of saplandscape, the following error exists:

image

## Solution
reference azure resource group implicitly to add dependency between RG and KV, so that RG will be provisioned before KVs.

## Tests

https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=14577&view=results

## Notes


rukawata (approved):
thanks for fixing that!

Issue #776, opened 2020-09-10
SAP Workload VNET


mkdeegan:
**SAP Workload VNET**
Separate the the tasks related to the creation of the SAP Workload VNET and resources common to all SDU's from the Monolithic Code base, to a standalone Terraform execution unit.

PR #902, opened 2020-11-06
Create Saplandscape module

YunziZhangAzure:
## Problem

## Solution

Based on requirements, in sap_landscape, the following resources are provisioned
SAP vnet 

iSCSI subnet + nsg + iSCSI devices if defined(required) 

prvt and user KVs, with SSH key pair stored in user KV. 

## Tests
see test pipeline down below.

integration test

https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=14753&view=results

## Notes

The output is not finalized. Will add more fields later based on the demands.


deploy/terraform/run/sap_landscape/output.tf
```diff
@@ -1,7 +1,19 @@
...
+
+output "landscape_key_vault_user_arm_id" {
+  value = try(module.sap_landscape.kv_user[0].id, "")
+}
+
+output "sid_public_key_secret_name" {
+  value = try(module.sap_landscape.sid_public_key_secret_name, "")
+}
+
+output "nics_iscsi" {
```
rukawata, line 17: since this is already the private ip, maybe rename to like `iscsi_private_ip` rather than nics_iscsi?

deploy/terraform/terraform-units/modules/sap_landscape/infrastructure.tf
```diff
@@ -0,0 +1,64 @@
...
+resource "azurerm_resource_group" "resource_group" {
+  count    = local.rg_exists ? 0 : 1
+  name     = local.rg_name
+  location = local.region
+}
+
+// Imports data of existing resource group
+data "azurerm_resource_group" "resource_group" {
+  count = local.rg_exists ? 1 : 0
+  name  = split("/", local.rg_arm_id)[4]
```
rukawata, line 16: I know we use this in the previous code, but local.rg_name actually already considered the name for imported resource_group. Maybe we can change that in a new PR :)

`rg_name   = local.rg_exists ? try(split("/", local.rg_arm_id)[4], "") : try(local.var_rg.name, format("%s%s", local.prefix, local.resource_suffixes.sdu_rg))`

deploy/terraform/terraform-units/modules/sap_landscape/key_vault_sap_landscape.tf
```diff
@@ -0,0 +1,126 @@
...
+  count        = local.enable_landscape_kv ? 1 : 0
+  name         = format("%s-sid-sshkey-pub", local.prefix)
+  value        = local.sid_public_key
+  key_vault_id = azurerm_key_vault.kv_user[0].id
+}
+
+// random bytes to product
+resource "random_id" "saplandscape" {
+  byte_length = 4
+}
```
rukawata, line 126: Not sure...is this used anywhere? I think after having the naming module, this may not be needed.
My bad, probably I forgot to remove it when implementing KV with naming module.

rukawata (commented):
Thanks! Overall looks good!
Some nit comments :)

deploy/terraform/terraform-units/modules/sap_landscape/key_vault_sap_landscape.tf
YunziZhangAzure: Good catch! Will remove this.

deploy/terraform/terraform-units/modules/sap_landscape/infrastructure.tf
YunziZhangAzure: good point!

deploy/terraform/run/sap_landscape/output.tf
YunziZhangAzure: Right!

rukawata (approved):
Thanks for addressing the comments!
I think landscape_infrastructure output has room to improve, let's do that once seperation is completed :)

rukawata:
Closes #776

PR #904, opened 2020-11-07
Fixes for deployments without zone information

KimForss:
## Problem
Deployment without zonal information fails

## Solution
Use  max(local.zone_count,1) instead of local.zone_count

## Tests

## Notes

### DIFF
--- a/deploy/terraform/run/sap_system/imports.tf
+++ b/deploy/terraform/run/sap_system/imports.tf
@@ -11,7 +11,7 @@ data "terraform_remote_state" "deployer" {
     container_name       = local.tfstate_container_name
     key                  = local.deployer_tfstate_key
     subscription_id      = local.saplib_subscription_id
-    use_msi              = true
+   // use_msi              = true
   }
 }
 
--- a/deploy/terraform/run/sap_system/imports.tf
+++ b/deploy/terraform/run/sap_system/imports.tf
@@ -11,7 +11,7 @@ data "terraform_remote_state" "deployer" {
     container_name       = local.tfstate_container_name
     key                  = local.deployer_tfstate_key
     subscription_id      = local.saplib_subscription_id
-   // use_msi              = true
+    use_msi              = true
   }
 }
 
--- a/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
@@ -56,7 +56,7 @@ locals {
   // Zones
   zones            = try(local.anydb.zones, [])
   zonal_deployment = length(local.zones) > 0 ? true : false
-  db_zone_count    = try(length(local.zones), 1)
+  db_zone_count    = length(local.zones)
 
   // SAP vnet
   var_infra       = try(var.infrastructure, {})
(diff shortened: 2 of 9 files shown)


YunziZhangAzure (commented):
Looks good to me!
This pr should fix the error in the release pipeline. I will try re-running the release pipeline after this pr is merged. 

We can enhance the codes about zone/avset/ppg in the future.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
```diff
@@ -55,14 +55,14 @@ resource "azurerm_linux_virtual_machine" "dbserver" {
   resource_group_name = var.resource_group[0].name
   location            = var.resource_group[0].location
 
-  proximity_placement_group_id = local.zonal_deployment ? var.ppg[count.index % local.db_zone_count].id : var.ppg[0].id
+  proximity_placement_group_id = local.zonal_deployment ? var.ppg[count.index % max(local.db_zone_count,1)].id : var.ppg[0].id
   //Ultra disk requires zonal deployment
   availability_set_id = local.enable_ultradisk ? null : (
-    local.zonal_deployment && local.db_server_count == local.db_zone_count ? null : azurerm_availability_set.anydb[count.index % local.db_zone_count].id
+    local.zonal_deployment && local.db_server_count == local.db_zone_count ? null : azurerm_availability_set.anydb[count.index % max(local.db_zone_count,1)].id
```
rukawata, line 61: in the `variable_local.tf`, `db_zone_count` is >= 1 anyway, I do not see the point of changing to use max.

`db_zone_count    = try(length(local.zones), 1)`

I think it would make more sense if `db_zone_count    = length(local.zones)`, then the max logic would make sense.

rukawata (commented):
Thanks for working on fixing this.
Some comments.

deploy/terraform/run/sap_system/imports.tf
```diff
@@ -11,7 +11,7 @@ data "terraform_remote_state" "deployer" {
     container_name       = local.tfstate_container_name
     key                  = local.deployer_tfstate_key
     subscription_id      = local.saplib_subscription_id
-    use_msi              = true
+   // use_msi              = true
```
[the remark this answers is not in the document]
KimForss: sorry - it was left there so that I can test locally

deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
KimForss: actually the try in variables_local.tf  db_zone_count    = try(length(local.zones), 1) -> 0 for an empty list

rukawata: You are right! My bad.
But if we use the max logic, do you still see the need to default this value to 1?

KimForss: no, will change that

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -56,7 +56,7 @@ locals {
   // Zones
   zones            = try(local.anydb.zones, [])
   zonal_deployment = length(local.zones) > 0 ? true : false
-  db_zone_count    = try(length(local.zones), 1)
+  db_zone_count    = length(local.zones)
```
rukawata, line 59: Please change the same in `hdb_node/variables_local.tf`.

KimForss: done

rukawata (approved):
Thanks for fixing the bug.

PR #914, opened 2020-11-11
Fix ObserverName

KimForss:
## Problem
The Observer VM has an incorrect name it has LIN instead of l as part of the name

## Solution
Use the correct variable in vm.tf

## Tests

## Notes


rukawata (commented):
`db_platformcode = substr(var.db_platform, 0, 3)` is still defined in the same file.
You may want to change them all.

KimForss:
> `db_platformcode = substr(var.db_platform, 0, 3)` is still defined in the same file.
> You may want to check and remove if not needed.

Removed

rukawata (approved):

rukawata (approved):

PR #916, opened 2020-11-12
Region fix, zone logic validation, DNS variables

KimForss:
## Problem

- The naming logic is not detecting characters that are not allowed in for instance storage account names.

- East US2 has the wrong code

- The zone logic for naming is not correct in all cases

- There second NIC allows for a separate host name but the code does not address it

## Solution
The naming module has fixes and updates

## Tests

## Notes
The PR also lays the foundation for support of dynamic IP addresses

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
+++ b/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
@@ -8,10 +8,11 @@ locals {
   vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
   library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
 
-  sdu_storageaccount_name            = replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
-  library_storageaccount_name        = replace(lower(format("%s%s%ssaplib%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
-  terraformstate_storageaccount_name = replace(lower(format("%s%s%stfstate%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
-  deployer_storageaccount_name       = replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
-  vnet_storageaccount_name           = replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
+  // Storage account names must be between 3 and 24 characters in length and use numbers and lower-case letters only. The name must be unique.
+  sdu_storageaccount_name            = substr(replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/",""),0,var.azlimits.stgaccnt)
+  library_storageaccount_name        = substr(replace(lower(format("%s%s%ssaplib%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/",""),0,var.azlimits.stgaccnt)
+  terraformstate_storageaccount_name = substr(replace(lower(format("%s%s%stfstate%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/",""),0,var.azlimits.stgaccnt)
+  deployer_storageaccount_name       = substr(replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified)),"/[^a-z0-9]/",""),0,var.azlimits.stgaccnt)
+  vnet_storageaccount_name           = substr(replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified)),"/[^a-z0-9]/",""),0,var.azlimits.stgaccnt)
 
 }
--- a/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
+++ b/deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
@@ -8,10 +8,10 @@ locals {
   vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
   library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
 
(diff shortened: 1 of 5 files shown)


deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -108,4 +108,37 @@ locals {
     )
   ]
 
+  //For customer who want to have an alternative name for the second IP address
+  app_computer_dnsnames = [for idx in range(var.app_server_count) :
```
rukawata, line 112: maybe name it `app_secondary_dnsnames` to be consistent with the output var?
Same applies to the below vars.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -146,7 +146,7 @@ variable region_mapping {
     "centralus"          = "ceus"
     "eastasia"           = "eaas"
     "eastus"             = "eaus"
-    "eastus2"            = "eau2"
+    "eastus2"            = "eus2"
```
rukawata, line 149: Can you separate this change into a PR - so I can also add this change to beta/2.3?
Or you can also create a separate PR against beta/2.3 for this.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -264,14 +264,19 @@ variable db_zones {
   default     = []
 }
 
+variable custom_prefix {
```
rukawata, line 267: I think I need more background on this `custom_prefix`, as of where it will be used or how would it be passed into naming module from SDU/SID.

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -264,14 +264,19 @@ variable db_zones {
...
 
 locals {
 
   location_short = upper(try(var.region_mapping[var.location], "unkn"))
 
   env_verified      = upper(substr(var.environment, 0, var.sapautomation_name_limits.environment_variable_length))
-  vnet_verified     = upper(substr(var.sap_vnet_name, 0, var.sapautomation_name_limits.sap_vnet_length))
-  dep_vnet_verified = upper(substr(var.management_vnet_name, 0, var.sapautomation_name_limits.sap_vnet_length))
+  vnet_verified     = upper(trim(substr(var.sap_vnet_name, 0, var.sapautomation_name_limits.sap_vnet_length), "-_"))
+  dep_vnet_verified = upper(trim(substr(var.management_vnet_name, 0, var.sapautomation_name_limits.sap_vnet_length), "-_"))
```
rukawata, line 279: Does this mean that vnet name should not have - or _?
If so, can you add a comment to make it clear for other readers :)

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
```diff
@@ -8,10 +8,10 @@ locals {
   vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
   library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
 
-  sdu_storageaccount_name            = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified))
-  library_storageaccount_name        = lower(format("%s%s%ssaplib%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified))
-  terraformstate_storageaccount_name = lower(format("%s%s%stfstate%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified))
-  deployer_storageaccount_name       = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified))
-  vnet_storageaccount_name           = lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified))
+  sdu_storageaccount_name            = replace(lower(format("%s%s%sdiag%s", local.env_verified, local.location_short, local.vnet_verified, local.random_id_verified)),"/[^a-z0-9]/","")
```
rukawata, line 11: Good that you enforce the SA name limit here.
Can you add a comment line with reference to the rules and the rule itself?
eg. `Storage account names must be between 3 and 24 characters in length and use numbers and lower-case letters only. The name must be unique.`

rukawata (commented):
Still fixing the UNIT test pipeline.
Most part looks good. Some small comments, please check :)

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
YunziZhangAzure: Is it possible set regex to  a local variable and pass it to replace function? It may look simpler?

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
KimForss: I'll do a separate PR on that against beta/2.3

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
KimForss: It is fine if the Vnet name has "-" or "_" but - as the vnet_verified will be part of the storage account name we need to remove those, that is why the replace is on the storage account names

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
KimForss: The regex part is just for the string replace so not sure if moving it to a variable makes it much easier :)

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
KimForss: For instance Achmea wants to have a prefix that is different for different environments and also different from the rg name, Scheer and H&M want that too,
For Achmea the logic is different for the Windows and the Linux environments. By providing a way to pass that in as a parameter they will not need to add that logic into the naming module. 

The way they use it from their customized module.tf is by reading it from a variable and then passing it into the naming module.

module "sap_namegenerator" {
  source               = "../../achmea/sap_namegenerator"
  custom_prefix   = var.prefix
  environment     = var.infrastructure.environment

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
KimForss: done

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
rukawata: I see, so basically this will override the naming convention completely. 
Ok, we can get this code in, and let's see how this will be used in the module later on.

rukawata (approved):

YunziZhangAzure (approved):

PR #921, opened 2020-11-15
Add zone logic for ultra disk

KimForss:
The zone logic was missing the ultradisk clause

## Problem
The zone logic was missing the ultradisk clause

## Solution

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
@@ -58,10 +58,7 @@ resource "azurerm_linux_virtual_machine" "dbserver" {
     local.zonal_deployment && local.db_server_count == local.db_zone_count ? null : azurerm_availability_set.anydb[count.index % max(local.db_zone_count, 1)].id
   )
 
-  zone = local.zonal_deployment ? (
-    local.db_server_count == local.db_zone_count ? local.zones[count.index % max(local.db_zone_count, 1)] : null) : (
-    null
-  )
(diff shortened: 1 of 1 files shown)


deploy/terraform/terraform-units/modules/sap_system/anydb_node/vm-anydb.tf
```diff
@@ -58,10 +58,7 @@ resource "azurerm_linux_virtual_machine" "dbserver" {
     local.zonal_deployment && local.db_server_count == local.db_zone_count ? null : azurerm_availability_set.anydb[count.index % max(local.db_zone_count, 1)].id
   )
 
-  zone = local.zonal_deployment ? (
-    local.db_server_count == local.db_zone_count ? local.zones[count.index % max(local.db_zone_count, 1)] : null) : (
-    null
-  )
+  zone = local.enable_ultradisk || local.db_server_count == local.db_zone_count ? local.zones[count.index % max(local.db_zone_count, 1)] : null
```
rukawata, line 61: I understand this is now consistent with hdb module logic for zone and av set.
Can we revisit the logic for av set? Do we really need the zonal_deployment variable in the condition? Can we make the avset logic simpler and just the opposite for zone?

KimForss: We can :) - not sure if I can though 😄

rukawata: Ok, we will first get this change in so it's consistent, then let's refine the logic.

rukawata (approved):

PR #922, opened 2020-11-15
Dont deploy webdipatcher lb if no webdispatchers

KimForss:
## Problem
If there are no web dispatcher there is no need for a load balancer

## Solution
Control the creation of the load balancer based on the webdispatcher_count

## Tests

## Notes


rukawata (approved):
LGTM!
Can we have another PR to do the same for SCS?

PR #924, opened 2020-11-16
Only deploy scs loadbalancer if scs servers are deployed

KimForss:
## Problem
The current design will always deploy an scs load balancer if an app tier is deployed

## Solution
Control this the same way as the load balancer

## Tests

## Notes


rukawata (approved):
Thanks. LGTM!

PR #925, opened 2020-11-16
Trim whitespace from custom prefix

KimForss:
## Problem
An Azure resource name cannot begin with whitespace

## Solution
Trim the whitespace from the resource prefix. The reason why the trimming is in the module and not the naming module is that it allows the customers to override the prefix by providing a value for custom_prefix.

## Tests

## Notes


rukawata (approved):
LGTM!

PR #930, opened 2020-11-17
[ANF] Support deployment to existing avsets (db)

KimForss:
## Problem
Customers who want to use ANF will need to pin them to an existing empty availity set to guarantee close proximity. Our current design will always create the availability set.

## Solution
Provide the ability to provide the arm_id of the availability set and use that in the automation.

## Tests
Add the ARM ID of the existing availability set to the input json in the database section

  "avset_arm_ids": [
        "/subscriptions/xxxx/resourceGroups/PROTO-SID-RG/providers/Microsoft.Compute/availabilitySets/PROTO-SID_db_avset"
      ]

## Notes

### DIFF
--- a/deploy/terraform/run/sap_landscape/output.tf
+++ b/deploy/terraform/run/sap_landscape/output.tf
@@ -14,6 +14,11 @@ output "iscsi_private_ip" {
   value = try(module.sap_landscape.nics_iscsi[*].private_ip_address, [])
 }
 
-output "landscape_infrastructure" {
-  value = try(module.sap_landscape.infrastructure_w_defaults, {})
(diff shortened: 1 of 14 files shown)


rukawata (commented):
Overall looks good. Can you rename `sapsystem_zones.json` into `sapsystem_full.json` and add this key there?

sapsystem_full.json will be the place we track all the possible inputs.

KimForss:
> Overall looks good. Can you rename `sapsystem_zones.json` into `sapsystem_full.json` and add this key there?
> 
> sapsystem_full.json will be the place we track all the possible inputs.

Done

rukawata (approved):
LGTM!

PR #932, opened 2020-11-17
Remove ansible automatic execution logic

YunziZhangAzure:
## Problem
Currently, setting ansible_execution as true will automatically start running ansible which is against the requirement.

## Solution
Remove the related logic.

## Tests

## Notes


rukawata (commented):
Thanks for addressing this!
Although not impact the functional side, probably also need to remove related code from sap_system.json (`"ansible_execution": false`) and variables_local.tf (  `ansible_execution      = try(var.options.ansible_execution, false)`).

YunziZhangAzure:
> Thanks for addressing this!
> Although not impact the functional side, probably also need to remove related code from sap_system.json (`"ansible_execution": false`) and variables_local.tf ( `ansible_execution = try(var.options.ansible_execution, false)`).

Good catch! It is updated.

rukawata (approved):
LGTM!

PR #935, opened 2020-11-18
Trim output of saplandscape

YunziZhangAzure:
## Problem
The output of saplandscape needs to be trimmed. Only pass the parameters being used

## Solution
Since Ansible needs iscsi authentication type and username, only these two variables will be output.

## Tests
https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=15622&view=results

## Notes


rukawata (approved):
LGTM! Thanks for addressing Morgan's comments :)

PR #937, opened 2020-11-19
import existing key vaults to sap deployer

YunziZhangAzure:
## Problem
Based on the backlog feature of phase 2

## Solution
use case 1:
The user specifies user_key_vault_id in input json, then the corresponding key vault will be used as the user keyvault of deployer.
There should exist {environment}-sshkey-pub, {environment}-sshkey in this keyvault to fetch public key from. 

use case 2:
The user specifies private_key_vault_id in input json, then the corresponding key vault will be used as the private keyvault of deployer.

Note:
As soon as bootstraping deployer is done, the user should make sure deployer-msi has access to the secrets of keyvault for subsequent steps in the workflow.

## Tests
Test locally

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
+++ b/deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
@@ -2,8 +2,8 @@
 data "azurerm_client_config" "deployer" {}
 
 resource "azurerm_key_vault" "kv_prvt" {
-  count                      = local.enable_deployers ? 1 : 0
-  name                       = local.keyvault_names.private_access
+  count                      = (local.enable_deployers && ! local.prvt_kv_exist) ? 1 : 0
+  name                       = local.prvt_kv_name
   location                   = azurerm_resource_group.deployer[0].location
   resource_group_name        = azurerm_resource_group.deployer[0].name
   tenant_id                  = data.azurerm_client_config.deployer.tenant_id
@@ -14,8 +14,15 @@ resource "azurerm_key_vault" "kv_prvt" {
   sku_name = "standard"
 }
 
+// Import an existing private Key Vault
+data "azurerm_key_vault" "kv_prvt" {
+  count               = (local.enable_deployers && local.prvt_kv_exist) ? 1 : 0
+  name                = local.prvt_kv_name
+  resource_group_name = local.prvt_kv_rg_name
+}
+
 resource "azurerm_key_vault_access_policy" "kv_prvt_msi" {
-  count        = local.enable_deployers ? 1 : 0
+  count        = (local.enable_deployers && ! local.prvt_kv_exist) ? 1 : 0
   key_vault_id = azurerm_key_vault.kv_prvt[0].id
 
   tenant_id = data.azurerm_client_config.deployer.tenant_id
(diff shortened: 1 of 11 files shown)


deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf
```diff
@@ -139,4 +138,22 @@ locals {
...
+  ppk_name = format("%s-sshkey", local.prefix)
+  pk_name  = format("%s-sshkey-pub", local.prefix)
+  pwd_name = format("%s-password", local.prefix)
+
+  // Extract information from the specified key vault arm ids
+  user_kv_name    = local.user_kv_exist ? split("/", local.user_key_vault_id)[8] : ""
+  user_kv_rg_name = local.user_kv_exist ? split("/", local.user_key_vault_id)[4] : ""
+
+  prvt_kv_name    = local.prvt_kv_exist ? split("/", local.prvt_key_vault_id)[8] : ""
+  prvt_kv_rg_name = local.prvt_kv_exist ? split("/", local.prvt_key_vault_id)[4] : ""
```
rukawata, line 157: For those vars, it is not clear that this is only used for existing KVs.
So can we consolidate the value with naming convention, then replace the variable used in new resource name as well? 

eg:
user_kv_name    = local.user_kv_exist ? split("/", local.user_key_vault_id)[8] : local.keyvault_names.user_access

Then in new KV resource [here](https://github.com/Azure/sap-hana/blob/d97be222b6ed0d1afbef21324abfe48581a26f64/deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf#L39), directly use `user_kv_name`.

deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
```diff
@@ -141,8 +157,26 @@ resource "random_password" "deployer" {
...
-  count        = (local.enable_deployers && local.enable_password) ? 1 : 0
+  count        = (local.enable_deployers && local.enable_password && ! local.user_kv_exist) ? 1 : 0
   name         = format("%s-password", local.prefix)
   value        = local.password
   key_vault_id = azurerm_key_vault.kv_user[0].id
 }
+
+data "azurerm_key_vault_secret" "pk" {
+  count        = (local.enable_deployers && local.enable_key && local.user_kv_exist) ? 1 : 0
+  name         = format("%s-sshkey-pub", local.prefix)
```
rukawata, line 168: if we have a var defined for pk/ppk name already `local.ppk_name`, can we just use that? Same applies to the resource type block :)

rukawata (commented):
Overall looks fine :)
Some small suggestions - please see comments.

deploy/terraform/terraform-units/modules/sap_deployer/variables_local.tf
YunziZhangAzure: agree! kv_name updated.
But to keep the implicit dependency between kv and rg, the resource group name can't be replaced by user_kv_rg_name.

deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
YunziZhangAzure: Updated!

rukawata (approved):
LGTM!

PR #940, opened 2020-11-20
Removing invalid char in KV names

KimForss:
## Problem
sap_library was not using the naming module for deployer naming

## Solution
update the code to use the naming module

## Tests
normal eployment should suffice

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_namegenerator/keyvault.tf
+++ b/deploy/terraform/terraform-units/modules/sap_namegenerator/keyvault.tf
@@ -1,15 +1,15 @@
 locals {
 
-  sdu_private_keyvault_name = format("%s%s%s%sp%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid, local.random_id_verified)
-  sdu_user_keyvault_name    = format("%s%s%s%su%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid, local.random_id_verified)
+  sdu_private_keyvault_name = format("%s%s%s%sp%s", replace(local.env_verified, "/[^A-Za-z0-9]/", ""), local.location_short, replace(local.vnet_verified, "/[^A-Za-z0-9]/", ""), var.sap_sid, local.random_id_verified)
+  sdu_user_keyvault_name    = format("%s%s%s%su%s", replace(local.env_verified, "/[^A-Za-z0-9]/", ""), local.location_short, replace(local.vnet_verified, "/[^A-Za-z0-9]/", ""), var.sap_sid, local.random_id_verified)
 
-  deployer_private_keyvault_name = format("%s%s%sprvt%s", local.env_verified, local.location_short, local.dep_vnet_verified, local.random_id_verified)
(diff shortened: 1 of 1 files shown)


deploy/terraform/bootstrap/sap_library/variables_local.tf
```diff
@@ -44,11 +6,10 @@ locals {
   // Derive resource group name for deployer
   deployer                = try(var.deployer, {})
   deployer_environment    = try(local.deployer.environment, "")
-  deployer_location_short = try(var.region_mapping[local.deployer.region], "unkn")
   deployer_vnet           = try(local.deployer.vnet, "")
-  deployer_prefix         = upper(format("%s-%s-%s", local.deployer_environment, local.deployer_location_short, substr(local.deployer_vnet, 0, 7)))
+  deployer_prefix         = module.sap_namegenerator.naming.prefix.DEPLOYER
   // If custom names are used for deployer, providing resource_group_name and msi_name will override the naming convention
-  deployer_rg_name = try(local.deployer.resource_group_name, format("%s-INFRASTRUCTURE", local.deployer_prefix))
+  deployer_rg_name = try(local.deployer.resource_group_name, format("%s%s", local.deployer_prefix,module.sap_namegenerator.naming.resource_suffixes.deployer_rg))
```
rukawata, line 12: This name may be incorrect, hence cause failure in pipeline where it says the deployer tfstate can not be found ([link](https://dev.azure.com/azuresaphana/Azure-SAP-HANA/_build/results?buildId=15713&view=logs&j=d764d31b-fb26-5fc3-93b2-7fc094a2e94a&t=082f2e6a-4a77-542e-5616-b3548fbbcdf9))
Please check. Thanks!

deploy/terraform/terraform-units/modules/sap_namegenerator/resourcegroup.tf
```diff
@@ -4,15 +4,15 @@ locals {
     upper(format("%s-%s-%s_%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.codename, var.sap_sid))) : (
     upper(format("%s-%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified, var.sap_sid))
   )
-  deployer_name = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.dep_vnet_verified))
-  vnet_name     = upper(format("%s-%s-%s", local.env_verified, local.location_short, local.vnet_verified))
-  library_name  = upper(format("%s-%s", local.env_verified, local.location_short))
+  deployer_name = upper(format("%s-%s-%s", local.deployer_env_verified, local.location_short, local.dep_vnet_verified))
```
rukawata, line 7: Now that `deployer_name` uses deployer_env_verified, we should also change in `bootstrap/sap_deployer/module.tf,` so it passes `deployer_environment` instead of `environment`. Otherwise the code is confusing.

The same for run.

Also, we what is passed is `management_vnet_name`, it should be used not `dep_vnet_verified`.

I think there may still something require more thinking.

I would suggest the following:
1. for this PR, leave the naming for deployer_rg_name in sap_library.
1. a new PR that change the logic in naming module to takes separate env, loc, vnet info for deployer/saplib/landscape/sid
1. a third PR, change the naming for deployer_rg_name in sap_library.

rukawata (approved):

PR #947, opened 2020-11-22
[TAGGING] Support custom tagging for database VMs

KimForss:
## Problem
Customers want the ability to apply tags on the Virtual machines at creation time
## Solution
Add a property in the application node in the json that supports providing tags for the database tier VMs

## Tests
Add the following snippet to the input json in the database section
 "tags" :{
        "tagname": "Value",
        "tagname2": "Value"
}
## Notes


deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -121,6 +121,11 @@ locals {
 
   db_systemdb_password = "db_systemdb_password"
 
+  ///////////////////////////////////////
+  // Tags
+  tags = try(local.anydb.tags,[])
+  ////////////////////////////////////////
```
rukawata, line 127: The first comment line and the last are not the same length...anyway, let's not add those blocks until we have a defined way to write comment blocks other than using /**/ :)

rukawata: I think tags is a map not list. So default should be `{}`.

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -87,6 +87,12 @@ locals {
   availabilityset_arm_ids = try(local.hdb.avset_arm_ids, [])
   availabilitysets_exist  = length(local.availabilityset_arm_ids) > 0 ? true : false
 
+  ///////////////////////////////////////
+  // Tags
+  tags = try(local.hdb.tags,[])
+  ////////////////////////////////////////
```
rukawata, line 93: Same as above.

rukawata (commented):
Some small comments, please check - thanks!

rukawata (approved):
LGTM!

PR #952, opened 2020-11-23
when iscsi is not provisioned, the output of iscsi should be empty string

YunziZhangAzure:
## Problem
when iscsi is not provisioned, the output of iscsi should be empty string

## Solution
Add a condition to determine if there's iscsi to be provisioned.

## Tests

## Notes


rukawata (approved):
LGTM!

PR #953, opened 2020-11-23
Bring secrets in existing kv for deployer

YunziZhangAzure:
## Problem
Based on the discussion about the scope of "importing existing KV", the logic needs to be updated.

## Solution
Use case 1:
The user brings a KV but doesn't bring secrets. Automation will generate secret, i.e. key pair or password, in the existing KV.

Use case 2:
The user brings a KV and the secret name. Automation will not generate secrets in the existing KV, only import them to deploy deployers.

## Tests
Test locally

## Notes
Based on discussion, deployer-msi will be given permission to access secrets in deployer KV manually by users, out of Automation.


deploy/terraform/bootstrap/sap_deployer/module.tf
```diff
@@ -4,15 +4,18 @@ Description:
...
+  infrastructure          = var.infrastructure
+  deployers               = var.deployers
+  options                 = var.options
+  ssh-timeout             = var.ssh-timeout
+  sshkey                  = var.sshkey
+  user_key_vault_id       = var.user_key_vault_id
+  private_key_vault_id    = var.private_key_vault_id
+  public_key_secret_name  = var.public_key_secret_name
+  private_key_secret_name = var.private_key_secret_name
+  password_secret_name    = var.password_secret_name
```
rukawata, line 17: now that we have several inputs that are related to existing key vault, does it make sense to pass in all of them as a dictionary as `infrastructure`, `deployers`, etc...?

rukawata (commented):
Overall looks good! Just one thought, let me know what you think :)

rukawata (approved):
LGTM! Thanks for addressing the comment :)

PR #955, opened 2020-11-24
import existing kv to saplibrary

YunziZhangAzure:
## Problem
Based on the discussion, users should be able to bring their own Azure KVs.

## Solution
User can specify the arm id KVs, then the KV will be imported.

## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_library/output.tf
+++ b/deploy/terraform/bootstrap/sap_library/output.tf
@@ -26,9 +26,17 @@ output "fileshare_sapbits_name" {
   value     = module.sap_library.fileshare_sapbits_name
 }
 
-output "user_vault_name" {
+output "library_user_key_vault_name" {
   sensitive = true
-  value     = module.sap_library.user_vault_name
+  value     = module.sap_library.library_user_key_vault_name
+}
+
+output "library_user_key_vault_id" {
+  value = module.sap_library.library_user_key_vault_id
+}
+
+output "library_private_key_vault_id" {
+  value = module.sap_library.library_private_key_vault_id
 }
 
 output "remote_state_resource_group_name" {
--- a/deploy/terraform/terraform-units/modules/sap_library/output.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/output.tf
@@ -22,13 +22,24 @@ output "fileshare_sapbits_name" {
   value = local.fileshare_sapbits_name
 }
 
-
 output "random_id" {
   value = random_id.post_fix.hex
 }
 
-output "user_vault_name" {
-  value = azurerm_key_vault.kv_user.name
+output "library_user_key_vault_name" {
+  value = local.user_kv_exist ? local.user_kv_name : azurerm_key_vault.kv_user[0].name
+}
+
+output "library_user_key_vault_id" {
+  value = local.user_kv_exist ? local.user_key_vault_id : azurerm_key_vault.kv_user[0].id
+}
+
+output "library_private_key_vault_name" {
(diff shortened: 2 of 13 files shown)


deploy/terraform/bootstrap/sap_library/output.tf
```diff
@@ -31,6 +31,14 @@ output "user_vault_name" {
   value     = module.sap_library.user_vault_name
 }
 
+output "library_user_key_vault_id" {
+  value = module.sap_library.library_user_key_vault_id
+}
+
+output "library_private_key_vault_id" {
+  value = module.sap_library.library_private_key_vault_id
+}
```
rukawata, line 40: What are those outputs for?

deploy/terraform/terraform-units/modules/sap_library/output.tf
```diff
@@ -22,13 +22,24 @@ output "fileshare_sapbits_name" {
...
+  value = local.user_kv_exist ? local.user_kv_name : azurerm_key_vault.kv_user[0].name
+}
+
+output "library_user_key_vault_id" {
+  value = local.user_kv_exist ? local.user_key_vault_id : azurerm_key_vault.kv_user[0].id
+}
+
+output "library_private_key_vault_name" {
+  value = local.prvt_kv_exist ? local.prvt_kv_name : azurerm_key_vault.kv_prvt[0].name
+}
```
rukawata, line 39: What is the reason of adding this output? I do not see this output in the root module for both run and bootstrap though.
And can we make the output variable name consistent with `user_vault_name`? Or change the `user_vault_name` to `library_user_key_vault_name` to be consistent with this one.

rukawata (commented):
Overall looks good. Just some question regarding outputs. Thanks!

deploy/terraform/bootstrap/sap_library/output.tf
YunziZhangAzure: Since now we support importing KVs and the KVs imported may not be in the same resource group as the library's, so we need to store the arm id of KVs and use them to locate the KVs when retrieving the information of KVs, e.g: username/password of downloader.

deploy/terraform/terraform-units/modules/sap_library/output.tf
YunziZhangAzure: I'm adding this for future use. Currently, all secrets are stored in user kv, there's no need to know which is the private kv right now, but it seems we have a potential use case to store the secrets in private kv in the future, I thought it would be good to have this output about private key vault in the child module now and when we are ready to switch to private KV, we can just update the output of root module.

For the name, I can change user_vault_name to the library_user_key_vault_name if you agree.

rukawata: That sounds good. Thanks!

deploy/terraform/bootstrap/sap_library/output.tf
```diff
@@ -26,11 +26,19 @@ output "fileshare_sapbits_name" {
   value     = module.sap_library.fileshare_sapbits_name
 }
 
-output "user_vault_name" {
+output "library_user_key_vault_name" {
   sensitive = true
   value     = module.sap_library.user_vault_name
 }
```
rukawata, line 32: The value needs to be changed accordingly :)
```suggestion
output "library_user_key_vault_name" {
  sensitive = true
  value     = module.sap_library. library_user_key_vault_name
}
```

YunziZhangAzure: Good catch! Thanks! Updated.

rukawata (approved):

PR #956, opened 2020-11-24
fix the secret name of deployer's output 

YunziZhangAzure:
## Problem
The output name of pub/prvt secrets of deployer'kv is wrong

## Solution
Fix the bug.

## Tests

## Notes


rukawata (approved):
Thanks for fixing that!

PR #963, opened 2020-11-30
[NO DEPLOYER VM] Decouple SAP Library from SAP Deployer

KimForss:
## Problem
If the customer provides a key vault the code should use that. This would remove the dependency on the deployer state file and would allow a deployment with out it

## Solution
deployer_key_vault_arm_id = try(var.key_vault.kv_user_id, try(data.terraform_remote_state.deployer[0].outputs.deployer_kv_user_arm_id, ""))

## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_library/imports.tf
+++ b/deploy/terraform/bootstrap/sap_library/imports.tf
@@ -4,6 +4,7 @@
 */
 
 data "terraform_remote_state" "deployer" {
+  count = local.use_deployer ? 1 : 0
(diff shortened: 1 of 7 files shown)


rukawata (commented):
Thanks for working on this :) 
However, this falls into the feature request to deploy without deployer, and also is part of TFC/TFE support. Therefore I would not review this PR until we have a proper design discussion on this (the behavior needs to be consistent with all modules, and Morgan mentioned also to restructure the bootstrap/run folder).

PR #965, opened 2020-11-30
import username secret into landscape

YunziZhangAzure:
## Problem
User should be able to override the username of VM by passing the name of secret stored username.

## Solution
If customers specify "kv_iscsi_username" and authentication type of iscsi is "password", the secret will be imported and its value will be used as the username in logon to VM.

## Tests
Test locally

## Notes


rukawata (approved):
Thanks!

PR #967, opened 2020-11-30
use import username as logon username

YunziZhangAzure:
## Problem
If the user specify the name of secret which is used to store username, the imported username should be used as the logon username.

## Solution

## Tests
Tests locally.

## Notes


rukawata (approved):
Thanks for fixing that!

PR #968, opened 2020-11-30
[INFRASTRUCTURE] The fault domain count is different per region - application tier fixes

KimForss:
## Problem
Deploy the correct number of fault domains for the application tier

## Solution
Persist the fault domain count in a json file and use it for the count

## Tests
Deploy a new environment

## Notes
`    az vm list-skus --resource-type availabilitySets --query '[?name==`Aligned`].{​Location:locationInfo[0].location, MaximumFaultDomainCount:capabilities[0].value}​' -o Table`

avset and VMs are recreated after fault domain count changes: file bug with azurerm [#9577](https://github.com/terraform-providers/terraform-provider-azurerm/issues/9577)


deploy/configs/faults.json
```diff
@@ -0,0 +1,174 @@
+[
```
rukawata, line 1: Can we name the file so other people can better understand? eg. max_fault_domain_count.json

deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
```diff
@@ -64,6 +65,12 @@ locals {
   resource_suffixes = var.naming.resource_suffixes
 
   region  = try(var.infrastructure.region, "")
+
+  faultdomain_count = try(tonumber(compact(
+    [for pair in local.faults : 
+      pair.Location == local.region ? pair.MaximumFaultDomainCount : ""
```
rukawata, line 71: Region may spell with uppercase/lowercase/ or mix...so I would compare with both uppercase/lowercase.

deploy/terraform/terraform-units/modules/sap_system/app_tier/infrastructure.tf
```diff
@@ -126,7 +126,7 @@ resource "azurerm_availability_set" "scs" {
   location                     = var.resource_group[0].location
   resource_group_name          = var.resource_group[0].name
   platform_update_domain_count = 20
-  platform_fault_domain_count  = 2
+  platform_fault_domain_count  = local.faultdomain_count
```
rukawata, line 129: SCS should be 2, not the max.

rukawata (commented):
Thanks for adding this. See some small comments.

rukawata:
Another thing is I noticed that by changing this number, avset got re-created, hence VMs got re-created.
I'm not sure this is the expected behavior, opening an issue with terraform.

deploy/configs/faults.json
KimForss: done

deploy/terraform/terraform-units/modules/sap_system/app_tier/infrastructure.tf
KimForss: For instance Woolworths have a standby SCS - making the count 3 - so I would keep the max value here

rukawata (approved):

rukawata (approved):

PR #969, opened 2020-12-01
Add pre-conditions for importing username

YunziZhangAzure:
## Problem

## Solution
Whether authentication type is ssh key or password, username will be stored in KV anyway 
 
## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/deployer_full.json
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer_full.json
@@ -21,6 +21,7 @@
         "kv_prvt_id": "",
         "kv_sshkey_prvt" : "",
         "kv_sshkey_pub" : "",
+        "kv_username" : "",
         "kv_pwd": ""
     }
 }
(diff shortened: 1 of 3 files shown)


rukawata (commented):
The code looks fine :)
My only question is, why not import username when key is used? I think logon with key also would require username.

YunziZhangAzure:
> The code looks fine :)
> My only question is, why not import username when key is used? I think logon with key also would require username.

I'm ok with either of them. 

My initial thought is, currently when authentication type is key, we don't store username in KV, then in parallel, so username should not store the username in KV when the user tries to specify the username, they should use authentication.username = "XXX" in input json instead.

Plus, our logic for importing is as follows:
count        = (local.enable_landscape_kv && local.enable_iscsi_auth_password && local.iscsi_username_exist) ? 1 : 0

https://github.com/Azure/sap-hana/blob/62c4b52668aef4e28d9e0f5a61e81bed12cdd4fb/deploy/terraform/terraform-units/modules/sap_landscape/key_vault_sap_landscape.tf#L142

If kv_username is specified in input json but local.enable_iscsi_auth_password is false, the data won't be imported which leads to the error when executing the following command:

username = local.enable_deployers ? (local.username_exist ? data.azurerm_key_vault_secret.username[0].value : try(local.deployer_input[0].authentication.username, "azureadm")): "" 

so I'm thinking to make them consistent. We can choose to modify the condition when importing data username, or change like what I did in this pr.

rukawata (approved):

PR #976, opened 2020-12-01
[POC]support tfvars and json. [Don't merge]

YunziZhangAzure:
## Problem

## Solution

## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/variables_local.tf
+++ b/deploy/terraform/bootstrap/sap_deployer/variables_local.tf
@@ -1,22 +1,22 @@
 locals {
 
-  environment = lower(try(var.infrastructure.environment, ""))
-  location    = try(var.infrastructure.region, "")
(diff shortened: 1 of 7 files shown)


deploy/terraform/bootstrap/sap_deployer/variables_local.tf
```diff
@@ -1,11 +1,11 @@
 locals {
 
-  environment = lower(try(var.infrastructure.environment, ""))
-  location    = try(var.infrastructure.region, "")
-  codename    = lower(try(var.infrastructure.codename, ""))
+  environment = lower(try(local.infrastructure.environment, ""))
```
KimForss, line 3: WE should not have try statements around environment and region as both are mandatory dields

rukawata: It will be mandatory since in `deploy/terraform/bootstrap/sap_deployer/tfvar_variables.tf` the default value is removed.

PR #979, opened 2020-12-02
Fix the broken lookups when size does not exist for that database type

KimForss:
## Problem
The db size lookup fails for keys that do not exist in both the hana and anydb json files

## Solution
Add logic that will only perform the lookup if  deployed

## Tests
Deploy using a custom size json with a custom size name map

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
@@ -281,9 +281,11 @@ locals {
     }
   ])
 
+  db_sizing = local.enable_deployment ? try(lookup(local.sizes, local.anydb_size).storage, []) : []
+
   data_disk_per_dbnode = (length(local.anydb_vms) > 0) ? flatten(
     [
-      for storage_type in lookup(local.sizes, local.anydb_size).storage : [
+      for storage_type in local.db_sizing : [
         for disk_count in range(storage_type.count) : {
           suffix               = format("%s%02d", storage_type.name, disk_count)
           storage_account_type = storage_type.disk_type,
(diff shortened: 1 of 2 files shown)


deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -281,9 +281,11 @@ locals {
     }
   ])
 
+  db_sizing = local.enable_deployment ? try(lookup(local.sizes, local.anydb_size).storage, []) : []
```
rukawata, line 284: 1. I think with the try logic, you no longer need the enable_deployment condition.
2. Can we change this to use default in lookup function?
eg:
```
anydb_size_details = lookup(local.sizes, local.anydb_size, [])
db_sizing = local.anydb_size_details != [] ? local.anydb_size_details.storage : []
```

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -315,11 +317,16 @@ locals {
     ]
   ])
 
-  storage_list = lookup(local.sizes, local.anydb_size).storage
-  enable_ultradisk = try(compact([
-    for storage in local.storage_list :
-    storage.disk_type == "UltraSSD_LRS" ? true : ""
-  ])[0], false)
+  enable_ultradisk = local.enable_deployment ? (
```
rukawata, line 320: same as above, the exception has been handled with db_sizing, not sure if we still need enable_deployment.

rukawata (commented):
Thanks for fixing this! Some suggestions and let me know your thoughts.

rukawata (approved):

PR #983, opened 2020-12-02
[INFRASTRUCTURE]Always use standard load balancer for Hana

KimForss:
## Problem
Hana deployments should always use Standard load balancer

## Solution

## Tests

## Notes


rukawata (approved):
Thanks for fixing this!

PR #991, opened 2020-12-03
Provide the option to use a non zero based index

KimForss:
## Problem
Some customers want to have their resource naming using a zero based index and some want a one based index

## Solution
Provide the offset used in resource naming as part of the naming module

## Tests

## Notes


deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -290,5 +290,10 @@ locals {
   //The separator to use between the prefix and resource name
   separator = "_"
 
+  //The offset used in resource naming, default is 0
+  // 0 -> disk0, disk1
+  // 1 -> disk1, disk2
+  offset = 0
```
rukawata, line 296: Do you want to make offset a possiblilty from input? This way no need to change the module file.
For example, under options, you can have a key called disk_offset.

KimForss: I like that approach - but we can do that later

KimForss: Actually it was easier than I thought - change is done

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -88,6 +88,11 @@ variable deployer_vm_count {
   default = 1
 }
 
+variable resource_offset {
+  type    = number
+  default = 1
```
rukawata, line 93: default should be 0?

deploy/terraform/terraform-units/modules/sap_system/app_tier/variables_local.tf
```diff
@@ -66,6 +66,9 @@ locals {
 
   region  = try(var.infrastructure.region, "")
 
+    //Allowing changing the base for indexing, default is zero-based indexing, if customers want the first disk to start with 1 they would change this
```
rukawata, line 69: nit:
```suggestion
  //Allowing changing the base for indexing, default is zero-based indexing, if customers want the first disk to start with 1 they would change this
```

deploy/terraform/terraform-units/modules/sap_system/hdb_node/variables_local.tf
```diff
@@ -70,6 +70,9 @@ locals {
 
   rg_name = try(var.infrastructure.resource_group.name, format("%s%s", local.prefix, local.resource_suffixes.sdu_rg))
 
+    //Allowing changing the base for indexing, default is zero-based indexing, if customers want the first disk to start with 1 they would change this
```
rukawata, line 73: nit:
```suggestion
  //Allowing changing the base for indexing, default is zero-based indexing, if customers want the first disk to start with 1 they would change this
```

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
KimForss: argh - yes :)

rukawata (approved):

PR #998, opened 2020-12-07
Enhance input json files

YunziZhangAzure:
## Problem
Current input json is not complete

## Solution
1. Add some missing fields in input json
2. Since name of vnet is required, test pipeline will use jq to populate this field dynamically.

## Tests
See test in test pipeline

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_deployer/deployer.json
+++ b/deploy/terraform/bootstrap/sap_deployer/deployer.json
@@ -4,6 +4,7 @@
     "environment": "np",
     "vnets": {
       "management": {
+        "name": "NP-EAUS-DEP00-vnet",
         "address_space": "10.0.0.0/26",
         "subnet_mgmt": {
           "prefix": "10.0.0.16/28"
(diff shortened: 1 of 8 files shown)


deploy/terraform/bootstrap/sap_deployer/deployer.json
```diff
@@ -4,6 +4,7 @@
     "environment": "np",
     "vnets": {
       "management": {
+        "name": "NP-EAUS-DEP00-vnet",
```
rukawata, line 7: vnet name would follow naming convention, it makes sense to have it in deployer_full.json but not the deployer.json. Any specific reasons?

rukawata (commented):
Thanks for adding the missing parameters :) Just question regarding one input param.

deploy/terraform/bootstrap/sap_deployer/deployer.json
YunziZhangAzure: It seems this field is required according to the configuration doc from PM.
https://github.com/Azure/sap-hana/blob/documentation/v2.3/documentation/SAP_Automation_on_Azure/Software_Documentation/configuration-deployer.md

rukawata: I see! Thanks for going through the documentation.
I think this is incorrect in the documentation, naming convention should be used when possible :)

PR #1000, opened 2020-12-08
Correct the VM naming for scaleout

KimForss:
## Problem
The current logic causes issues when deploying across multiple zones. In the current logic the server names are defined in two separate lists which get concatenated. If we have 3 servers then the HA name list will be

S1,S2,S3, S1HA, S2HS,S3HA. If we want to deploy across two Availability zones our code will put the first VM in the first zone, the second in the second and the end result will be that
Zone 1 have the servers S1, S3,S2HA and Zone 2 will have S2, S1HA, S3HA

## Solution
Instead of concatenating the names and the ha names the names need to be intertwined so that the example list from above will be S1, S1HA, S2, S2HA, S3, S3HA

## Tests

## Notes


YunziZhangAzure (commented):
Like what we said, please add more description about the problem and the failing case.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -217,25 +217,33 @@ locals {
...
-    ]
-  )
 
+  dbnodes = local.anydb_ha ? (
+    flatten([for idx, dbnode in try(local.anydb.dbnodes, [{}]) :
+      [
+        {
+          name           = try("${dbnode.name}-0", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx], local.resource_suffixes.vm))
+          computername   = try("${dbnode.name}-0", local.computer_names[idx], local.resource_suffixes.vm)
+          admin_nic_ip   = lookup(dbnode, "admin_nic_ips", [false, false])[0]
```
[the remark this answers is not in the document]
KimForss: role is a Hana feature - and that is why it is not herer

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
```diff
@@ -26,14 +26,14 @@ locals {
...
+      format("%sd%s%sz%s%s%02dl%d%s", lower(var.sap_sid), lower(var.db_sid), local.separator, length(var.db_zones) == 2 ? var.db_zones[0] : var.db_zones[idx % max(length(var.db_zones), 1)], local.separator, idx + var.resource_offset, 0, local.random_id_vm_verified)) : (
       format("%sd%s%02dl%d%s", lower(var.sap_sid), lower(var.db_sid), idx + var.resource_offset, 0, local.random_id_vm_verified)
     )
   ]
 
   anydb_vm_names_ha = [for idx in range(var.db_server_count) :
     length(var.db_zones) > 0 ? (
-      format("%sd%s%sz%s%s%02dl%d%s", lower(var.sap_sid), lower(var.db_sid), local.separator, var.db_zones[idx % max(length(var.db_zones), 1)], local.separator, idx + var.resource_offset, 1, local.random_id_vm_verified)) : (
+      format("%sd%s%sz%s%s%02dl%d%s", lower(var.sap_sid), lower(var.db_sid), local.separator, length(var.db_zones) == 2 ? var.db_zones[1] : var.db_zones[idx % max(length(var.db_zones), 1)], local.separator, idx + var.resource_offset, 1, local.random_id_vm_verified)) : (
```
YunziZhangAzure, line 36: I'm a little confused here:
If length(db_azones) = 2, var.db_zones[idx % max(length(var.db_zones), 1)] = var.db_zones[1 % 2] = var.db_zones[1]

so why do we need to have a separate condition? length(var.db_zones) == 2 ? var.db_zones[0]

It seems to me that the original code has same logic as your change.

YunziZhangAzure:
Btw, I have a question. I thought our HA scenario contains only 2 servers, why we do we have 6 servers in your case?

KimForss:
Scaleout scenario

KimForss:
The reason is to make sure that all HA servers go to var.db_zones[1] is zone count is 2.

[idx %2] would put the first in [0] the second in [1] the third in [0] etc

deploy/terraform/terraform-units/modules/sap_namegenerator/vm.tf
KimForss: Scaleout scenario

YunziZhangAzure:
Do we think of "Zone 1 have the servers S1, S3,S2HA and Zone 2 will have S2, S1HA, S3HA" as a failure. I'm thinking it is functioning correct. 
PR 1000 is to deal with 2 zones, what's going to happen when zone = 3? Does it also have an ideal output naming?

As such, I don't think this enhancement has to be done.

KimForss:
> Do we think of "Zone 1 have the servers S1, S3,S2HA and Zone 2 will have S2, S1HA, S3HA" as a failure. I'm thinking it is functioning correct.
> PR 1000 is to deal with 2 zones, what's going to happen when zone = 3? Does it also have an ideal output naming?
> 
> As such, I don't think this enhancement has to be done.

It has to be done - as it is not functioning correctly now. The customer needs the ability to specify the names to be consistent inside of a zone. and HA for databases would never cross 3 zones as none of the SAP databases have multi HA capability - it is always a HA pair

KimForss:
The rebase broke it

PR #1003, opened 2020-12-09
 [ENVIRONMENT] Support providing the deployer environment as a parameter

KimForss:
## Problem
The are cases when the deployer is in a different environment from the sap library

## Solution
Add the support for multiple environment in the naming module, i.e deployer environment and sdu environment.

## Tests

## Notes

### DIFF
--- a/deploy/terraform/bootstrap/sap_library/module.tf
+++ b/deploy/terraform/bootstrap/sap_library/module.tf
@@ -16,8 +16,10 @@ module "sap_library" {
 }
 
 module "sap_namegenerator" {
-  source      = "../../terraform-units/modules/sap_namegenerator"
-  environment = lower(try(var.infrastructure.environment, ""))
-  location    = try(var.infrastructure.region, "")
-  random_id   = module.sap_library.random_id
+  source                = "../../terraform-units/modules/sap_namegenerator"
+  environment           = lower(try(var.infrastructure.environment, ""))
+  deployer_environment  = local.deployer_environment
+  management_vnet_name  = local.deployer_vnet
+  location              = try(var.infrastructure.region, "")
+  random_id             = module.sap_library.random_id
 }
--- a/deploy/terraform/bootstrap/sap_library/module.tf
+++ b/deploy/terraform/bootstrap/sap_library/module.tf
@@ -17,9 +17,10 @@ module "sap_library" {
 
 module "sap_namegenerator" {
   source                = "../../terraform-units/modules/sap_namegenerator"
-  environment           = lower(try(var.infrastructure.environment, ""))
-  deployer_environment  = local.deployer_environment
-  management_vnet_name  = local.deployer_vnet
-  location              = try(var.infrastructure.region, "")
+  environment           = var.infrastructure.environment
+  deployer_environment  = var.deployer.environment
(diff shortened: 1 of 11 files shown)


deploy/terraform/bootstrap/sap_library/module.tf
```diff
@@ -16,8 +16,10 @@ module "sap_library" {
...
 module "sap_namegenerator" {
-  source      = "../../terraform-units/modules/sap_namegenerator"
-  environment = lower(try(var.infrastructure.environment, ""))
-  location    = try(var.infrastructure.region, "")
-  random_id   = module.sap_library.random_id
+  source                = "../../terraform-units/modules/sap_namegenerator"
+  environment           = lower(try(var.infrastructure.environment, ""))
+  deployer_environment  = local.deployer_environment
+  management_vnet_name  = local.deployer_vnet
+  location              = try(var.infrastructure.region, "")
```
YunziZhangAzure, line 23: It seems the location of deployer is not passed to the naming module. If the deployer's region is different from saplibrary's, the name of deployer rg name could be wrong?

KimForss: It could but having deployer in another region would be bad because of the network latency between the regions and we should not support that in my opinion

deploy/terraform/bootstrap/sap_library/module.tf
```diff
@@ -16,8 +16,11 @@ module "sap_library" {
 }
 
 module "sap_namegenerator" {
-  source      = "../../terraform-units/modules/sap_namegenerator"
-  environment = lower(try(var.infrastructure.environment, ""))
-  location    = try(var.infrastructure.region, "")
-  random_id   = module.sap_library.random_id
+  source                = "../../terraform-units/modules/sap_namegenerator"
+  environment           = var.infrastructure.environment
+  deployer_environment  = var.deployer.environment
```
YunziZhangAzure, line 21: Since the logic to be implemented is "If the environment of deployer is specified, it will be used, otherwise infrastructure.environment will be used"

I think the following code is better:
 deployer_environment  = try(var.deployer.environment, "")

Same for location:

 deployer_location = try(var.deployer.region, "")

deploy/terraform/run/sap_library/variables_local.tf
```diff
@@ -1,59 +1,14 @@
...
-}
 
 variable "tfstate_resource_id" {
   description = "The resource id of tfstate storage account"
   default     = ""
 }
 
 locals {
-  // Sap library's environment
-  environment = upper(try(var.infrastructure.environment, ""))
```
YunziZhangAzure, line 47: The environment field shouldn't be removed. It's being used when retrieving SPN. Please see the error in the test pipeline

deploy/terraform/bootstrap/sap_library/module.tf
KimForss: As we chatted - I added that support

deploy/terraform/run/sap_library/variables_local.tf
KimForss: I changed the code to use the var.infrastructure.environment variable -

YunziZhangAzure (approved):
LGTM

PR #1004, opened 2020-12-10
Add an indicator about enabling iscsi

YunziZhangAzure:
## Problem
1. When the iscsi is not provisioned, the username of iscsi shouldn't be stored in the KV of sap landscape.

2. When sub_iscsi is specified in input json, even if iscsi_count = 0, subnet of iscsi is supposed to be provisioned or imported. 
When sub_iscsi is specified in input json, iscsi_nsg should be created or imported.

## Solution
Add an indicator enable_iscsi
Add indicator enable_sub_iscsi

## Tests
see test pipeline

## Notes


rukawata (approved):
LGTM!

PR #1005, opened 2020-12-10
[INFRASTRUCTURE] Update LoadBalance types to standard for all tiers

KimForss:
## Problem
SAP should always use the Standard Load Balancer.

From documentation:
We recommend using Azure Standard Load Balancer for all SAP scenarios. If virtual machines in the backend pool require public outbound connectivity, or if they are used in an Azure zone deployment, the standard load balancers require additional configurations.

https://docs.microsoft.com/en-us/azure/architecture/reference-architectures/sap/sap-netweaver 

## Solution
Change the SKU to Standard

## Tests

## Notes
The PR will change the Load Balancer therefore the pipeline might fail


YunziZhangAzure (approved):
LGTM!

PR #1013, opened 2020-12-14
[Beta2.3]refactor the input

YunziZhangAzure:
## Problem
template_samples folder should be removed.
"ansible_execution": true should be removed

## Solution

## Tests

## Notes


deploy/input.json
```diff
@@ -267,7 +267,6 @@
   },
   "options": {
     "enable_secure_transfer": true,
-    "ansible_execution": true,
```
rukawata, line 270: This together with the json under templates/tempGen/* needs to have this field still since they serve the current master not the feature branch.

YunziZhangAzure: I'm a little confused here. 

The master branch does need ansible_execution, but this pr is to remove this field in feature/remote-tfstate2, then the change will be merged to beta/v2.3. Will this impact current master branch?

KimForss:
Isn't the plan to replace the master branch with the new codebase?

deploy/input.json
rukawata: Sorry it was indeed confusing. My point was, those json are used by current master but not beta branches, hence we do not need to modify those files, instead if at all, they should be deleted (also the pipeline that uses those jsons should be removed, once master got updated with beta).

rukawata:
> I'm a little confused here. The master branch does need ansible_execution, but this pr is to remove this field in feature/remote-tfstate2, then the change will be merged to beta/v2.3. Will this impact current master branch? 
It would not impact the current master since the change stays in tfstate2, later beta, until we merge beta to master :)
The pipeline picks only the pipeline and corresponding files from the branch it was defined - which is the master.

deploy/input.json
YunziZhangAzure: Got it! Will remove related files.

rukawata (approved):

PR #1015, opened 2020-12-17
[Beta2.3]Scaleout naming

KimForss:
## Problem
The current logic causes issues when deploying across multiple zones. In the current logic the server names are defined in two separate lists which get concatenated. If we have 3 servers then the HA name list will be

S1,S2,S3, S1HA, S2HS,S3HA. If we want to deploy across two Availability zones our code will put the first VM in the first zone, the second in the second and the end result will be that
Zone 1 have the servers S1, S3,S2HA and Zone 2 will have S2, S1HA, S3HA

## Solution
Instead of concatenating the names and the ha names the names need to be intertwined so that the example list from above will be S1, S1HA, S2, S2HA, S3, S3HA

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
@@ -224,12 +224,14 @@ dbnodes = local.anydb_ha ? (
         {
           name           = try("${dbnode.name}-0", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx], local.resource_suffixes.vm))
           computername   = try("${dbnode.name}-0", local.computer_names[idx], local.resource_suffixes.vm)
+          role           = try(dbnode.role, "db")
           admin_nic_ip   = lookup(dbnode, "admin_nic_ips", ["false", "false"])[0]
           db_nic_ip      = lookup(dbnode, "db_nic_ips", ["false", "false"])[0]
         },
         {
           name           = try("${dbnode.name}-1", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx + local.node_count], local.resource_suffixes.vm))
           computername   = try("${dbnode.name}-1", local.computer_names[idx + local.node_count])
+          role           = try(dbnode.role, "db")
           admin_nic_ip   = lookup(dbnode, "admin_nic_ips", ["false", "false"])[1]
           db_nic_ip      = lookup(dbnode, "db_nic_ips", ["false", "false"])[1]
         }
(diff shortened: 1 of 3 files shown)


deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -217,25 +217,37 @@ locals {
...
-      name         = try("${dbnode.name}-1", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx + local.node_count], local.resource_suffixes.vm))
-      computername = try("${dbnode.name}-1", local.computer_names[idx + local.node_count], local.resource_suffixes.vm)
-      role         = try(dbnode.role, "worker"),
-      db_nic_ip    = lookup(dbnode, "db_nic_ips", [false, false])[1],
-      admin_nic_ip = lookup(dbnode, "admin_nic_ips", [false, false])[1]
-      } if local.anydb_ha
-    ]
-    ]
+
+dbnodes = local.anydb_ha ? (
```
rukawata, line 221: instead of making the list twice, why not add the element within the loop?
eg (not tested but just showing the idea):
```
  dbnodes = compact([for idx, dbnode in try(local.anydb.dbnodes, [{}]) : [{
    name         = try("${dbnode.name}-0", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx], local.resource_suffixes.vm))
    computername = try("${dbnode.name}-0", local.computer_names[idx], local.resource_suffixes.vm)
    role         = try(dbnode.role, "worker"),
    db_nic_ip    = lookup(dbnode, "db_nic_ips", [false, false])[0]
    admin_nic_ip = lookup(dbnode, "admin_nic_ips", [false, false])[0]
    },
    local.anydb_ha ? {
      name         = try("${dbnode.name}-1", format("%s%s%s%s", local.prefix, var.naming.separator, local.virtualmachine_names[idx + local.node_count], local.resource_suffixes.vm))
      computername = try("${dbnode.name}-1", local.computer_names[idx + local.node_count], local.resource_suffixes.vm)
      role         = try(dbnode.role, "worker"),
      db_nic_ip    = lookup(dbnode, "db_nic_ips", [false, false])[1],
      admin_nic_ip = lookup(dbnode, "admin_nic_ips", [false, false])[1]
    } : ""
    ]
  ])
```

KimForss: That was what I wanted to do but did not know how :)

KimForss: I would keep this instead of the double list - it is easier to understand for a non developer :)

rukawata (approved):

PR #1017, opened 2020-12-18
Make the terraform generated password more complex

KimForss:
## Problem
The current password is not complex enough
Error: creating Linux Virtual Machine "xxxx" (Resource Group "YYYYY"): compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The supplied password must be between 6-72 characters long and must satisfy at least 3 of password complexity requirements from the following:\r\n1) Contains an uppercase character\r\n2) Contains a lowercase character\r\n3) Contains a numeric digit\r\n4) Contains a special character\r\n5) Control characters are not allowed" Target="adminPassword"
## Solution
Add these settings
  length           = 32
  min_upper        = 2
  min_lower        = 2
  min_numeric      = 2

## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
+++ b/deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
@@ -158,7 +158,11 @@ resource "random_password" "deployer" {
     && ! local.pwd_exist
     && local.input_pwd == null
   ) ? 1 : 0
-  length           = 16
+  
+  length           = 32
+  min_upper        = 2
+  min_lower        = 2
+  min_numeric      = 2
   special          = true
   override_special = "_%@"
 }
--- a/deploy/terraform/terraform-units/modules/sap_landscape/key_vault_sap_landscape.tf
+++ b/deploy/terraform/terraform-units/modules/sap_landscape/key_vault_sap_landscape.tf
@@ -114,7 +114,11 @@ resource "random_password" "iscsi_password" {
     && local.enable_iscsi_auth_password
     && ! local.iscsi_pwd_exist
   && try(local.var_iscsi.authentication.password, null) == null) ? 1 : 0
-  length           = 16
+
+  length           = 32
+  min_upper        = 2
+  min_lower        = 2
+  min_numeric      = 2
   special          = true
   override_special = "_%@"
 }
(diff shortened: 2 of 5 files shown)


deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
```diff
@@ -158,7 +158,11 @@ resource "random_password" "deployer" {
     && ! local.pwd_exist
     && local.input_pwd == null
   ) ? 1 : 0
-  length           = 16
+  
+  length           = 32
+  min_upper        = 2
+  min_lower        = 2
+  min_numeric      = 2
```
rukawata, line 165: why 2 not 1?

deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
```diff
@@ -158,7 +158,11 @@ resource "random_password" "deployer" {
     && ! local.pwd_exist
     && local.input_pwd == null
   ) ? 1 : 0
-  length           = 16
+  
+  length           = 32
```
rukawata, line 162: It says between 6-72 character, why 32?

rukawata (commented):
Thanks for the PR.
This works now, will it be possible to only set it somewhere, and pass it to the modules? And at the same time provide override in each module.

deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
KimForss: 2 makes it more random :)

deploy/terraform/terraform-units/modules/sap_deployer/key_vault.tf
KimForss: It could be shorter as well - it was my arbitrary choice - twice as long as 16 :)

KimForss:
> Thanks for the PR.
> This works now, will it be possible to only set it somewhere, and pass it to the modules? And at the same time provide override in each module.

I was planning to do that in another PR - with the current design all tiers will have a separate password which is not what customers want - typically all the VMs in the same SDU should have the same credentials

rukawata (approved):
I think it would be preferred to get confirmation from customers what is the normal requirement for password.
Also to re-factor so that those numbers can be easily customized would be better - can be in a separate PR. 
So will approve this for now.

PR #1020, opened 2020-12-19
[CREDENTIALS] Providing a way to provide credentials to landscape AB#137

KimForss:
## Problem
Currently the code uses an ssh key from the landscape key vault so that all systems in the landscape will share SSH keys. This does not work for username/password combinations.

## Solution
Add a credentials section to the landscape json and store the credentials from there into the landscape keyvault

## Tests
Add the following section to the json file
  "credentials": {
    "username": "azureadm",
    "password": "P@ss123W0rd"
  }

## Notes


.pipeline/tf-saplandscape-test.yml
```diff
@@ -37,6 +37,7 @@ stages:
         parameters:
           branch_name: $(sourceBranchName)
           saplandscape_env: "$(Build.BuildId)"
+          timeoutInMinutes: 20 # how long to run the step before automatically cancelling
```
rukawata, line 40: we have 120 min set for ansible test.
I will set it to 30 min for all terraform test - will be a separate PR.
Can you remove this from this PR?

deploy/terraform/run/sap_landscape/output.tf
```diff
@@ -3,11 +3,11 @@ output "vnet_sap_arm_id" {
 }
 
 output "landscape_key_vault_user_arm_id" {
-  value = try(module.sap_landscape.kv_user[0].id, "")
+  value = try(module.sap_landscape.kv_user, "")
 }
 
 output "landscape_key_vault_private_arm_id" {
-  value = try(module.sap_landscape.kv_prvt[0].id, "")
+  value = try(module.sap_landscape.kv_prvt, "")
```
rukawata, line 10: this way you are getting the the kv object, not id.
can we think about:
1. do we need id or the object?
2. if we need id, then output id only.
3. if we need object, then change the output name from xxx_arm_id to xxx.

deploy/terraform/run/sap_landscape/saplandscape_full.json
```diff
@@ -28,12 +28,18 @@
...
         "kv_iscsi_pwd": "",
         "kv_spn_id": ""
     },
     "sshkey": {},
-    "options": {}
+    "options": {},
+    "credentials" : {
+        "username" : "azureadm",
+        "password" : ""
+    }
```
rukawata, line 44: It would be better if it is consistent with sap_system.json:

```
 "authentication": {
      "type": "key",
      "username": "azureadm"
    }
```

or with password

```
 "authentication": {
      "type": "password",
      "username": "azureadm"
    }
```

Also, I would see sshkey section nested within authentication, eg:

```
 "authentication": {
      "type": "key",
      "username": "azureadm",
      "path_to_public_key": "sshkey.pub",
      "path_to_private_key": "sshkey"
    }
```

With any of the above changes, will impact Morgan's documentation - please communicate with him and get confirmed before we go ahead with this.

rukawata (commented):
Sorry, I think we can't move forward with this PR until Morgan' signs of on the new section in JSON.

deploy/terraform/run/sap_landscape/output.tf
KimForss: It is actually the id only - so probably better to rename the variable xxx_id

deploy/terraform/run/sap_landscape/variables_global.tf
```diff
@@ -8,20 +8,31 @@ variable "options" {
...
   description = "Details of ssh key pair"
   default = {
-    path_to_public_key  = "~/.ssh/id_rsa.pub",
-    path_to_private_key = "~/.ssh/id_rsa"
+    username = "azureadm"
+  }
+
+  validation {
+    condition = (
+      length(var.sshkey) >= 1
```
hdamecharla, line 19: Should this be var.authentication instead of var.sshkey? based on the changes made as part of https://github.com/Azure/sap-hana/pull/1047

deploy/terraform/run/sap_landscape/variables_global.tf
```diff
@@ -8,20 +8,31 @@ variable "options" {
...
+    )
+    error_message = "Either ssh keys or user credentials must be specified."
+  }
+  validation {
+    condition = (
+      length(trimspace(var.sshkey.username)) != 0
```
hdamecharla, line 25: would this be var.authentication.username, similar to the above comment?

deploy/terraform/run/sap_landscape/variables_global.tf
KimForss: yes - just fixing it :)

deploy/terraform/run/sap_landscape/variables_global.tf
KimForss: yup, fixing it now.

hdamecharla (approved):
@KimForss , approved the PR. All validations and tests are passing.

PR #1023, opened 2020-12-21
Support deployment of SAP_Library  to existing resource group

KimForss:
## Problem
Support deployment of SAP Library to existing resource group, as some customers want to put all the environment resources in the same resource group

## Solution
If arm_id is specified deploy to the resource group specified by it. This will enhance the code as we can get rid of the "is_existing" 
parameter
## Tests

## Notes

### DIFF
--- a/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
+++ b/deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
@@ -32,10 +32,10 @@ locals {
 
   // Resource group
   var_rg    = try(local.var_infra.resource_group, {})
-  rg_exists = try(local.var_rg.is_existing, false)
-  rg_arm_id = local.rg_exists ? try(local.var_rg.arm_id, "") : ""
+  rg_arm_id = try(var.infrastructure.resource_group.arm_id, "")
+  rg_exists = length(local.rg_arm_id) > 0 ? true : false
 
-  rg_name = try(var.infrastructure.resource_group.name, format("%s%s", local.prefix, local.resource_suffixes.library_rg))
+  rg_name   = local.rg_exists ? try(split("/", local.rg_arm_id)[4], "") : try(var.infrastructure.resource_group.name, format("%s%s", local.prefix, local.resource_suffixes.deployer_rg))
 
   // Storage account for sapbits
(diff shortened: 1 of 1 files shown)


deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
```diff
@@ -32,10 +32,10 @@ locals {
 
   // Resource group
   var_rg    = try(local.var_infra.resource_group, {})
-  rg_exists = try(local.var_rg.is_existing, false)
-  rg_arm_id = local.rg_exists ? try(local.var_rg.arm_id, "") : ""
+  rg_arm_id = try(var.infrastructure.resource_group.arm_id, "")
+  rg_exists = length(local.rg_arm_id) > 0 ? true : false
```
YunziZhangAzure, line 36: "rg_exists = length(local.rg_arm_id) > 0"
 
is more concise.

deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
```diff
@@ -32,10 +32,10 @@ locals {
...
 
-  rg_name = try(var.infrastructure.resource_group.name, format("%s%s", local.prefix, local.resource_suffixes.library_rg))
+  rg_name   = local.rg_exists ? try(split("/", local.rg_arm_id)[4], "") : try(var.infrastructure.resource_group.name, format("%s%s", local.prefix, local.resource_suffixes.deployer_rg))
```
YunziZhangAzure, line 38: Why change to deployer_rg?

YunziZhangAzure:
Currently, we have already supported deploying SAP to an existing resource group but with a different method, i.e. specifying "is_existing" and "arm_id"

I can see this PR is to implement the same feature by getting rid of "is_existing" and using "arm_id". I agree with doing such enhancement. Please modify the description in the "Problem" section to indicate this is an enhancement, instead of a feature.

deploy/terraform/terraform-units/modules/sap_library/variables_local.tf
KimForss: because I don't know how to copy & paste :)

YunziZhangAzure (approved):

PR #1024, opened 2020-12-23
[BUG] Support multiple disks for app tier

KimForss:
## Problem
Currently the code naming the disks fails when there are more disks defined for the app tier servers than servers. The code was using the count.index and not the vm_index property

## Solution
Correct the faulty logic

## Tests

## Notes


YunziZhangAzure (approved):
Thanks for addressing it!

PR #1026, opened 2020-12-29
Support bringing your own keyvault for SAP System

KimForss:
## Problem
Currently the code will always create the 2 key vaults in the SDU, sometimes customers want to provide their existing keyvaults to be used

## Solution
Add logic to support using existing key vaults

## Tests

## Notes


deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
```diff
@@ -111,7 +111,7 @@ locals {
   secret_sid_pk_name = try(local.landscape_tfstate.sid_public_key_secret_name, "")
 
   // Define this variable to make it easier when implementing existing kv.
-  sid_kv_user = try(var.sid_kv_user[0], null)
+  sid_kv_user = var.sid_kv_user_id
```
rukawata, line 114: please rename the local var to reflect this is an id :)

rukawata (commented):
Overall LGTM, just one thing on the var naming.

deploy/terraform/terraform-units/modules/sap_system/anydb_node/variables_local.tf
KimForss: done

rukawata (approved):

PR #1027, opened 2020-12-30
Added a check that replaces "_" in VNet names

KimForss:
## Problem
Currently the code fails if the VNet Names have "_" in the first 7 characters of the name. It fails when trying to create the key vault secrets

## Solution
Replacing the "_" characters in vnet names

## Tests
Create a vnet with "_" in the first 7 characters of the name and use that for deployment

## Notes


rukawata:
I think it should be okay to have _ in vnet name. The essential problem seems that we can't have _ in kv secret names?
If that is the case, I would only remove _ from kv secret names.

rukawata (approved):
As discussed, please document this new requirement in the naming document.

PR #1030, opened 2021-01-04
Define the container names in the naming module

KimForss:
## Problem
Currently the tfstate & sapbits names are hardcoded in the code. This causes problems for customers who have naming conventions even for container naming in storage accounts. 

There is a need to store the output ansible files in Azure as storing them in the file system will not work for automated deployments. This PR also adds a container "Ansible" in the terraform storage account.

## Solution
Add the names into the naming module,

## Tests
Re-deploy the solution

## Notes


rukawata (commented):
Thanks for adding supportability to name container with naming module. 
But the ansible container is newly introduced and I do not know what this is for.
Can we get some clarification on that?

deploy/terraform/bootstrap/sap_library/saplibrary_full.json
```diff
@@ -12,5 +12,36 @@
...
+        "account_replication_type" : "LRS",
+        "account_kind": "StorageV2",
+        "tfstate_blob_container" :{
+            "is_existing" : false,
+            "name": "tfstate"
+        },
+        "ansible_blob_container" :{
+            "is_existing" : false,
+            "name": "ansible"
+        }
```
rukawata, line 45: Thanks for removing the ansible container part. Can you also remove this?

deploy/terraform/terraform-units/modules/sap_namegenerator/variables_local.tf
```diff
@@ -208,6 +208,7 @@ variable resource_suffixes {
     "admin_nic"           = "-admin-nic"
     "admin_subnet"        = "admin-subnet"
     "admin_subnet_nsg"    = "adminSubnet-nsg"
+    "ansible"             = "ansible"
```
rukawata, line 211: this also needs to be removed. Thanks!

KimForss: removed

rukawata (approved):
LGTM!

PR #1031, opened 2021-01-04
Use vnet from landscape state file

KimForss:
## Problem
The current code reads the vnet arm_id from the SDU json when it shoud read it from the landscape state file

## Solution
Read the VNet ARM ID from the landscape state file

## Tests
No steps needed

## Notes

### DIFF
--- a/deploy/terraform/run/sap_system/variables_local.tf
+++ b/deploy/terraform/run/sap_system/variables_local.tf
@@ -40,11 +40,6 @@ locals {
   // The environment of sap landscape and sap system
   environment = upper(try(var.infrastructure.environment, ""))
 
-  vnet_sap_name = local.vnet_sap_exists ? try(split("/", local.vnet_sap_arm_id)[8], "") : try(local.var_vnet_sap.name, "sap")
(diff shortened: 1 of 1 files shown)


deploy/terraform/run/sap_system/variables_local.tf
```diff
@@ -115,6 +104,18 @@ locals {
   deployer_tfstate_key         = try(var.deployer_tfstate_key, "")
   landscape_tfstate_key        = try(var.landscape_tfstate_key, "")
 
+  // SAP vnet
+  var_infra       = try(var.infrastructure, {})
+  var_vnet_sap    = try(local.var_infra.vnets.sap, {})
+  vnet_sap_arm_id = try(data.terraform_remote_state.landscape.vnet_sap_arm_id, try(local.var_vnet_sap.arm_id, ""))
```
rukawata, line 110: The only thing changed is this line, but I think previous behavior is correct.
This `vnet_sap_arm_id` is only used when someone introduce a vnet that is not deployed with our `sap_landscape` module (meaning created not using our tf code).

rukawata:
Close as invalid.

PR #1047, opened 2021-01-14
Renaming sshkey block to authentication AB#137

KimForss:
## Problem
The deployed systems may use either sshkey or username/password authentication. Currently the ssh keys are provided in a "sshkey" block. This is not a logical block name for username/password combinations

## Solution
Rationalize the input json for both ssh and username/password authentication by renaming the sshkey block to authentication.

## Tests

## Notes


deploy/terraform/terraform-units/modules/sap_landscape/variables_local.tf
```diff
@@ -65,8 +65,8 @@ locals {
 
   // By default, ssh key for iSCSI uses generated public key. Provide sshkey.path_to_public_key and path_to_private_key overides it
   enable_iscsi_auth_key = local.enable_iscsi && local.iscsi_auth_type == "key"
-  iscsi_public_key      = local.enable_iscsi_auth_key ? (local.iscsi_key_exist ? data.azurerm_key_vault_secret.iscsi_pk[0].value : try(file(var.sshkey.path_to_public_key), tls_private_key.iscsi[0].public_key_openssh)) : null
-  iscsi_private_key     = local.enable_iscsi_auth_key ? (local.iscsi_key_exist ? data.azurerm_key_vault_secret.iscsi_ppk[0].value : try(file(var.sshkey.path_to_private_key), tls_private_key.iscsi[0].private_key_pem)) : null
+  iscsi_public_key      = local.enable_iscsi_auth_key ? (local.iscsi_key_exist ? data.azurerm_key_vault_secret.iscsi_pk[0].value : try(file(var.authentication.path_to_public_key), tls_private_key.iscsi[0].public_key_openssh)) : null
```
hdamecharla, line 68: Is iSCSI in this context, a SBD VM?

KimForss: Yes, however I have not really tested that part at all ;)

hdamecharla (commented):
As far as I remember, we need either the ```admin_password``` or ```admin_ssh_key``` to be specified when creating the Linux VM resources. As an additional setting when ```admin_password``` is specified ```disable_password_authentication``` **must** be set to ```false```.  

Are we tracking if password authentication is disabled when username/password is not provided? Or, should we let user handle it's state?

KimForss:
> As far as I remember, we need either the `admin_password` or `admin_ssh_key` to be specified when creating the Linux VM resources. As an additional setting when `admin_password` is specified `disable_password_authentication` **must** be set to `false`.
> 
> Are we tracking if password authentication is disabled when username/password is not provided? Or, should we let user handle it's state?

We are tracking it - see here:
https://github.com/Azure/sap-hana/blob/dc5799f20e775889645398b6451410f44d979aaa/deploy/terraform/terraform-units/modules/sap_landscape/iscsi.tf#L80

hdamecharla:
> We are tracking it - see here:
> https://github.com/Azure/sap-hana/blob/dc5799f20e775889645398b6451410f44d979aaa/deploy/terraform/terraform-units/modules/sap_landscape/iscsi.tf#L80
Thank you 👍

hdamecharla (approved):
@KimForss  Approving PR

PR #1054, opened 2021-01-28
[HOTFIX] Outputs fail when scaling down an existing deployment

KimForss:
## Problem
If we scale down the deployment (reduce the server count) the code fails

## Solution
Control the dimensions of the output slices by adding a server count to them

## Tests
Do a deployment with more than 2 app servers and then scale down the count to 1

## Notes


hdamecharla (approved):
Hi @KimForss , approved #1054

PR #1062, opened 2021-02-04
[CONNECTIVITY] Add a route table to the environment AB#136

KimForss:
## Problem
We need a route table to route the traffic to an Azure Load Balancer

## Solution
Add a route table for those cases when the automation creates the Virtual Network

## Tests
Normal deployment

## Notes


hdamecharla (approved):
Hi @KimForss, Approved #1062

PR #1063, opened 2021-02-05
Add a route to the firewall environment AB#136 

KimForss:
## Problem
We need to be able to route the traffic to a specified endpoint

## Solution
Adding a user defined route and route table association for each created subnet

## Tests

## Notes


hdamecharla (approved):
Hi @KimForss, Approved #1063

PR #1068, opened 2021-02-12
[BREAKING CHANGE]  Incorrect VNet name in prefix

KimForss:
## Problem
The current code does not read the Virtual network from the environment state file. This causes incorrect an incorrect naming prefix to be generated

## Solution
Read the correct VNet Identifier

## Tests

## Notes
This will also add a version label for the deployments so that we can trace based on the statefile what version was used to deploy the infrastructure


hdamecharla (approved):
Hi @KimForss, approved #1068 👍

PR #1073, opened 2021-02-26
[HOTFIX] Changed the NIC order

KimForss:
## Problem
In Linux only the primary network interface gets the default gateway configured.

## Solution
Add the app & db nic as the primary nic as the secondary NICs are all in the same subnet

## Tests

## Notes


hdamecharla (approved):
Hi @KimForss, approved #1073

PR #1096, opened 2021-03-23
ENHANCEMENT Support 4 IPs for HA Win CS

KimForss:
## Problem
A clustered Windows SCS needs 4 IP addresses and the current design only has two

## Solution

## Tests
Add two additional IPs for HA Windows SCS

## Notes


hdamecharla (approved):
Hi @KimForss, Approved #1096 👍🏾

PR #1101, opened 2021-04-06
Use the value of ARM_SUBSCRIPTION_ID as sub id

### LINKED ISSUE #1100: Beta Branch: Incorrectly using string ARM_SUBSCRIPTION_ID rather than value of that variable
https://github.com/Azure/sap-hana/blob/7d88ba083a0963f2e0ec97325db99902c2ff6f7e/deploy/scripts/installer.sh#L154

In my testing of the beta branch I was seeing errors during prepare_region.sh dur trying to lookup the subscription with an invalid subscription value; after a little investigation I found the above line. Changing it to read

     STATE_SUBSCRIPTION=${ARM_SUBSCRIPTION_ID}

gets past the problem.

rtamalin:
When initialising the STATE_SUBSCRIPTION, use the value of the
ARM_SUBSCRIPTION_ID variable, rather than the variable name, as the
subscription id.

Fixes #1100

### DIFF
--- a/deploy/scripts/installer.sh
+++ b/deploy/scripts/installer.sh
@@ -151,7 +151,7 @@ else
         ARM_SUBSCRIPTION_ID=$(echo "${temp}" | cut -d= -f2 | tr -d \" | xargs)
     fi
     
-    STATE_SUBSCRIPTION=ARM_SUBSCRIPTION_ID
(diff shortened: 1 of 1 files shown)


rtamalin:
Fixed by https://github.com/Azure/sap-hana/commit/5587945f6b7f8c7a28149fb732b6f5c444c9200a

PR #1108, opened 2021-04-14
Add support for using predefined LUN ranges + ability to add additional data disks

KimForss:
## Problem
Currently the code does not support adding data disks in a non destructive fashion, this is a problem when using custom sizing

## Solution
Add a section in the disk sizing json for the added disks to ensure they will be added to the end of the list of disks TF creates

## Tests

## Notes


hdamecharla (approved):
Hi @KimForss, Approved #1108 👍🏾  🙏🏾

PR #1111, opened 2021-04-20
Changing Kdump LUN id From 3 to 2 

balram2697:
## Problem
We use LUN ID '0' for boot lun, LUN ID '1' for config lun and LUN ID '3' for kdump lun ..hence we are skipping LUN ID '2'. So as a standard instead of skipping 1 ID in between, we will start using Kdump lun id as '2'.

## Solution
Changing Kdump LUN ID '3' to '2' so that we don't have to skip any lun id in between.

## Tests
Locally tested this Script with LUN_ID '2' on HANA Instance, it's working as expected (Able to identify the dedicated kdump lun)


tools/enable-kdump.sh
```diff
@@ -43,13 +43,13 @@ done
 
 # sg_scan -i command will give mapping for the devices
 # output look like
-# /dev/sg32: scsi17 channel=0 id=3 lun=3
+# /dev/sg32: scsi17 channel=0 id=2 lun=2
```
jaawasth, line 46: make lun_id a var and use it everywhere instead of changing at all places, do test it once [just the lun discovery part]

balram2697: Done

shaaga (approved):

jaawasth (approved):

PR #1113, opened 2021-04-22
Plugins folder is not available inside .terraform

prasroy:
## Problem
Plugins folder is not available inside .terraform

## Solution
We request to comment out the blocks

## Tests
remover.sh execution went ahead

## Notes
The entire code didn't executed successfully


KimForss:
Resolved

PR #1115, opened 2021-04-23
Fix some ansible issues for roles-os/1.5-disk-setup

rtamalin:
When I ran the OS setup playbooks this morning using the latest version of the `alfa` branch I hit some error which I fixed up as follows:

  * Indent the msg argument to the debug action
  * Use `~` rather than `+` when combining facts with facts, or
    facts with strings to avoid type incompatibility issues; the
    Jinja2 `~` concatenation operator automatically converts it's
    arguments to strings.
  * Ensure that references to conditionally initialised `pvs_data`,
    `pvs_log` and `pvs_sap` use `default([])` when generating the
    comma joined string from the list, so that an empty string is
    generated rather than an error.
  * Add a condition check for item.pvs being empty (using `| bool`
    yields False for an empty string) when creating VGs so as to
    avoid errors for disk groups that had no items defined in the
    terraform outputs.

Also removed training whitespaces present in the file.

### DIFF
--- a/deploy/ansible/roles-os/1.5-disk-setup/tasks/main.yml
+++ b/deploy/ansible/roles-os/1.5-disk-setup/tasks/main.yml
@@ -14,7 +14,7 @@
 
 # -------------------------------------+---------------------------------------8
 #
-# <Comment Header> 
+# <Comment Header>
 #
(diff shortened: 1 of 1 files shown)


KimForss:
Thanks

PR #1116, opened 2021-04-23
Mkdeegan/bom processing

mkdeegan:
## Cleanup
remove files brought in by #1110, as this should not have been merged

## Feature
Playbook for BOM processing and BOM Processing Role


mkdeegan:
Kimmo,
I removed the files that #1110 introduced
I added the BOM playbook and role.

Please double check the #1110 stuff. If you have a better reversal I can recreate the BOM drop

KimForss (approved):

PR #1117, opened 2021-04-23
roles-os/1.3-repository tweaks

rtamalin:
Quieten the Ansible warnings that are generated as a result of running
rm, zypper, or yum directly from a command or shell module.

Reword some comments and name attributes to either better reflect what
the Ansible code is doing, and/or shorten them. Also reformat some long
lines where possible.


hdamecharla (approved):
Hi @KimForss/@mkdeegan , approved changes in #1117  from @rtamalin 👍🏾

PR #1119, opened 2021-04-25
Ansible OS roles changes - packages, sudo, timezone

msftrobiro:
Ansible OS packages
- 1.0 sudoers - new, sets logging for sudo
- 1.4 packages - removed ntp*
                        - changed to OS-major version as RHEL8/SLES15 have differences (e.g. xulrunner is no longer deliverd for RHEL8)
- 1.6 timezone - simple role that adds static/global timezone
- 2.1 SAP OS packages - added uuidd
                                     - changed package list to include placeholders (compat-sap-*) for all 5 versions and more in future


KimForss (approved):
Thanks, looks goot to me

PR #1123, opened 2021-04-27
Re-work 1.4-packages to add tier and state fields

rtamalin:
By updating the package list specification to specify the target tier as
well as the desired state (present or absent) this allows us to specify
both packages we want installed, as well as packages we want removed,
for a given target tier. Note that an 'all' tier entry will always be
considered a match for the active target tier.

When calling the package module we extract the matching list of packages
for a given state into a list to be passed as the name parameter per the
recommended calling pattern; this minimises the calls to the underlying
package manager.

Add definitions for 'distro_name' and 'distro_id' to the 1.4-packages
role defaults to make it easier to read the code.
NOTE: This is an example of the sort of definitions that it is useful
to include in a group_vars/all file, making then available to every
playbook, role, and task without having to specify them or explicitly
include them.

Add target tier specification to the top level playbook associated with
the roles-os roles.

Add support to test_menu.sh to pass any arguments specified when running
it as additional arguments when calling ansible-playbook. For example
this allows verbosity control for the running ansible playbook.

### DIFF
--- a/deploy/ansible/roles-os/1.4-packages/defaults/main.yml
+++ b/deploy/ansible/roles-os/1.4-packages/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+# TODO: Maybe move these to a group_vars/all/distro file so that they
+# can be shared by all playbooks/tasks automatically, and extend with
+# standardised versions of all similar patterns used in the playbooks.
+distro_name: "{{ ansible_os_family|upper }}-{{ ansible_distribution_major_version }}"
+distro_id: "{{ ansible_os_family|lower ~ ansible_distribution_major_version }}"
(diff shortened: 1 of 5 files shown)


rtamalin:
Note that I've marked this PR as draft because it depends upon my other PR #1122 and should be rebased once that one merges.

deploy/ansible/roles-os/1.4-packages/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# TODO: Maybe move these to a group_vars/all/distro file so that they
+# can be shared by all playbooks/tasks automatically, and extend with
+# standardised versions of all similar patterns used in the playbooks.
+distro_name: "{{ ansible_os_family|upper }}-{{ ansible_distribution_major_version }}"
+distro_id: "{{ ansible_os_family|lower ~ ansible_distribution_major_version }}"
```
KimForss, line 6: nit : no newline at the end

rtamalin: Will fix ;-)

mkdeegan (approved):
Looks good

PR #1124, opened 2021-04-28
Implement the 1.7-chrony role

rtamalin:
role-os/1.7-chrony:
===================

This role ensures that `chrony` is installed on the system, and if not
will trigger it to be installed. It also complains if `ntp` is still
installed, which should have been removed by the 1.4-packages role.

It then determines if there are any custom chrony config settings that
have been specified and if so generates the list of candidate entries
that will need to be present.

If there are no custom settings then:
  * remove any existing ansible managed block that exists in the config
    file that may have previously added using blockinfile.
  * ensure that the config is updated to (re-)enable any distro default
    pool or server settings that my be commented out; SLES ships with a
    default config that has a commented out pool entry. Similarly if we
    are switching back from custom settings to the defaults we will need
    to uncomment the default pool or server settings.

If there are custom settings then:
  * ensure that we comment out the existing distro default settings that
    may exist before adding the custom settings via an ansible managed
    block.
  * add an ansible managed block, using blockinfile, that specifies the
    identified custom settings.

Note that it is possible to switch back and forth between custom and
defaults due to workflow described above.

Whether or not the chronyd service needs to be installed is tracked by
the `chronyd_restart_required` fact, which amalgamates the status of
the various actions, such as the package being installed, or config
file changes being made. At the end of the role's run this fact is
checked to determine if the service does in fact need to be restarted.

Additionally we also ensure that the service is enabled, since it is
installed as disabled on SUSE systems.

Finally we have a catchall that ensures the service running, in case it
didn't get restarted to to config file changes, but is not actually
running.

Other Changes:
==============
In the ansible-input-api.yaml add an OS Config setting section to hold
the new `chrony_pool` and `chrony_servers` settings that can be used to
provide custom settings.

Update the SLE 12 & 15 distro package lists in the 1.4-packages role to
include chrony as a package to be installed for the `os` tier/phase.

Update the top level playbook for phase 01 to include the 1.7-chrony
role when it is run.

### DIFF
--- a/deploy/ansible/playbook_01_os_base_config.yaml
+++ b/deploy/ansible/playbook_01_os_base_config.yaml
@@ -57,6 +57,9 @@
       name:     roles-os/1.4-packages
 
   - include_role:
+      name:     roles-os/1.7-chrony
+
+  - include_role:
       name:     roles-os/1.20-prometheus
     when:   prometheus
             and             ansible_os_family|upper == "SUSE"
--- a/deploy/ansible/roles-os/1.4-packages/vars/os-packages.yaml
+++ b/deploy/ansible/roles-os/1.4-packages/vars/os-packages.yaml
@@ -79,6 +79,7 @@ packages:
       - { tier: 'os', package: 'cifs-utils',             state: 'present' }
 
     suse12:
+      - { tier: 'os', package: 'chrony',         state: 'present' }
       - { tier: 'os', package: 'libyui-qt-pkg7', state: 'present' }
       - { tier: 'os', package: 'glibc',          state: 'present' }
       - { tier: 'os', package: 'systemd',        state: 'present' }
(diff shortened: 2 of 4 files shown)


rtamalin:
Note that I have tested this using my SLES based deployment, switching back and forth between custom and default settings successfully, and did some investigation to see what the SLE and RHEL default config files look like for RHEL 7/8 and SLES 12/15, and it should work for all those cases...

rtamalin:
Note that this PR is based on PR #1123 so I will rebase it once that has landed and remove the draft status then. However the code should be ready for review now.

rtamalin:
Rebased onto current alfa branch and ready for review @KimForss @mkdeegan

hdamecharla (approved):
Hi @rtamalin , approved #1124 👍🏾 

Hi @KimForss , can we merge this into alfa please?  🙏🏾

PR #1130, opened 2021-04-30
Implement 1.11-accounts

rtamalin:
Implement a role that can be used to create admin account, and
associated groups, with matching /etc/sudoers.d/<username> files.

The role supports both adding and removing admin users and groups,
as well as allowing account settings to be associated with a given
playbook tier, and enabled/disabled via the associated attribute.

The details of the admin groups and accounts to be created are specified
in the top level vars/admin-accounts.yml file. Currently this file holds
plain text passwords and static uid and gid definitions; this needs to
change to creating/retrieving passwords from a secure source such as a
keyvault, and allow for dynamic uid & gid assignments which are tracked
in a persistent fashion in a keyvault.

Currently just two "flavours" of sudoers roles are defined:
  * admin_no_password:
    all commands permitted without need to enter a password.
  * admin_password_required:
    all commands permitted but password validation required.

Additional flavours can be added by adding new templates with a name
in the format 'sudoers_<role_name>.j2'. Optionally a vars file with
the same name can be created, which will be dynamically loaded if
found, that can hold any useful fact definitions that may aid in
processing the matching template file.


rtamalin:
@KimForss @mkdeegan I'm tested this creating and removing the reference example user/group, so the "plumbing" is working correct. The next step would be to add the appropriate user settings to the `vars/admin-accounts.yml` for any reference admin user we want to create in our test deployments for the `os` playbook tier and similarly add initial details for any SAP accounts we want to manage for the other playbook tiers... Hopefully it should just be a matter of re-including this role as the "2.3 create SAP users" step, at least for the first pass...

deploy/ansible/roles-os/1.0-sudoers/defaults/setup_user_sudoers.yml
```diff
@@ -0,0 +1,16 @@
...
+
+
+# Default role to use if none specified
+sudoers_role:  admin_no_password
+
+# Default to creating a sudoers file
+sudoers_state: present
+
+# Where to create sudoers files
+sudoers_path:  /etc/sudoers.d
```
mkdeegan, line 16: as discussed, this is unit of work should not be cross role, and is better performed from 1.11-accounts.
Please relocate.

deploy/ansible/roles-os/1.0-sudoers/tasks/setup_user_sudoers.yml
```diff
@@ -0,0 +1,77 @@
...
+    validate:   /usr/sbin/visudo -cf %s
+  when:
+    -           sudoers_state == "present"
+
+- name:         Remove sudoers file for {{ sudoers_user }} if required
+  file:
+    path:       "{{ sudoers_path }}/{{ sudoers_user }}"
+    state:      absent
+  when:
+    -           sudoers_state == "absent"
```
mkdeegan, line 77: as discussed, this is unit of work should not be cross role, and is better performed from 1.11-accounts.
Please relocate.

deploy/ansible/roles-os/1.0-sudoers/templates/sudoers_admin_no_password.j2
```diff
@@ -0,0 +1,2 @@
+# User rules for {{ sudoers_user }} with role {{ sudoers_role }}
+{{ sudoers_user }} ALL=(ALL) NOPASSWD:ALL
```
mkdeegan, line 2: as discussed, this is unit of work should not be cross role, and is better performed from 1.11-accounts.
Please relocate.

deploy/ansible/roles-os/1.0-sudoers/templates/sudoers_admin_password_required.j2
```diff
@@ -0,0 +1,2 @@
+# User rules for {{ sudoers_user }} with role {{ sudoers_role }}
+{{ sudoers_user }} ALL=(ALL) PASSWD:ALL
```
mkdeegan, line 2: as discussed, this is unit of work should not be cross role, and is better performed from 1.11-accounts.
Please relocate.

deploy/ansible/roles-os/1.0-sudoers/vars/sudoers_default.yml
```diff
@@ -0,0 +1,7 @@
+---
+# -------------------------------------+---------------------------------------8
+#
+# Empty vars file that should exist to avoid fails from with_first_found
+# searches.
+#
+# -------------------------------------+---------------------------------------8
```
mkdeegan, line 7: as discussed, this is unit of work should not be cross role, and is better performed from 1.11-accounts.
Please relocate.

mkdeegan (changes requested):

deploy/ansible/roles-os/1.0-sudoers/defaults/setup_user_sudoers.yml
rtamalin: Done

deploy/ansible/roles-os/1.0-sudoers/tasks/setup_user_sudoers.yml
rtamalin: Done

deploy/ansible/roles-os/1.0-sudoers/templates/sudoers_admin_no_password.j2
rtamalin: Done

deploy/ansible/roles-os/1.0-sudoers/templates/sudoers_admin_password_required.j2
rtamalin: Done

deploy/ansible/roles-os/1.0-sudoers/vars/sudoers_default.yml
rtamalin: Done

rtamalin:
@mkdeegan As requested I've moved the bits I added to the 1.0-sudoers role into the 1.11-accounts role... Wasn't able to entirely eliminate the use of includes though. ;-)

KimForss:
Let's merge this and revisit if needed

KimForss (commented):
Looks good to me

PR #1140, opened 2021-05-10
Fix validate.sh

rtamalin:
When attempting to run validate.sh it hung, attempt to run jq against
stdin. Upon investigating I identified that there were some occurences
of jq in the validate.sh that were missing references to the parameter
file when being call.

Once I fixed them I noticed a pattern of calling jq with output piped
to the td command to remove double quotes; the jq command can do this
automatically with the -r/--raw-output option, so I updated the code
to reflect this.

I also noticed that sometimes the parameter file was being cat'd to
the stdin of jq, and other times was being passed as the last argument
to jq; I've cleaned things up so that the calling pattern for jq is
consistent, passing the parameter file as the last argument to the jq
command when it is called.

I also added single quoting of the query string argument to some cases
that used characters that could be interpretted by the shell, e.g. the
string "databases[0].platform" on it's work will work correctly 99.9%
of the time, but if there ever happened to be a file with the name
"database0.platform" in the runtime current directort for the jq command
it would cause the "databases[0].platform" pattern to be changed to the
file name, with obvious weird, and hard to debug, effects.

Eliminated some unnecessary whitespace in the jq command lines.

### DIFF
--- a/deploy/scripts/validate.sh
+++ b/deploy/scripts/validate.sh
@@ -96,11 +96,11 @@ then
 fi
 
 # Read environment
-environment=$(jq .infrastructure.environment "${parameterfile}" | tr -d \")
-region=$( jq .infrastructure.region "${parameterfile}" | tr -d \")
+environment=$(jq --raw-output .infrastructure.environment "${parameterfile}")
+region=$(jq --raw-output .infrastructure.region "${parameterfile}")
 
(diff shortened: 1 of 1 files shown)


rtamalin:
@KimForss FYI noticed an issue with validate.sh and as part of fixing it I made the use of `jq` in that file more consistent.

KimForss (approved):

PR #1143, opened 2021-05-11
Further cleanup of jq related issues.

rtamalin:
deploy_utils.sh
===============
Add fail_if_null() function that can be used to fail if a specified
variable that was initialised with the result of a jq query has the
value 'null', which happens when you query an non-existent entry.

Add error_msg() function that can be used to report error messages
in a consistent fashion.

TODO:
  * Moved colourised message support into deploy_utils.sh and have
    error_msg() output errors in red.

validate.sh
===========
The validate.sh script's usage of jq doesn't correctly handle the
case where we get a 'null' back when attempting to lookup an invalid
reference.

Additionally in a number of cases the checks for values being set in
the config files appear to have been mistakenly inverted, checking for
retrieve values being an empty, rather than non-empty, string. This
seems to have "worked" previously due to the the 'null' value being
returned, and thus the inverted check appeared to do the right thing.

Additionally I added support for tracking if errors occurred during
the validating checking, allowing the validate.sh command to exit
with a non-zero exit status indicating failure.  The error() function
makes this easy to manage as well as printing the error message in
bold red text.

The new heading() function prints headings in cyan with a following
line of '-' characters in a consistent fashion, with consistent
formatting/layout.

Fixed a whitespace issue in the indentation of the displayed subnet
address range for the deployer.

TODO:
  * Convert the repeated pattern of code for the subnet processing
    when validating sap_system config files into a routine.
  * Convert the output echo commands to use printf commands with
    format strings the handle the whitespace indentation.

prepare_region.sh
=================
Updates jq queries to use --raw-output and removes post processing
that is no longer required.

Also catch any non-zero exit status, rather than specifically 255,
when calling other scripts and exit out immediately with that exit
status.

install_deployer.sh, remove_region.sh, remover.sh
====================================================================
Updates jq queries to use --raw-output and removes post processing
that is no longer required.

install_library.sh
============
Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

installer.sh
============
Ensire that if we get a 'null' back when checking for .deployer.use
that we treat it as a false.

Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

Also quotes all jq query strings that contain shell pattern match
characters to avoid potential for side effects.

install_workloadzone.sh
=======================
Simplifies and logically fixes some if conditions to simply check
for values being empty or non-empty value, rather than a negated
non-empty or empty value, respectively. In some cases we were
checking if a value was non-empty, and if so dynamically retrieved
the correct value again.

Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

Also quotes all jq query strings that contain shell pattern match
characters to avoid potential for side effects.deploy_utils.sh
===============
Add fail_if_null() function that can be used to fail if a specified
variable that was initialised with the result of a jq query has the
value 'null', which happens when you query an non-existent entry.

Add error_msg() function that can be used to report error messages
in a consistent fashion.

TODO:
  * Moved colourised message support into deploy_utils.sh and have
    error_msg() output errors in red.

validate.sh
===========
The validate.sh script's usage of jq doesn't correctly handle the
case where we get a 'null' back when attempting to lookup an invalid
reference.

Additionally in a number of cases the checks for values being set in
the config files appear to have been mistakenly inverted, checking for
retrieve values being an empty, rather than non-empty, string. This
seems to have "worked" previously due to the the 'null' value being
returned, and thus the inverted check appeared to do the right thing.

Additionally I added support for tracking if errors occurred during
the validating checking, allowing the validate.sh command to exit
with a non-zero exit status indicating failure.  The error() function
makes this easy to manage as well as printing the error message in
bold red text.

The new heading() function prints headings in cyan with a following
line of '-' characters in a consistent fashion, with consistent
formatting/layout.

Fixed a whitespace issue in the indentation of the displayed subnet
address range for the deployer.

TODO:
  * Convert the repeated pattern of code for the subnet processing
    when validating sap_system config files into a routine.
  * Convert the output echo commands to use printf commands with
    format strings the handle the whitespace indentation.

install_deployer.sh, prepare_region.sh, remove_region.sh, remover.sh
====================================================================
Updates jq queries to use --raw-output and removes post processing
that is no longer required.

install_library.sh
============
Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

installer.sh
============
Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

Also quotes all jq query strings that contain shell pattern match
characters to avoid potential for side effects.

install_workloadzone.sh
=======================
Simplifies and logically fixes some if conditions to simply check
for values being empty or non-empty value, rather than a negated
non-empty or empty value, respectively. In some cases we were
checking if a value was non-empty, and if so dynamically retrieved
the correct value again.

Leverages fail_if_null() to fail when jq queries result in a 'null'
value.

Also updates jq queries to use --raw-output and removes the post
processing that is no longer required.

Also quotes all jq query strings that contain shell pattern match
characters to avoid potential for side effects.

### DIFF
--- a/deploy/scripts/deploy_utils.sh
+++ b/deploy/scripts/deploy_utils.sh
@@ -75,3 +75,26 @@ function init() {
     
     
 }
+
+function error_msg {
+    echo "Error!!! ${@}"
+}
+
+function fail_if_null {
+    local var_name="${1}"
+
+    # return immeditaely if no action required
+    if [ "${!var_name}" != "null" ]; then
+        return
+    fi
+
+    shift 1
+
+    if (( $# > 0 )); then
+        error_msg "${@}"
+    else
+        error_msg "Got a null value for '${var_name}'"
+    fi
+
+    exit 1
+}
--- a/deploy/scripts/install_deployer.sh
+++ b/deploy/scripts/install_deployer.sh
@@ -98,8 +98,8 @@ fi
 
 
 # Read environment
-environment=$(cat "${parameterfile}" | jq .infrastructure.environment | tr -d \")
-region=$(cat "${parameterfile}" | jq .infrastructure.region | tr -d \")
+environment=$(jq --raw-output .infrastructure.environment "${parameterfile}")
+region=$(jq --raw-output .infrastructure.region "${parameterfile}")
 key=$(echo "${parameterfile}" | cut -d. -f1)
 
 if [ ! -n "${environment}" ]
--- a/deploy/scripts/install_library.sh
+++ b/deploy/scripts/install_library.sh
@@ -95,11 +95,14 @@ if [ $param_dirname != '.' ]; then
 fi
 
 # Read environment
-environment=$(jq .infrastructure.environment "${parameterfile}" | tr -d \")
-region=$(jq .infrastructure.region "${parameterfile}" | tr -d \")
+environment=$(jq --raw-output .infrastructure.environment "${parameterfile}")
+region=$(jq --raw-output .infrastructure.region "${parameterfile}")
 key=$(echo "${parameterfile}" | cut -d. -f1)
 
-use_deployer=$(jq .deployer.use "${parameterfile}" | tr -d \")
+use_deployer=$(jq --raw-output .deployer.use "${parameterfile}")
+if [ "${use_deployer}" == "null" ]; then
+    use_deployer=false
+fi
 
 if [ ! -n "${environment}" ]
 then
@@ -128,7 +131,7 @@ then
 fi
 
 
-echo $use_deployer
+echo "Use Deployer: $use_deployer"
 
 if [ false != $use_deployer ]
 then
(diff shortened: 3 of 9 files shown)


rtamalin:
@KimForss @hdamecharla @mkdeegan This patchset attempts to address all of cases where we are querying with `jq` and possibly getting a `null` result such that we fail in such cases, instead of attempting to use `null` as the value for a queried config setting or az CLI command attribute.

Creating it as draft initially while I test it.

The validate.sh changes are verified at least, with no errors being reported for the configs I have tested it against.

I will test the rest of the changes as part of getting my deployment up and working again. ;-)

KimForss (approved):

PR #1145, opened 2021-05-19
Update the default sap disk sizes for HANA DB nodes

rtamalin:
The existing default size of 64(G) is no longer viable as we are
attempting to create multiple LVs in the vg_sap VG, one of which
is 64G in size.

This change updates the default sizes for all the other models to
match that used for the S4Demo model.


rtamalin:
@KimForss @mkdeegan Don't know if this is being addressed elsewhere, but here is a first pass at a patch that "fixes" this issue ;-)

KimForss (approved):

PR #1149, opened 2021-05-21
Fix some typos I spotted while trying to use remover.sh

rtamalin:
The remover.sh script was failing due to a missing $ before a variable
dereference.


KimForss (approved):
looks good to me

PR #1150, opened 2021-05-24
Add further ansible.cfg settings

rtamalin:
Configure the default inventory, private_key_file and remote_user
settings so that is not necessary to specify them when running the
ansible-playbook (or ansible) command.

Note that while the private_key_file name will be specified, the
actual file will still need to be downloaded separately.


rtamalin:
@KimForss @mkdeegan @hdamecharla This patch makes it easier to run ansible commands without needing to specify additional paramters; the generated inventory includes the ansible_user to specify the account, and the generated ansible.cfg specifies a default inventory path and the SSH key file path. However you still need to run retrieve the SSH key file from the storage account. That will be my next goal.

KimForss (approved):

PR #1151, opened 2021-05-26
Alfa

PankajMeshramCSA:
## Problem
Added the Task for Services state  

## Solution
This Task will let the customers specify custom list of services which they want to stop, start, enable or disable 

## Tests
In the vars, please refer to os-services.yaml where one can specify list of services. You would see few services which I have tested against red hat8 OS. 

## Notes

### DIFF
--- a/deploy/ansible/roles-os/1.16-services/tasks/main.yaml
+++ b/deploy/ansible/roles-os/1.16-services/tasks/main.yaml
@@ -1,0 +1,54 @@
+---
+# /*---------------------------------------------------------------------------8
+# |                                                                            |
+# |               Task: 1.16       - Services for OS                           |
+# |                                                                            |
+# +------------------------------------4--------------------------------------*/
+
+- name:             1.16 -  Enable/Disable and Stop/Start Required Services
+  include_vars:     os-services.yaml
+
+# Analyse the service list for this distribution selecting only those
+# services assigned to the active tier or 'all'.
+- name:             Determine services appropriate for tier
+  set_fact:
+    services_to_process: "{{ services[distro_id] |
+                           selectattr('tier', 'in', ['all', tier]) |
+                           selectattr('node_tier', 'in', ['all', node_tier]) |
+                           list }}"
+
+# Print list of matching services if verbosity is 1 or greater
+- debug:
+    var: services_to_process
+    verbosity: 1
+
+# Extract the list of service names whose state match the specified value and
+# pass them as the argument to the name parameter; this is the recommended
+# approach as it only calls the underlying distro specific service manager
+# once per state value.
+- name:             "1.16 - Enable & Start Services : {{ distro_name }}"
+  service:
+    name:           "{services_to_process                      selectattr('state', 'equalto', item.state) |
+                        map(attribute='service') |
+                        list }}"
+    state:          "{{ item.state }}"
+    enabled:        "{{ item.enabled }}"
+  loop:
+    - { state: 'present' }  # First start the required services
+    - { enabled: 'yes' }   # ensure that the required service is enabled
+
+- name:             "1.16 - Disable & Stop Services : {{ distro_name }}"
+  service:
+    name:           "{services_to_process                      selectattr('state', 'equalto', item.state) |
+                        map(attribute='service') |
+                        list }}"
+    state:          "{{ item.state }}"
+    enabled:        "{{ item.enabled }}"
+  loop:
+    - { state: 'stopped' }  # First stop the services which are not required
+    - { enabled: 'no' }   # ensure that the service is stopped
+
+
+# /*----------------------------------------------------------------------------8
+# |                                    END                                      |
+# +------------------------------------4---------------------------------------*/
--- a/deploy/ansible/roles-os/1.16-services/tasks/main.yaml
+++ b/deploy/ansible/roles-os/1.16-services/tasks/main.yaml
@@ -44,15 +44,6 @@
                       map(attribute='service') |
                       list }}"
 
-- name:             "1.16 - Stop Services : {{ distro_name }}"
-  service:
-    name:           "{{ item }}"
-    state:          stopped
-  loop:           "{{ services_to_process |
-                      selectattr('state', 'equalto', 'stopped') |
-                      map(attribute='service') |
-                      list }}"
-
 - name:             "1.16 - Disable Services : {{ distro_name }}"
   service:
     name:           "{{ item }}"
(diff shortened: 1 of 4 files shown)


deploy/ansible/roles-os/1.16-services/tasks/main.yaml
```diff
@@ -0,0 +1,70 @@
...
+- name:             "1.16 - Stop Services : {{ distro_name }}"
+  service:
+    name:           "{{ item }}"
+    state:          stopped
+  loop:           "{{ services_to_process |
+                      selectattr('state', 'equalto', 'stopped') |
+                      map(attribute='service') |
+                      list }}"
+
+- name:             "1.16 - Disable Services : {{ distro_name }}"
```
rtamalin, line 56: So I would probably move the disable to happen before the stop... It is a subtle thing, but usually a good idea to disable the service before stopping it, just in case there is any sort of service dependency chain that can trigger it to be automatically restarted again if not marked as disabled...

The usual logic I apply is that I do a start before an enable, to ensure that the service start is working correctly before enabling it to auto start. And conversely when undoing those I usually do them in reverse order, i.e. disable first then stop...

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
```diff
@@ -0,0 +1,44 @@
...
+# In this list one can specify the custom list of services to be enabled,disabled
+# OR started, stopped 
+
+#Example - Add a line specifying the service, SAP node teir & state 
+# redhat8:
+#      - { tier: 'os', service: 'auditd',       node_tier: 'all',     state: 'started' }
+
+services:
+    redhat7:
+      - { tier: 'os', service: '',  node_tier: 'all',    state: 'started' }
```
rtamalin, line 22: You probably should either fill out entries for RHEL 7, or comment these lines out, as I think an empty service string is being interpretted as "all services" from what your verbose output showed yesterday.

Since this is the same line repeated 3 times, probably for now remove the duplicates, and comment out the remaining line.

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - { tier: 'os', service: '',  node_tier: 'all',   state: 'started' }
+      - { tier: 'os', service: '',  node_tier: 'all',   state: 'started' }
+    
+ 
+    redhat8:
+      - { tier: 'os', service: 'auditd',       node_tier: 'all',     state: 'started' }
+      - { tier: 'os', service: 'multipathd',   node_tier: 'all',     state: 'stopped' }
+      - { tier: 'os', service: 'auditd',       node_tier: 'all',     state: 'disabled' }
+      - { tier: 'os', service: 'nfs-server',   node_tier: 'all',     state: 'enabled' }
```
rtamalin, line 31: Do we want all nodes to be enabled as NFS servers? I believe our current deployment strategy has one node act as an NFS server, while the other nodes that need to access the shared data only need to be NFS clients and not full servers. So you probably need to specify the appropriate node_tier value here...

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - { tier: 'os', service: 'sbd',          node_tier: 'all',     state: 'enabled' }
+       
+   
+    suse12:
+      - { tier: 'os', service: '',  node_tier: 'all',     state: 'started' }
```
rtamalin, line 36: The same as for the RHEL 7 case, probably best to not have duplicate lines, and just leave a single line here, commented out, for reference, until we actually have SLE 12 specific settings.

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - { tier: 'os', service: '',  node_tier: 'all',     state: 'started' }
+      - { tier: 'os', service: '',  node_tier: 'all',     state: 'started' }
+ 
+
+    suse15:
+      - { tier: 'os', service: '',  node_tier: 'all',     state: 'started' }
```
rtamalin, line 42: The same as for the RHEL 7 case, probably best to not have duplicate lines, and just leave a single line here, commented out, for reference, until we actually have SLE 15 specific settings.

rtamalin (changes requested):
You also are missing a change in the `playbook_01_os_base_config.yaml` so that this role will actually be called...

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
PankajMeshramCSA: done as requested.

PankajMeshramCSA:
Hi @Fergal Mc ***@***.***>,

I have resolved all the requests and pushed the changes. Could you please check and review now ?

Pankaj Meshram
Cloud Solution Architect- SAP on Azure
United Kingdom
One Commercial Partner
Phone: +44 (118) 9095515
Mobile: +44 7973 667128
***@***.******@***.***>

***@***.***D75273.56402500]

***@***.******@***.***D75273.56402500]

Planned Leave - 31st May to 4th June

rtamalin:
Looks good now @PankajMeshramCSA.

deploy/ansible/roles-os/1.16-services/vars/os-services.yaml
```diff
@@ -0,0 +1,41 @@
...
+services:
+    redhat7:
+#      - { tier: 'os', service: '',  node_tier: 'all',    state: 'started' }
+     
+    
+ 
+    redhat8:
+      - { tier: 'os', service: 'auditd',       node_tier: 'all',     state: 'started' }
+      - { tier: 'os', service: 'multipathd',   node_tier: 'all',     state: 'stopped' }
+      - { tier: 'os', service: 'auditd',       node_tier: 'all',     state: 'disabled' }
```
KimForss, line 29: Why is auditd specified twice`?

PankajMeshramCSA: I remember this point was raised where a service should be first disabled and stopped. I have added both the states now where a service for example "auditd" is started and enabled or a service for example "multipathd" is disabled and stopped. please review.

rtamalin: The point about disable then stopp is for the internal implementation, not the state represented in the table here.

If the target goal, specified in the table here, is to disable a service, then you implicitly need to also stop any running instances of it as well. But when doing so it is best to disable the service from starting before stopping any running instances, to ensure that nothing can trigger an automatic restart of the service when you stop it.

PR #1153, opened 2021-05-27
Ensure ~/bin/terraform is preferred terraform command

rtamalin:
If you have just created the ~/bin directory, and installed terraform
under it, but there was previously a terraform in a different PATH
directory, you may still end up using that other terraform unless you
force an update of the command hash table.


rtamalin:
@mkdeegan This is just a minor/subtle correction; under rare circumstances if you had never had a ~/bin directory before, and then you create it and add commands to it that may exist elsewhere in your PATH entries, unless you either update the PATH variable or trigger a command hash rebuild, you may continue to use the original version of the command, rather than the one in ~/bin.

It isn't a problem if there is no other occurrence of the specified command, as the bash command hash lookup fails so it has to search the path elements manually for a matching command.

It is only an issue for interactive usage, and I'm familiar with it from working with Python venvs; if you create a venv, then activate it and then subsequently install modules in it that provide commands that already exist on your system, if you run some of them from the shell session under which you installed them, without either running hash or deactivating/reactivating the venv, you can frequently end up running the wrong command. ;-)

mkdeegan (approved):
LGTM

PR #1154, opened 2021-05-27
Changing Kdump LUN id From 3 to 2

balram2697:
## Problem
We use LUN ID '0' for boot lun, LUN ID '1' for config lun and LUN ID '3' for kdump lun ..hence we are skipping LUN ID '2'. So as a standard instead of skipping 1 ID in between, we will start using Kdump lun id as '2'.

## Solution
Changing Kdump LUN ID '3' to '2' so that we don't have to skip any lun id in between.

## Tests
Locally tested this Script with LUN_ID '2' on HANA Instance, it's working as expected (Able to identify the dedicated kdump lun)


prtyag (approved):
LGTM

jaawasth (approved):

KimForss (approved):

PR #1157, opened 2021-05-28
TFVars support for deployer

KimForss:
## Problem
We want to be able to support tfvars formatted variables to simplify the path to DevOps

## Solution
Change the module parameters to use created dictionaries instead of the json structures. This allows a gradual transition for customers

## Tests
Try a normal deployment with a json file and then the same with a tfvars file. A sample tfvars file can be found here: documentation\SAP_Automation_on_Azure\Process_Documentation\WORKSPACES\DEPLOYER\MGMT-WEEU-DEP00-INFRASTRUCTURE\MGMT-WEEU-DEP00-INFRASTRUCTURE.tfvars

## Notes

### DIFF
--- a/deploy/scripts/install_deployer.sh
+++ b/deploy/scripts/install_deployer.sh
@@ -95,11 +95,19 @@ if [ $param_dirname != '.' ]; then
 fi
 
 
+ext=$(echo ${parameterfile} | cut -d. -f2)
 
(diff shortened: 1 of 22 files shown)


hdamecharla (approved):
Hi @KimForss, approved #1157  👍🏾 🙏🏾

PR #1163, opened 2021-06-03
Fix playbook_01 tags enablement

rtamalin:
Previous change just added tags to the include_role actions, but this
approach only works for import_role. For include_role the tags apply
to the include_role statement itself, and not to the included tasks
so unless they also have the same tags configured, or are tagged with
always, they won't be run.
    
To get the desired effect, namely run only the tasks associated with
the included role when the associated task is specified, we need to
specify both a tags list for the include_role, and dynamically apply
the same tags to the included tasks using the apply mechanism. Or
you can wrap the include_role in a block action, with associated
tags, which will apply to both the include_role action and any tasks
that it includes.

### DIFF
--- a/deploy/ansible/playbook_01_os_base_config.yaml
+++ b/deploy/ansible/playbook_01_os_base_config.yaml
@@ -38,48 +38,57 @@
   - set_fact:
       tier:     os
 
-  - include_role:
-      name:     roles-os/1.0-sudoers
+  - block:
+      - include_role:
+          name:     roles-os/1.0-sudoers
     tags:
       - 1.0-sudoers
 
-  - include_role:
-      name:     roles-os/1.1-swap
+  - block:
+      - include_role:
+          name:     roles-os/1.1-swap
     tags:
       - 1.1-swap
 
-  - include_role:
-      name:     roles-os/1.2-hostname
+  - block:
+      - include_role:
+          name:     roles-os/1.2-hostname
     tags:
       - 1.2-hostname
 
-  - include_role:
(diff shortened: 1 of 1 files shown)


rtamalin:
@KimForss @mkdeegan @hdamecharla This sort of change is purely for developer convenience, and provides a way to limit the scope of what is being tested when running the test_menu.sh without having to change the playbook to comment out the other roles... I use a version of this pattern regularly in other projects, though in those cases I'm usually using `import_role` so the `block` wrapping is not needed...
And if this not something we really care about pursuing, then I'm happy to drop this change, and revert the previous change that added the tags...

The main difference between `import_*` and `include_*` is that with an `import_*` the associated code is pulled in and evaluated at "compile" time, i.e. when the ansible-playbook command is initially run, and any problems encountered will result in immediate errors before any plays/tasks are run, whereas for `include_*` the associated code/actions are only evaluated when the `include_*` statement is encountered when running through the plays/tasks, so errors will only be reported at that point.

The main impact of this is that in the case of `import_*` statements any variables/definitions are evaluated at the compile time, whereas with the `include_*` their evaluation is deferred until they are encountreed; this can lead to subtle differences in behaviour, frequently with regard to things like is defined/undefined checks, because variables/attributes defined by "later" operations can appear to be defined early with `import_*` handling.

If there isn't a strong reason to use `include_role` potentially we should probably try to switch to `import_role` as it can result in some performance improvements... Though in our case most of the "slowness" is related to waiting for large data transfers to complete, rather than long running code segments. ;-)

KimForss (approved):

PR #1176, opened 2021-06-11
Ensure SLE 15 SP2 deployments works correctly

rtamalin:
When managing SLE 15 SP2 based images the Ansible automatic Python
interpreter detection process is selecting the /usr/bin/python
interpreter, which is a Python2 version, but the images don't have
all the packages installed for the Python2 stream that are required
to fully support Ansible packaging related operations.

We really should be using the Python3 environment which is already
setup correctly on the SLE 15 SP2 images.

NOTE: It is possible to select the desired Python interpreter using
the ansible_python_interpreter setting, and that would probably be
the better fix in the long run, and would allow this change to be
dropped.


mkdeegan (approved):
LGTM

PR #1185, opened 2021-06-16
WebDispatcher installation on -web VMs

msftrobiro:
## Problem
Ansible playbook to install SAP Web Dispatcher on standalone web VMs

## Solution
New playbook, ini param file for SAP installation for Web Dispatcher

## Tests

## Notes
Two ToDos open to be addressed
- SAPWEBDISP.SAR needs to be part of BOM download, it is NOT part of S/4 Maintenance Planner operation and thus needs to be separately added.
- This PR only covers standalone WDisp, integrated (in ASCS) is NOT covered at the moment. Also not covered presently is ILB or AppGw in front of WebDispatcher VM(s), in Terraform


.vscode/settings.json
```diff
@@ -1,20 +1,20 @@
...
-    "peacock.color": "#42b882"
+        "statusBar.background": "#215732",
+        "statusBar.foreground": "#e7e7e7",
+        "statusBarItem.hoverBackground": "#2f7c47",
+        "titleBar.activeBackground": "#215732",
+        "titleBar.activeForeground": "#e7e7e7",
+        "titleBar.inactiveBackground": "#21573299",
+        "titleBar.inactiveForeground": "#e7e7e799"
+    }
 }
```
mkdeegan, line 20: This file should not be getting committed to GitHub ... 
It should be ignored

deploy/ansible/playbook_05_04_sap_web_install.yaml
```diff
@@ -1,6 +1,6 @@
-# /*----------------------------------------------------------------------------8
+# /*---------------------------------------------------------------------------8
 # |                                                                            |
-# |                   Playbook for SAP Web Dispatcher Install                   |
+# |                 Playbook for SAP Application Server Install                |
```
mkdeegan, line 3: Name is wrong... Not Application Server

deploy/ansible/playbook_05_04_sap_web_install.yaml
```diff
@@ -9,25 +9,43 @@
 
 - hosts:          "{{ sap_sid|upper }}_WEB"
 
-  name:           SAP Installation - WEB
+  name:           SAP Installation - WebDispatcher
+  serial:         1
```
mkdeegan, line 13: Web Dispatchers don't need the serial parameter, I believe

deploy/ansible/roles-sap/5.4-web-install/tasks/main.yaml
```diff
@@ -0,0 +1,127 @@
...
+    sap_profile_dir:              /usr/sap/{{ sap_sid|upper }}/SYS/profile
+    sap_ciVirtualHostname:        "{{ query('inventory_hostnames', '{{ sap_sid|upper }}_PAS') | first }}"
+    sap_cd_package_hdbclient:     
+    sap_cd_package_cd1:           
+    sap_cd_package_cd2:           
+    sap_cd_package_cd3:           
+    sap_cd_package_cd4:           
+    sap_cd_package_cd5:           
+
+#------------------</DEBUGGING>------------------
```
mkdeegan, line 102: Review this code. The as written code follows a old process that will no longer work.
The other install roles have already been updated.

mkdeegan (changes requested):

.vscode/settings.json
msftrobiro: Thanks, corrected.

deploy/ansible/playbook_05_04_sap_web_install.yaml
msftrobiro: Thanks, corrected.

deploy/ansible/playbook_05_04_sap_web_install.yaml
msftrobiro: Unfortunately yes. We use a shared directory for sapinst_instdir. And without serial it clashes (and possibly other errors)
Could add hostname parameter to the installation directory but that should be then done for all installations, as each App servers also overwrites the installation dir since they use same path on nfs.

deploy/ansible/roles-sap/5.4-web-install/tasks/main.yaml
msftrobiro: As discussed, was working off the older code. Corrected and tested with new format.

mkdeegan (approved):

PR #1188, opened 2021-06-23
Improve Terraform driven initial deployer setup

rtamalin:
Update the sap_deployer module to generate a configure_deployer.sh
script under /tmp on the deployer node which is then run using a
remote-exec operation.

The new configure_deployer.sh script is derived from the previous
remote-exec inline scripting, with changes to leverage a Python
virtualenv to host Ansible and other Python tools and depenencies,
rather than pip installing them into the system Python environment
which risks both breaking the system package management, and being
broken by any subsequent Python related package installs.

The generated /etc/profile.d/deployer_setup.sh script should also
correctly add the new Terraform and Ansible bin directories to the
session PATH, rather than replacing the session PATH with a fixed
list of paths that pertained when the terraform remote-exec was
triggered. Note that this script is only included in interactive
sessions, such as a user login via SSH, but not from non-interactive
sessions such as using SSH to run a specific command on a system.

The deploy/ansible/*test_menu.sh scripts have been updated to set
their PATH and ANSIBLE_COLLECTIONS_PATHS environment variables
appropriately to ensure that they can access the required tools 
even when run from a non-interactive session.

Similarly update the deploy/scripts shell scripts to ensure that
they also include the relevant paths in their PATH settings to
ensure that they can find the relevant tools when running under
non-interactive sessions. Note that this change may need to be
re-visited if it is expected that non-deployer environmemnts that
may run these scripts will have the relevant path entries setup
in an incompatible fashion.

### DIFF
--- a/deploy/ansible/test_menu.sh
+++ b/deploy/ansible/test_menu.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export PATH=/opt/terraform/bin:/opt/ansible/bin:${PATH}
+
 cmd_dir="$(dirname "$(readlink -e "${BASH_SOURCE[0]}")")"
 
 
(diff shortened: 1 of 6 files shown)


hdamecharla (commented):
@rtamalin , looks good 👍🏾

PR #1189, opened 2021-06-24
Update cloudinit growpart config management

rtamalin:
Change the existing "disablement" to an "enablement" and rename the
terraform entities to better reflect that it is configuring how the
growpart mechanism is used.

Since we haven't yet completed the partition resize support, better
to let customers make use of the automatic expansion of the OS disk
rather than leaving them with a minimal footprint partition layout
even if they have increased the size of the disk.

Whenever we want to disable growpart expansion we can just switch
the `auto` back to `off` again.


rtamalin:
@KimForss @mkdeegan Sorry for the delay getting this patch ready - got side tracked by meetings/travel paperwork this afternoon.

KimForss (approved):
Looks good to me

PR #1214, opened 2021-07-21
Create the Fencing SPN

KimForss:
## Problem
This PR will create the SPN that will be used by the fencing agent

## Solution
Create the Azure AD application and the service principal and save the attributes in the workload keyvault

## Tests

## Notes
Default behaviour can be overridden by using the create_fencing_spn variable


hdamecharla (approved):
Hi @KimForss, approved #1214 👍🏾 🙏🏾

PR #1217, opened 2021-07-23
Combined all the package handling into the same playbook

KimForss:
## Problem
Currently we handle the packages in two ansible tasks

## Solution
Merge all package handling into the same task controlled by the tier 'os' for the OS packages 'sapos' for the SAP specific packages

## Tests

## Notes


hdamecharla (approved):
Hi @KimForss, Approved #1217 👍🏾 🙏🏾

PR #1225, opened 2021-07-28
Rh testing and changes

hdamecharla:
## Problem
Encountered a few issues while deploying in RHEL

## Solution
Make changes to fix the problem

## Tests
Run on local installation

## Notes


KimForss (commented):
Looks good to me

PR #1229, opened 2021-08-05
Update Getting_started_with_the_SAP_Deployment_Automation_cloudshell.md

mottach:
User needs to login with their service principal before executing the preparing the region step

## Problem

## Solution

## Tests

## Notes


KimForss (commented):
Cloudshell and SPNs do not work well together - from the cloudshell I would rather do an az login so that I get logged in as "me"

PR #1233, opened 2021-08-18
Shellfixes for running in CloudShell

hdamecharla:
## Problem
The client_id and object_id are not populated when running from CloudShell as documented here:
https://github.com/hashicorp/terraform-provider-azurerm/issues/7787

## Solution
export the client_id of executing user when running from cloudshell

## Tests
run prepre_region.sh

## Notes


KimForss (approved):
Looks good to me

PR #1235, opened 2021-08-18
Provide a way to control the deployment of load balancers for non-HA deployments

KimForss:
## Problem
Currently we always deploy load balancers in from to db, scs and web. This adds unneccesary cost for cases where HA is not required

## Solution
Provide a switch that controls if load balancers are created for non HA deployments

## Tests
Add the following variable to your tfvars parameter file: **use_loadbalancers_for_standalone_deployments=false** 

If using json add: 
  **"use_loadbalancers_for_standalone_deployments" : true**

## Notes


hdamecharla (approved):
Hi @KimForss, approved #1235 👍🏾 🙏🏾

hdamecharla (approved):
Hi @KimForss, approved #1235 👍🏾 🙏🏾

PR #1240, opened 2021-08-31
Don't track the python bytecode files

machacekondra:
## Problem
There is no need to track the `.pyc` files.

## Solution
Removing it from the code and adding it to `gitignore`

## Tests
No.

## Notes


KimForss (approved):

KimForss:
Thanks

PR #1244, opened 2021-09-06
Kimforss create sap password

KimForss:
## Problem
The current model expects the hana password to be provided in clear text in sap-parameters.yaml, this is not a recommended practice

## Solution
Add logic in Ansible that will create a random password and store it in the workload zone keyvault if no password is provided. If a password is provided Ansible will store that password in KeyVault

## Tests
Run the ansible playbooks without the password_master value

## Notes


deploy/ansible/roles-misc/0.1-passwords/tasks/main.yaml
```diff
@@ -14,12 +21,37 @@
...
-    - name: "Debug"
-      ansible.builtin.debug:
-        var:  keyvault_secret_show_sap_password_id
\ No newline at end of file
+    - name: "Debug, Error code"
+      ansible.builtin.debug: 
+        msg: "{{ keyvault_secret_show_sap_password_id.rc }}"
+    - name: Construct SAP system password 
+      ansible.builtin.set_fact:
+        sap_password: "{{ lookup('password', '/tmp/sappasswordfile chars=ascii_lowercase,ascii_uppercase,digits length=14') }}"
```
hdamecharla, line 35: @kimforss, we might need to review this when our oracle updates get pushed as it can support additional chars for passwords.

hdamecharla (approved):
Hi @KimForss, approved #1244 👍🏾 🙏🏾

deploy/ansible/roles-misc/0.1-passwords/tasks/main.yaml
KimForss: OK

PR #1248, opened 2021-09-08
Get SAP Library information from keyvault

KimForss:
## Problem
Currently the sap-bits key etc are manual inputs, this is both time consuming and error prone

## Solution
Persist the SAP Library details in deplyer keyvault so that Ansible is able to retrieve the details

## Tests
Re-deploy the SAP Library, the SAP Workload Zone to get the secrets stored

## Notes


hdamecharla (approved):
Hi @KimForss, approved #1248 👍🏾 🙏🏾

PR #1258, opened 2021-09-18
Add protection for deletion of resource groups still containing items

KimForss:
## Problem
Currently a terraform destroy deletes the resource group it created. If there are items created outside of Terraform these would be deleted as well.

## Solution
Add the feature flag blocking terraform destroying resource groups with content

## Tests
Deploy a SAP Deployment and add a VM to the resource group. Try to run terrafrom detroys
## Notes
This will be the default behaviour in Terraform Azure providers v.3 and onwards


hdamecharla (approved):
Hi @KimForss, approved #1258 👍🏾 🙏🏾

PR #1260, opened 2021-09-20
Just warn instead of error some ansible lint rules

machacekondra:
## Problem
Make the `ansible-lint` job green, by just print the errors, but not making the job fail in case of rules: `command-instead-of-shell` and `experimental`. 

## Solution

## Tests

## Notes


KimForss (approved):
Looks good to me

PR #1262, opened 2021-09-22
Disk changes, logical volumes will be conditionally created based on …

msftvapolasa:
…the volumes groups. Files systems will be created based on the logical volumes created.

## Problem
Currently the logical volumes and file systems are not created based on the vg availability. Due to this issue, intermittently  the blkid is not generated. 
Ansible "file system" module is not idempotent. It checks for file systems using blkid. Due to the above problem, it cannot find an entry of the previously created file systems. 
When the sap-os configuration playbook  is rerun , it will attempt to create and format the file systems but fails due to missing entries.

## Solution
We track  the logical volumes created and create the logical volumes based on volumes groups created in previous step.
We then track the logical volumes created and create the file systems based on the logical volumes created in previous step and then track the file systems created and pass it to mkfs.xfs to make an entry into blkid.

## Tests
${DEPLOYMENT_REPO_PATH}/deploy/ansible/test_menu.sh --> Select option2. 
Rerun the above command multiple time to test idempotency of ansible file system module.

## Notes
Ansible "file system" module is not idempotent when it there is not entry of the previously created file system in blkid.

### DIFF
--- a/deploy/ansible/roles-os/1.5-disk-setup/tasks/main.yml
+++ b/deploy/ansible/roles-os/1.5-disk-setup/tasks/main.yml
@@ -63,61 +63,125 @@
 
 # Get all the unique disk types from sap-parameters
 
-# -------------------------------------+---------------------------------------8
-#
-- name:                                Load the disk configuration settings
-  ansible.builtin.include_vars:        disks_config.yml
-
-- name:                                "Print unique disks"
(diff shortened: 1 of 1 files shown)


hdamecharla (approved):
Hi @msftvapolasa, approved #1262 👍🏾

PR #1268, opened 2021-09-24
[Proposal] Execute roles via roles static statement in play

machacekondra:
## Problem
This PR is just a proposal and open to discussion about the playbook structure.
- We can execute roles from `roles:` statement in the play.
- Use `pre_tasks` & `post_tasks` statements for the step file creation/deletion. 
- Use `vars` section to specify the tier.

This is just a possible way of doing the same thing with less verbose code.


machacekondra:
@KimForss @rhmk please take a look.

PR #1270, opened 2021-09-26
Kimforss private endpoint control plane

KimForss:
## Problem
1. Currently we do not support using private endpoints for the storage accounts and the keyvaults in the control plane.
2. Currently we do not use the private keyvault in any of the Terraform or Ansible components.

## Solution
Add a flag which enables the deployment of private endpoints for the storage accounts of SAP Library and the key vault of the deployer.

## Tests
Add 
use_private_endpoint=true
to the deployer and sap library tfvars file

or
"use_private_endpoint" : true
to the deployer and sap library json file

## Notes


hdamecharla (approved):
Hi @KimForss, approved #1270 👍🏾 🙏🏾

PR #1275, opened 2021-09-28
sapmnt nfs mount is missing on PAS

msftvapolasa:
## Problem
sapmnt nfs mount is missing on PAS

## Solution
Add tier==pas condition in sap-mounts. 
Call the sap-mounts block in pas install playbook.

## Tests
Install PAS.

## Notes


KimForss (approved):
Looks good to me

PR #1277, opened 2021-09-28
Add the ability to deploy HANA with just a single NIC

KimForss:
## Problem
Currently the HANA deployment has two network interfaces, not all customers want that.

The AnyDB and the App tiers already have that capability

## Solution
Provide an ability to use just a single network interface for HANA

## Tests
Provide 
hana_dual_nics=false in tfvars or json file

## Notes


sqlchow (approved):
Hi @KimForss, approved #1277

PR #1278, opened 2021-09-29
Remove '?' from sapbits downlaod URL

machacekondra:
## Problem
When the sas_token is generated it already contains question mark, and
so the current code contruct url with two '??', which is invalid.


machacekondra:
@KimForss please take a look.

KimForss (approved):

PR #1279, opened 2021-09-29
Don't fallback on 'none' string

machacekondra:
## Problem
This patch removes checks that rely on fallback to 'none' string and
rather check that the variable is undefined or defined.


machacekondra:
@KimForss please take a look.

hdamecharla (approved):
Hi @machacekondra, approved #1279 👍🏾 🙏🏾 

cc: @KimForss

PR #1282, opened 2021-09-30
[Proposal] Use redhat_sap.sap-hana-deployment role to install hana

machacekondra:
## Problem

This PR changes the `playbook_04_00_00_hana_db_install.yaml` to use `redhat_sap.sap_hana_deployment` instead of `roles-db/4.0.0-hdb-install`.

This is just a propsal and shows how we could reuse the roles, open to discussion, whether it makes sense or not.

Currently the `redhat_sap.sap_hana_deployment` role miss following features which the role `roles-db/4.0.0-hdb-install` has:
  - components and compoenents_root configuration
  - reinstall option

This also brings the question of naming the roles with number prefix, because in case of usage the external roles, which are not named  with any prefix, looses the meaning of using the number prefix for all other roles.


machacekondra:
@rhmk @KimForss please take a look whether this makes sense to you. Just a note that we could reuse more of the installation roles, not just HANA installation. This is working example but not meant to be merged, but rather discuss.

KimForss:
We need to be able to be distro agnostic, would this also work on SUSE?

machacekondra:
> We need to be able to be distro agnostic, would this also work on SUSE?

Yes, it would. The [sap-hana-deployment](https://github.com/redhat-sap/sap-hana-deployment/) role does exactly the same thing as `roles-db/4.0.0-hdb-install` role. It would be nice if we can develop and maintain just one role, if it does the same thing. I think we should maybe discuss and decide what's the best place to store such roles, which could be reused.

@rhmk can comment more here.

PR #1283, opened 2021-09-30
Env and region needed before env config path generated

rtamalin:
We need to ensure we have the env and region values assigned, either
via CLI options, or via prompted data entry, before we assign the path
to the environment specific file under ~/.sap_deployment_automation.


hdamecharla (approved):
Hi @rtamalin, approved #1283 👍🏾 🙏🏾 

cc: @KimForss

PR #1284, opened 2021-09-30
Fix some issues in load_config_vars

rtamalin:
In load_config_vars(), after saving the first argument as var_file, we
need to shift the parameters list by one to avoid trying to lookup the
var_file's path as a variable name in the var_file file.

Also since we don't want to set/update variable values if they are not
found in the var_file, just return immediately when var_file doesn't
exist, rather than repeatedly checking for it's existence, once per
specified variable.

### DIFF
--- a/deploy/scripts/deploy_utils.sh
+++ b/deploy/scripts/deploy_utils.sh
@@ -27,16 +27,23 @@ function save_config_vars() {
 }
 
 function load_config_vars() {
-    local var_file="${1}"
-    local var_name="${2}"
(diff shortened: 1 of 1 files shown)


rtamalin:
@KimForss @hdamecharla This cleans some stuff up, while retaining the same semantics as before, i.e. don't assign an empty value to a variable, either because it wasn't found in the file, was empty, or the specified file doesn't exist...

KimForss (approved):
Looks good to me

PR #1285, opened 2021-09-30
root_dirname used before it is assigned

rtamalin:
Remove the unnecessary use of the root_dirname prefix on the deployer
parameter tfvars file path when checking for environment and location
values.


rtamalin:
@KimForss @hdamecharla Found and fixed while doing cleanup to test prepare_region.sh

KimForss (approved):

PR #1286, opened 2021-10-01
Confirm environment and region non-empty in set_secrets.sh

rtamalin:
In order to safely construct the path to the relevant config file
we need to ensure that the environment and region values that are
provider via CLI option, or via prompted user input, are non-empty
values that conform to the expected validation pattern.

Also when prompting for the user to enter a password field we don't
need to explicitly manage terminal echoing if we are specifying the
-s option to the read command since it will both handle disablement
of terminal echo correctly, and more importantly, will restore the
terminal echo mode if the user Ctrl-C's the script run.

Additionally the -e option enables readline processing which makes
no sense when enterring a password, as we don't necessarily want
or need readline editing features when entering a password value.

### DIFF
--- a/deploy/scripts/set_secrets.sh
+++ b/deploy/scripts/set_secrets.sh
@@ -110,6 +110,23 @@ while [ -z "${region}" ]; do
     read -r -p "Region name: " region
 done
 
+# NOTE: If we want to be more specific on the permitted number of
+# letters in the environment name, e.g. between 3 and 6 letters,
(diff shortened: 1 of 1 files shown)


rtamalin:
@KimForss @hdamecharla When you get a chance

hdamecharla (approved):
Hi @rtamalin, approved #1286 👍🏾 🙏🏾  

cc: @KimForss

PR #1287, opened 2021-10-01
Enhance region and environment regexp checks

rtamalin:
Define valid_environment and valid_region_name helper functions in
deploy_utils.sh that can be used to validate that environment and 
region name strings conform to the expected formats.
 
As a starting point we are requiring the following:
    
  * environments are specified as a string of at most 5 characters,
    and at least 2 characters, starting with an uppercase letter,
    followed by one or more letters or numbers.
    
  * region names are specified as a string of lowercase letters
    optionally followed by numbers.
    
Update set_secrets.sh to leverage these new helper functions when
checking that provided environment and region values are specified
correctly.

### DIFF
--- a/deploy/scripts/deploy_utils.sh
+++ b/deploy/scripts/deploy_utils.sh
@@ -363,6 +363,25 @@ function print_script_name_and_function() {
     echo -e "\t[$(basename "")]: $(basename "$0") $1"
 }
 
+#
+# Input validation routines
+#
+
+# An environment value must be a string that is at most 5 characters
+# long, made up of uppercase letters and numbers, and must start with
+# an uppercase letter.
+function valid_environment() {
+    [[ "${environment}" =~ ^[[:upper:]][[:upper:][:digit:]]{1,4}$ ]]
+}
+
+# A region name must be a valid Azure lowercase region name, made
+# up of lowercase latters optionally followed by numbers.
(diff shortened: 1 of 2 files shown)


rtamalin:
@KimForss @hdamecharla Forgot that regions can have numbers in the name ;-)

deploy/scripts/deploy_utils.sh
```diff
@@ -363,6 +363,25 @@ function print_script_name_and_function() {
...
+function valid_environment() {
+    [[ ! "${environment}" =~ ^[[:upper:]][[:upper:][:digit:]]{1,4}$ ]]
+}
+
+# A region name must be a valid Azure lowercase region name, made
+# up of lowercase latters optionally followed by numbers.
+# NOTE: If we have the list of possible regions in a file somewhere
+# we can validate it is one of the entries in that list.
+function valid_region_name() {
+    [[ ! "${region}" =~ ^[[:lower:]]+[[:digit:]]*$ ]]
```
hdamecharla, line 382: Hi @rtamalin, this would be `^[[:lower:]]+[[:digit:]]$`

hdamecharla (approved):
Hi @rtamalin, approved #1287 👍🏾 🙏🏾

cc: @KimForss

PR #1288, opened 2021-10-04
Report adding the subscription secret in set_secrets.sh

rtamalin:
To avoid any misunderstandings, ensure that we report setting the
subscription id in the same way that we report the setting of the
SPN client id, secret and tenant id.


rtamalin:
@hdamecharla @KimForss This should help others avoid drawing incorrect conclusions from the output of set_secrets.sh like I did on Friday ;-)

KimForss (approved):

PR #1313, opened 2021-11-15
Patch hostname length

hdamecharla:
## Problem
Currently HANA database hostname is 14 characters in length. While the automation does not cater to the scenario of running a dbload from the database VM, a manual run could be initiated from the database VM that can cause the dbload to fail.

## Solution
Reduce the length of the randomized results to 2 chars. This is the simplest route to take as the other parts that make up the VM name do not change.

## Tests
pull the new code and deploy a new system.


KimForss (approved):
Looks good to me

PR #1314, opened 2021-11-15
BOM for BW4HANA2021-Initial Shipment

msftvapolasa:
## Problem
BOM for BW4HANA2021 doesn't exists.
generate_bom.sh generates old product id values. 

## Solution
generate_bom.sh generates old product id values. 
- Updated to reflect the latest code in alfa
- added new print line for "path"
- Updated extract directory for SWPM files
Created New BOM for SAPBW4HANA2021.

## Tests
Run generate_bom.sh
Run playbook 1,2,3,4,5,6,7,8 using the BOM value as BW4HANA2021SPS01_v0001ms.

## Notes


hdamecharla (changes requested):
Hi @msftvapolasa, could we cleanup the inifiles and remove the encryption hash and the DES hashes from these, please?

hdamecharla (approved):
Hi @msftvapolasa , approved #1314 , looks good to me 👍🏾 

cc: @KimForss

PR #1315, opened 2021-11-16
Patch ppg new guidance

hdamecharla:
## Problem
We wanted to adopt the terraform modules to allow for deployment using the new PPG guidance made available recently

## Solution
Provide code to order the deployment in a way which allows the scs VM to act as an anchor for zonal deployments
allow for application server deployments to adhere to zone base deployments
misc fixes

## Tests
run a sap_system deployment 

## Notes
none


KimForss (approved):
looks good to me

